PR #18, opened 2015-02-17
Remove base_path from Publishing API request body

jamiecobbett:
Publishing API currently ignores this field and is about to start rejecting it.
It uses the version in the request URL instead.

For more context: alphagov/content-store#104


jamiecobbett:
Bump. Simplification and fewer lines, what's not to love?

PR #19, opened 2015-03-18
Add tag management

KushalP:
This is PR is a large change and probably requires reviewing by @alext and @rboulton. It moves the tag management UI into Collections Publisher, keeping Panopticon up-to-date with any changes as they happen.

Worth saying that this is largely work done by @JordanHatch (thanks!), but the final two commits from me are to close off the story.

Story: https://www.pivotaltracker.com/n/projects/1110932/stories/79178100


config/initializers/services.rb
```diff
@@ -22,3 +22,6 @@ class ServiceNotRegisteredException < Exception; end
 
 require 'gds_api/publishing_api'
 CollectionsPublisher.services(:publishing_api, GdsApi::PublishingApi.new(Plek.new.find('publishing-api')))
+
+require 'gds_api/panopticon'
+CollectionsPublisher.services(:panopticon, GdsApi::Panopticon.new(Plek.new.find('panopticon'), bearer_token: 'example'))
```
alext: We'll need some mechanism to specify the real bearer token at deploy time, ideally without overwriting this whole file.

As an aside, the `services` method definition above really doesn't belong in here, but that's unrelated to this PR.

alext:
:+1: Looks good to me.  I'll let @rboulton give this a once-over as well...

elliotcm:
Happy to see this completed :star2:

rboulton:
:fireworks: :100:

JordanHatch:
Fantastic, thanks @KushalP! :palm_tree:

PR #21, opened 2015-03-23
Load tags from Content API into local database

KushalP:
Tags will be migrated from Panopticon into Collections Publisher. To
make this process smoother, create a Rake task to do most of the work
for us.

The task itself aims to only import `section` and `specialist_sector`
tags. It then creates any tags that don't already exist.

### DIFF
--- a/lib/tasks/load_tags_from_content_api.rake
+++ b/lib/tasks/load_tags_from_content_api.rake
@@ -1,0 +1,48 @@
+desc "Load tags from Content API into Collections Publisher."
+task :load_tags_from_content_api => :environment do
+  require 'gds_api/content_api'
+
+  content_api_endpoint = Plek.new.find("content_api")
+  content_api = GdsApi::ContentApi.new(content_api_endpoint)
+
+  puts "There are currently #{Tag.count} tags in the database"
+  puts "Fetching all tags from the Content API..."
+  tags = content_api.get_list!("#{content_api_endpoint}/tags.json").
+         with_subsequent_pages.
+         # Only import the tags that are sectors or specialist sectors.
+         select { |tag| ["section", "specialist_sector"].include?(tag.details.type) }
+
+  puts "Received a total of #{tags.count} relevant tags"
+  puts "Loading tags into Collections Publisher..."
+  tags.each do |api_tag|
+    unless Tag.where(slug: api_tag.slug).any?
+      case api_tag.details.type
+      when "section"
+        klass = MainstreamBrowsePage
+      when "specialist_sector"
+        klass = Topic
+      end
+
+      tag = klass.new(slug: api_tag.slug,
+                      description: api_tag.details.description,
+                      title: api_tag.title)
+      tag.save!
+    end
+  end
+
+  # On the first run of this task, the tags might not exist in the database.
+  # Traverse through all fetched tags again and assign parents accodingly.
+  puts "Assigning child tags to parents..."
+  tags.each do |api_tag|
+    existing_tag = Tag.where(slug: api_tag.slug).first
+
+    if existing_tag.parent.nil? and not api_tag.parent.nil?
+      parent = Tag.where(slug: api_tag.parent.slug).first
+      existing_tag.parent = parent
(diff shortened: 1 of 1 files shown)


lib/tasks/load_tags_from_content_api.rake
```diff
@@ -0,0 +1,39 @@
...
+
+  puts "There are currently #{Tag.count} tags in the database"
+  puts "Fetching all tags from the Content API..."
+  tags = content_api.get_list!("#{content_api_endpoint}/tags.json")
+
+  puts "Received a total of #{tags.count} tags"
+  puts "Loading tags into Collections Publisher..."
+  tags.each do |api_tag|
+    unless Tag.where(slug: api_tag.slug).any?
+      tag = Tag.new(slug: api_tag.slug,
```
rboulton: I think, if the tag exists already, we should be ensuring that its title and description match.  Otherwise, we might lose some data when we turn off panopticon.

lib/tasks/load_tags_from_content_api.rake
```diff
@@ -0,0 +1,39 @@
+desc "Load tags from Content API into Collections Publisher."
+task :load_tags_from_content_api => :environment do
+  require 'gds_api/content_api'
+
+  content_api_endpoint = Plek.new.find("content_api")
+  content_api = GdsApi::ContentApi.new(content_api_endpoint)
+
+  puts "There are currently #{Tag.count} tags in the database"
+  puts "Fetching all tags from the Content API..."
+  tags = content_api.get_list!("#{content_api_endpoint}/tags.json")
```
alext: This will pull in all tags from content_api.  We probably need to filter these by the types we're interested in (section/subsection, and topic/subtopic).

lib/tasks/load_tags_from_content_api.rake
```diff
@@ -0,0 +1,39 @@
...
+    existing_tag = Tag.where(slug: api_tag.slug).first
+
+    if existing_tag.parent.nil? and not api_tag.parent.nil?
+      parent = Tag.where(slug: api_tag.parent.slug).first
+      existing_tag.parent = parent
+      existing_tag.save!
+    end
+  end
+
+  puts "There are now #{Tag.count} tags in the database"
```
rboulton: Hopefully this will equal the number of tags in panopticon.  I can imagine there are some in collections-publisher but not in panopticon already, though, so perhaps we should explicitly check that there are no excess tags.

lib/tasks/load_tags_from_content_api.rake
KushalP: There are currently no tags in collections-publisher, so there's no issue of them not matching the first time this is run. I did this to prevent future runs preventing any changes our editors might have might.

lib/tasks/load_tags_from_content_api.rake
```diff
@@ -0,0 +1,39 @@
...
+  puts "Fetching all tags from the Content API..."
+  tags = content_api.get_list!("#{content_api_endpoint}/tags.json")
+
+  puts "Received a total of #{tags.count} tags"
+  puts "Loading tags into Collections Publisher..."
+  tags.each do |api_tag|
+    unless Tag.where(slug: api_tag.slug).any?
+      tag = Tag.new(slug: api_tag.slug,
+                    description: api_tag.details.description,
+                    title: api_tag.title)
```
alext: We ought to be creating tags of the approprite type (`Topic` or `MainstreamBrowsePage`) here.  As I understand it, there shouldn't really be any instances of the base `Tag` class in the db.

lib/tasks/load_tags_from_content_api.rake
KushalP: The different types we have tags for are as follows:

```
=> ["section", "specialist_sector", "organisation", "legacy_source"]
```

alext: From a quick look at the JSON from contentapi, I think the 2 we're interested in are "section", and "specialist_sector".

PR #28, opened 2015-04-16
Remove capistrano.

alext:
Not sure why this is here (it was added in the big 'initial commit').
Our deploy scripts live in a separate repo, so there's no need for it
here.


jamiecobbett:
Similarly: https://github.com/alphagov/email-alert-api/pull/69

PR #29, opened 2015-04-17
Add more integration tests.

alext:
This adds integration tests for creating/editing topics.  It also adds pending assertions for sending both topics and mainstream_browse_pages to the publishing-api.


spec/features/mainstream_browse_pages_spec.rb
```diff
@@ -55,6 +64,45 @@
...
+      "description" => "Voting",
+      "format" => "mainstream_browse_page",
+    })
+
+    # And the page should have been updated in Panopticon
+    assert_tag_updated_in_panopticon(
+      :tag_type => 'section',
+      :tag_id => 'citizenship',
+      :title => 'Citizenship in the UK',
+      :description => 'Voting',
```
rboulton: should this not send the fact that this is in draft state to panopticon?

alext: It should, but that's not what the code currently does.  These have therefore revealed a bug.  I'll pick that up in a separate PR if that's ok...

spec/features/topic_create_edit_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+    end
+
+    # And a draft should have been sent to publishing-api (pending)
+    #assert_publishing_api_put_draft_item('/working-at-sea', {
+      #"title" => "Working at sea",
+      #"description" => 'The sea, the sky, the sea, the sky...',
+      #"format" => "topic",
+    #})
+
+    # And the topic should have been created in Panopticon
```
rboulton: in draft state

rboulton:
This looks good to me.  I'll merge, so you can look at the draft state bug separately.

spec/features/mainstream_browse_pages_spec.rb
alext: Actually, after further investigation, the Panopticon API has a separate `/publish` endpoint for changing the state of a tag from draft to published.  The create/update endpoints don't change the state.

PR #35, opened 2015-04-21
Fix Tag#base_path for child tags.

alext:
This is currently joining the child slug with the parent slug.  This is
incorrect because the child slug actually includes the parent slug in
the database.

We should revisit this at some point because it probably makes sense for
the child slug to not include the parent slug.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -59,8 +59,8 @@ class Tag < ActiveRecord::Base
   end
 
   def base_path
-    base = has_parent? ? "/#{parent.slug}" : ''
-    "#{base}/#{slug}"
+    # a child tag's slug field includes the parent slug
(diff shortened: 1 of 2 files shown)


rboulton:
I agree it would be better for the child slug not to include the parent slug; in particular because then changing the slug of the parent becomes easier.

How hard would it be to fix this properly now?  ie, to make child slugs be standalone.

alext:
It would be possible to fix the modelling now, but it'll make the sector model replacement harder.  I'd prefer to address this once that other piece is complete.

rboulton:
Summary of offline discussion - we could fix this reasonably easily, but the migration from the sector model to the topics model will be easier with the current situation.  Therefore, we'll fix this after the migration.

PR #36, opened 2015-04-22
Allow links to be made to topics from browse pages

KushalP:
This change allows topics to be associated with mainstream browse pages. It's broken up into three sections:

1. Creating a join table
2. Associating topics to mainstream browse pages
3. Making the UI consistent with other applications

## Creating a join table

Commits c3a6830 and e9691a5 add a new model `TagAssociation` which creates a join table on the underlying `Tag` model. This is because both `Topic` and `MainstreamBrowsePage` are `Tag` models underneath. Friendly method of `topics` and `mainstream_browse_pages` have been added to these respective models so that we don't have do any heavy lifting for querying or updating these values.

## Associating topics to mainstream browse pages

Commits da35c26 through 9fcde18 wire up the `TagAssociation` join table to views on the `MainstreamBrowsePage` object.

## Making the UI consistent with other applications

Finally, commits 822957b through 302e0c2 make the UI nicer for the multi-select we've added. Other applications like alphagov/publisher use select2-rails to provide this improved UI for multi-select form fields so it's nothing new for our admin styles.


app/assets/javascripts/application.js
```diff
@@ -19,4 +21,12 @@ GOVUK.curatedLists.init();
       }
     });
   });
+
+  $(document).ready(function() {
+    var select2Element = $(".select2")
+
+    if (select2Element.length) {
+      select2Element.select2();
+    }
```
alext: Is this conditional necessary?  Doesn't jQuery let you do something like
```js
$(".select2").select2();
```

db/migrate/20150416095248_create_tag_associations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTagAssociations < ActiveRecord::Migration
+  def change
+    create_table :tag_associations do |t|
+      t.integer :from_tag_id, index: true, null: false
+      t.integer :to_tag_id, index: true, null: false
+
+      t.timestamps
+    end
+  end
```
alext: We could do with some indexes here. A unique index on `[:from_tag_id, :to_tag_id]` would be good to ensure data integrity.  We'd also then want one on `to_tag_id` to allow efficiently traversing in the reverse direction.

spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
```diff
@@ -0,0 +1,65 @@
+require "spec_helper"
+
+# NOCOMMIT: Don't commit this until it's passing!
```
alext: I don't think you meant to commit this comment...

spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  context "existing mainstream browse pages" do
+    let(:mainstream_browse_page) { create(:mainstream_browse_page) }
+    let(:topic)                  { create(:topic) }
+    let(:topic_two)              { create(:topic) }
+
+    before do
+      mainstream_browse_page.save
+      topic.save
+      topic_two.save
```
alext: You don't need this block if you replace the above `let` calls with `let!` calls.  `let` blocks are lazily evaluated, so the models won't be created until the method is first called.  As the code stands, it's actually doing a double-save (once from the create call when the variable is referenced, and once from the explicit save call.)

spec/models/tag_association_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+#  updated_at  :datetime
+#
+
+require 'spec_helper'
+
+describe TagAssociation do
+  before do
+    @mainstream_browse_page = MainstreamBrowsePage.new(
+      slug: "housing", title: "Housing", description: "All about housing")
+    expect(@mainstream_browse_page.save).to be_true
```
alext: Any reason not to use the factories here?

app/assets/javascripts/application.js
KushalP: Fixed in df32316.

db/migrate/20150416095248_create_tag_associations.rb
KushalP: Fixed as part of initial migration in 19f66ed.

spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
KushalP: You're right. Fixed in f40aa1f.

spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
KushalP: I didn't know that, thanks! Fixed in d66bab8.

spec/models/tag_association_spec.rb
KushalP: Fixed in 19b13ca.

app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -4,6 +4,35 @@ class MainstreamBrowsePagesController < ApplicationController
...
+    end
+
+    if @resource.update_attributes(msb_params)
+      PanopticonNotifier.update_tag(
+        presenter_klass.new(@resource)
+      )
+
+      redirect_to mainstream_browse_page_path(@resource)
+    else
+      render :edit
```
alext: This will need `@topics` assigned like the edit action, otherwise the view won't work correctly.

spec/models/tag_association_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(mainstream_browse_page.save).to be_true
+
+      expect(mainstream_browse_page.topics).to_not be_empty
+      expect(topic.destroy).to be_true
+
+      mainstream_browse_page.reload
+      expect(mainstream_browse_page.topics).to be_empty
+    end
+
+    it "destroys associated topics when mainstream browse page destroyed" do
```
alext: The wording of this is a little unclear, it suggests that the topics themselves should be destroyed...  Maybe "destroys the topic associations when a mainstream browse page is destroyed"

Similarly for the spec above...

app/controllers/mainstream_browse_pages_controller.rb
KushalP: Fixed in cbb7a3d.

spec/models/tag_association_spec.rb
KushalP: Fixed in cfcadc6.

PR #38, opened 2015-04-23
Replace Sector model with Topic model

alext:
This replaces all use of the Sector pseudo-model with the Topic model.  The Sector model used to wrap the sectors in the content API.  

Given that this app is now the canonical source of Topics, and merely mirrors them into the content API, there's no reason to use content API for this.  This will also make it easier to publish Topics and MainstreamBrowsePages to the publishing-api.


app/controllers/application_controller.rb
```diff
@@ -10,7 +10,20 @@ class ApplicationController < ActionController::Base
...
   end
 
+  private
+
   def require_gds_editor_permissions!
     authorise_user!("GDS Editor")
   end
+
+  # FIXME: clean this up when we're using content_ids in the URL.
+  def find_topic_for_sector_id
```
rboulton: Would this make sense to be a method on the Topic model?

Or perhaps even on the Tag model, since I think looking up Tags by path is likely to be wanted, too.

alext: I agree that would probably be the best place for it.  However, I plan to shortly change this to not use these slugs in the URLs, so this will go away very soon.

Happy to update this if you think it's worth it...

rboulton: Fine, leave it :)

PR #39, opened 2015-04-23
Reset the database on each CI run.

alext:
Rails' automatic schema maintenance doesn't cope 100% with switching
between branches with different schemas.  Specifically, it doesn't
handle changes to a column's null state.

Explicitly dropping and re-creating the test database on each run
ensures that we're running with a clean state.

This should prevent test failures like https://ci-new.alphagov.co.uk/job/govuk_collections_publisher_branches/174/console


jamiecobbett:
> Rails' automatic schema maintenance doesn't cope 100% with switching
between branches with different schemas. Specifically, it doesn't
handle changes to a column's null state.

:disappointed: Good catch though :+1:

PR #41, opened 2015-04-24
Existing topics appear when editing mainstream browse pages

KushalP:
When editing `MainstreamBrowsePage` items, show the already assigned
`Topic` values.

The response to this involved updating the use of
`options_from_collection_for_select` to provide the existing topics as
the fourth parameter (`selected`) in the form of an array of ids.

Also switched out the fields we're using to be symbols instead of
strings as that's what they're being converted to behind the scenes
anyway.


rboulton:
:+1:

PR #51, opened 2015-04-27
Remove decent_exposure.

alext:
It's no longer being used anywhere in the app.


rboulton:
:egg: :hatching_chick: :hatched_chick:

PR #52, opened 2015-04-27
Update rspec config to match newly created app.

alext:
RSpec in a new app sets up a separate spec_helper and rails_helper
files, and suggests a number of default options.  This PR updates the
config here to match, and enables the recommended options.

The most significant option is the disable_monkey_patching option, which
means that RSpec won't add any methods to any objects outside its scope.
This means that we now have to use the expect/allow style syntax, and
the deprecated should/should_not syntax won't work.  We also have to do
`RSpec.describe` at the top level of the test files.


rboulton:
:+1: for fewer :monkey:s

rboulton:
(Would merge, but requires a rebase)

alext:
rebased...

PR #53, opened 2015-04-27
MainstreamBrowsePage: only children can have topics associated

KushalP:
Top-level (parent) mainstream browse page records such as `/browse/driving` shouldn't have topics assigned to them. Only allow child `MainstreamBrowsePage` records to have topics assigned to them. Parents shouldn't have topics assigned to them.

### DIFF
--- a/spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
+++ b/spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
@@ -6,6 +6,18 @@ describe "associating topics to mainstream browse pages" do
     stub_all_panopticon_tag_calls
   end
 
+  context "parent mainstream browse pages" do
+    let!(:mainstream_browse_page) { create(:mainstream_browse_page) }
+
+    it "should not allow adding any topics" do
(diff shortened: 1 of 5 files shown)


rboulton:
Good work, thanks. :+1:

spec/features/associating_topics_to_mainstream_browse_pages_spec.rb
```diff
@@ -6,10 +6,28 @@
...
+
+      within "form.resource" do
+        expect(page).to_not have_selector("#mainstream_browse_page_topics")
+      end
+    end
+  end
+
+  context "existing mainstream browse pages" do
+    let!(:mainstream_browse_page_parent) { create(:mainstream_browse_page) }
+    let!(:mainstream_browse_page)        { create(:mainstream_browse_page) }
```
alext: You could change this to be `create(:mainstream_browse_page, :parent => mainstream_browse_page_parent)`, and then you wouldn't need the before block below...

KushalP: I've created #55 to do this.

PR #57, opened 2015-05-06
Add schema contract tests

alext:
This adds contract tests for the publishing-api presenters to test they produce output that conforms to the schema.

It also adds all the necessary mechanisms to run just the schema tests from a separate Jenkins job.  This will run any tests that are tagged `:schema_test => true`.


jamiecobbett:
> It also adds all the necessary mechanisms to run just the schema tests from a separate Jenkins job. This will run any tests that are tagged :schema_test => true.

Nice, we should copy this pattern to other apps.

alext:
> Nice, we should copy this pattern to other apps.

If this PR is considered to be a good approach, then maybe we could update the [example in govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/blob/master/docs/contract-testing-howto.md#setting-up-the-ci-builds) to point at this app.

PR #61, opened 2015-05-15
Use content_id in URLs

tijmenb:
Previously, `panopticon_slug` was used in URLs. 

The only code that now uses `Topic#panopticon_slug` is https://github.com/alphagov/collections-publisher/blob/4c078e0259ea1d5742f23149e925b966ba28000b/app/models/topic.rb#L43, but that appears to be correct. @rboulton ?

In preparation for https://trello.com/c/u7tOFFXk/109-merge-sectors-and-topics-interfaces-in-collections-publisher

from :train:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::Base
     authorise_user!("GDS Editor")
   end
 
-  def find_topic_for_sector_id
+  def find_topic
     @topic = Topic.find_by!(content_id: params[:sector_id])
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,14 +12,7 @@ class ApplicationController < ActionController::Base
     authorise_user!("GDS Editor")
   end
 
-  # FIXME: clean this up when we're using content_ids in the URL.
   def find_topic_for_sector_id
-    if params[:sector_id].include?('/')
-      parent_slug, child_slug = params[:sector_id].split('/', 2)
-      parent = Topic.find_by!(:slug => parent_slug)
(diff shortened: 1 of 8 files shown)


rboulton:
Yes, that use of panopticon_slug is correct, still.

I was wondering if we should redirect the old urls containing slugs; if a user has bookmarked a page this change will break that.  It might be convenient to be able to paste a topic page path into the url and go to the right place, too.

app/controllers/application_controller.rb
```diff
@@ -12,14 +12,7 @@ def require_gds_editor_permissions!
...
-  def find_topic_for_sector_id
-    if params[:sector_id].include?('/')
-      parent_slug, child_slug = params[:sector_id].split('/', 2)
-      parent = Topic.find_by!(:slug => parent_slug)
-      @topic = parent.children.find_by!(:slug => child_slug)
-    else
-      @topic = Topic.only_parents.find_by!(:slug => params[:sector_id])
-    end
+  def find_topic
+    @topic = Topic.find_by!(content_id: params[:sector_id])
```
alext: Should we rename the param to something like `topic_id` at this point to avoid confusion?

tijmenb: Indeed. It is still `:sector_id` because it's generated by the `resources :sectors` routes. I'm working on a PR now to remove anything related to sectors, this will be renamed then.

alext: That makes sense... Thanks.

tijmenb:
Recap irl conversation: redirection might be a very nice thing to do, but adds complexity and code that we want to avoid.

Searching for a topic page path is better supported by [adding the filterable tables module to tables](https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/javascripts/govuk-admin-template/modules/filterable_table.js).

PR #64, opened 2015-05-18
Refactor Topic & Browse page controllers

tijmenb:
This refactor removes the `TagCreateUpdatePublish` concern.

This adds a little bit of repetition between `MainstreamBrowsePagesController` and `TopicsController`, but the result is much cleaner and easier to understand.

- Cleaner templates. Gets rid of conditional.
- Clearer code paths - some generic `shared/tags` templates were only used in topics.

Trello: https://trello.com/c/u7tOFFXk/109-merge-sectors-and-topics-interfaces-in-collections-publisher


app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -1,46 +1,84 @@
...
+    browse_page = find_browse_page
+
+    if browse_page.update_attributes(browse_page_params)
+      PanopticonNotifier.update_tag(MainstreamBrowsePagePresenter.new(browse_page))
+      PublishingAPINotifier.send_to_publishing_api(browse_page)
+      redirect_to browse_page
     else
-      msb_params = tag_params
+      @topics_for_select = topics_for_select
+      render :edit
```
alext: At this point `@browse_page` isn't set, so the view won't work correctly.

app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -1,46 +1,84 @@
 class MainstreamBrowsePagesController < ApplicationController
-  include TagCreateUpdatePublish
-  tag_create_update_publish_for MainstreamBrowsePage
-
   before_filter :require_gds_editor_permissions!
 
-  def edit
-    set_topics_for_select
+  def index
+    @pages = MainstreamBrowsePage.sorted_parents
```
alext: We're using `@pages` here, and `@browse_page` in other actions.  Would be good to make these consistent.

app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -1,46 +1,84 @@
...
+  end
+
 private
-  def presenter_klass
-    MainstreamBrowsePagePresenter
+
+  def tag_params
+    params.require(:mainstream_browse_page)
+      .permit(:slug, :title, :description, :parent_id)
+  end
```
alext: Is there any reason to keep this method separate from `browse_page_params` any more?

app/controllers/topics_controller.rb
```diff
@@ -1,18 +1,72 @@
...
+  def create
+    topic = Topic.new
+    topic.attributes = topic_params
+
+    if topic.save
+      PanopticonNotifier.create_tag(TopicPresenter.new(topic))
+      PublishingAPINotifier.send_to_publishing_api(topic)
+      redirect_to topic
+    else
+      @resource = topic
```
alext: I think you meant to use `@topic` here, not `@resource`

app/models/tag.rb
```diff
@@ -58,6 +58,10 @@ def can_have_children?
     parent_id.blank?
   end
 
+  def child?
+    parent_id.present?
+  end
```
alext: This might be better implemented as an alias for `has_parent?`

app/views/mainstream_browse_pages/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= content_for :page_title, "Mainstream browse page" %>
+
+<header class="heading-with-actions">
+  <h1>Mainstream browse page</h1>
+
+  <% if gds_editor? %>
```
alext: This conditional isn't necessary.  Mainstream browse pages are restricted wholesale in the controller.

app/views/topics/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+        <span class="icon"></span> Add a topic
+      <% end %>
+    </nav>
+  <% end %>
+</header>
+
+<%= render 'shared/tags/table',
+      resources: @topics,
+      empty_message: "No topics exist yet.
+        #{link_to('Add one', new_topic_path)}".html_safe %>
```
alext: This link should only to be shown to GDS Editors...

alext:
Some small things, otherwise this look great.  Much simpler, and easier to understand.

PR #69, opened 2015-05-28
Remove model annotations in specs

tijmenb:
Currently the `annotate` gem helpfully adds annotations to both the models in `app/models` and the specs in `spec/models`.

This is duplicate and makes for noisy commits.


KushalP:
Yup, good deleting work here :+1:

PR #71, opened 2015-05-29
Curate lists for Mainstream Browse Pages

tijmenb:
This PR adds the ability to group the links on mainstream browse pages into "curated lists".

This works identical to the current functionality for curating the list contents of a Topic:

screen shot 2015-05-29 at 14 55 26

https://www.gov.uk/browse/benefits/tax-credits

### Trello

> Allow the content shown in 2nd level mainstream browse pages to be curated, in a similar way to that in which sub-topic page content can be curated. This curation should be limited to GDS editors (but viewing of mainstream browse pages is currently restricted in this way already, so this shouldn't be hard). Curated content should be able to be put into groups, with named headings, and have its order controlled. Content should be able to appear in more than one group.

https://trello.com/c/iy1tB6gU/124-add-curation-of-mainstream-browse-pages-to-collections-publisher

### Related PRs

- https://github.com/alphagov/collections-publisher/pull/70 (Make Lists belong to Tag)
- https://github.com/alphagov/govuk-content-schemas/pull/76 (Add curated links to mainstream browse pages)

Best reviewed per-commit.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -1,8 +1,9 @@
 class TagsController < ApplicationController
+  before_filter :find_tag
+  before_filter :require_gds_editor_permissions_to_edit_browse_pages!
+
   def republish
-    tag = Tag.find_by!(content_id: params[:tag_id])
-    authorise_user!("GDS Editor") if tag.is_a?(MainstreamBrowsePage)
-    PublishingAPINotifier.send_to_publishing_api(tag)
+    PublishingAPINotifier.send_to_publishing_api(@tag)
     redirect_to :back
   end
 end
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -1,0 +1,8 @@
+class TagsController < ApplicationController
+  def republish
+    tag = Tag.find_by!(content_id: params[:tag_id])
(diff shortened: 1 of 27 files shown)


app/models/mainstream_browse_page.rb
```diff
@@ -31,7 +31,12 @@ def base_path
     "/browse#{super}"
   end
 
+  def tag_type
```
alext: Would be good to rename this method to something like `legacy_tag_type` to make it clear that it's only applicable to the old-world stuff.

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,8 @@
+class TagsController < ApplicationController
+  def republish
+    tag = Tag.find_by!(content_id: params[:tag_id])
```
alext: worth using the `find_tag` before filter here?

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,8 @@
...
+    authorise_user!("GDS Editor") if tag.is_a?(MainstreamBrowsePage)
```
alext: This could use a before_filter to be consistent with the other controllers. (it might make sense to move the `editing_lists_for_mainstream_browse_page?` method up to the application_controller as well to avoid duplication.)

PR #81, opened 2015-06-09
Improve handling of "untagged" list items

tijmenb:
The curated lists feature in this app allows users to organise links that have been tagged elsewhere into lists.

It frequently happens that after an article is added to a list, the article is untagged by somebody else. This means we have a `ListItem` record in Collections Publisher that serves no purpose anymore. This is a "normal" thing to happen.

This PR attempts to improve the layout of the section.

- Removes the separate block for the lists, as that prevents the user from seeing the item in context.
- Shows the items in the list they originally appeared on.
- Change the button from a hard "remove" button into a "confirmation" button. This makes it clear that we are not deleting anything important, but just confirming a change that has been done elsewhere.

## Before

before

## After

The original grey block is gone, and the items therein are now shown in the original list. When hovering over the new remove-button a tooltip appears explaining the situation.

screen shot 2015-06-09 at 15 27 54
screen shot 2015-06-09 at 15 27 25

## Review

The first three commits are refactors and cleanups to make the final commit - 98466df - cleaner. Review per-commit and in the browser.

## Trello

Part of improvements to the list editor, but here as a separate PR to aid reviewability.

   https://trello.com/c/VXdUyg2g/149-collections-publisher-improve-the-curated-links-editor


rboulton:
This looks great - it will want a design review, but that'll be easier on preview so I've merged it.

PR #82, opened 2015-06-10
Re-layout list editor

tijmenb:
The aim of this PR is to improve the UX of the list editor.

- The editor now has a left hand side with all items tagged to the item. On the right all lists are displayed. This hopefully improves the understanding of the relationship between the two parts.
- Items from the left hand side are only "draggable". Previously it was possible to sort and drag items in this list, causing confusion. It also allows users to add a link to multiple lists, something which
could previously only be achieved by turning off javascript.
- When an item is included in one or more lists it is bold
- When the user is about to create the first list for a tag we show a special page explaining they're about to change the "sort mode" of the tag. This can be further improved by setting the ordering explicitly - but that's for a later PR.
- The layout of the tables is updated to fit more items on the screen and make dragging easier.
- Previously, when an item was dragged we calculated the exact rows of that had to be updated. For example, when the user dragged the fifth item in a list of 10 to the fourth place, we would update items 4 through 10. Finding out which items to update turned out to be a fruitful source of bugs and confusion. This behaviour has been changed to always update all items in a list on each update. In the previous example, instead of 4-10, we update 1-10. In a later PR, we should use bulk-updating to the server (which is now not very hard to implement).
- Related, keeping data in sync is hard on JS. `showEmptyLabels()` and `redrawUncategorizedList()` are purposely "naive". They look at the state of the world and re-render on each update, instead
of trying to only update things that changed, which would be more performant.

## Before

before

## After

screen shot 2015-06-10 at 11 48 26

screen shot 2015-06-10 at 11 49 05

screen shot 2015-06-10 at 11 51 12

## Review

In a browser. Split view on GH.

## Follow ups

Suggested future improvements:

- Remove the `title` column from `list_items` in the database and only use it from the API.
- Save the list items in a bulk ajax call.
- Add a filtering box for the tagged items
- Add links to tagged items on the site and/or editor (whitehall/publisher)
- Add preview of the lists on `topics#show` and `browse_pages#show`
- Add independent scrolling for the left and right hand side to allow easier dropping.
- Set the a-to-z or alphabetical "sort modes" explicitly, instead of letting it depend on `tag.lists.any?`


rboulton:
I think it would be good to deploy this branch to preview before merging it, and get some of the people who will be using this tool to try it out.  Code review to follow.

rboulton:
This is excellent.

PR #83, opened 2015-06-10
Show previews and sort mode for tags

tijmenb:
This PR:

- Adds previews for lists on the show pages of `MainstreamBrowsePage` and `Topic`.
- Adds a little icon to all relevant pages to show wether the tag is alphabetical or curated.

Hopefully this will it easier to grasp the different "sort modes" for a tag.

Needs some wordsmithery.

## Before

screen shot 2015-06-10 at 13 20 58
screen shot 2015-06-10 at 13 21 00
screen shot 2015-06-10 at 13 21 02

## After

screen shot 2015-06-10 at 13 21 13
screen shot 2015-06-10 at 13 21 18

screen shot 2015-06-10 at 13 21 23
(curated)

screen shot 2015-06-10 at 13 22 58
(alphabetized)

### DIFF
--- a/app/assets/stylesheets/views/mainstream_browse_pages.scss
+++ b/app/assets/stylesheets/views/mainstream_browse_pages.scss
@@ -98,4 +98,17 @@ h1 .type {
   border-top: 1px solid #eee;
   padding: 10px 0 0;
   position: relative;
+
+  small i {
+    color: #2A2A2A;
+    margin-right: 3px;
+  }
+
+  ul {
+    padding-left: 0;
+
+    li {
+      list-style: none;
+    }
+  }
 }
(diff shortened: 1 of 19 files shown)


alextea:
I wonder if rather than using icons we could just add "(curated)" to the title of the curated lists?

I think this would be better as there would be less visual clutter, and it would be more obvious than the icons alone.

I'd like to hear what the content team think though as they will be using this.

tijmenb:
Updated PR after some feedback. Removed the second icon from the show page and added a tooltip that tells users whether the tag is curated or a-to-z.

Updated screenshot:

screen shot 2015-06-11 at 13 13 42

rboulton:
Code looks good. :sparkles:

PR #84, opened 2015-06-11
Add dirty state for tags

tijmenb:
When a list is updated the tag is marked as "dirty", which means it has changes that have not been published to the content store and aren't shown on the site.

Adding a label will help users to see which tags have unpublished changes - currently only the appearance of the "Publish" button would indicate the published state.


rboulton:
Sounds good - could you attach screenshots?

PR #85, opened 2015-06-11
Add "Unpublished changes" label for tags

tijmenb:
When a list is updated the tag is marked as "dirty", which means it has changes that have not been published to the content store and aren't shown on the site.

Adding a label will help users to see which tags have unpublished changes - currently only the appearance of the "Publish" button would indicate the published state.

## Before

screen shot 2015-06-11 at 13 44 54
screen shot 2015-06-11 at 13 45 58

## After

screen shot 2015-06-11 at 13 41 45
screen shot 2015-06-11 at 13 45 52

### DIFF
--- a/app/assets/stylesheets/topics.scss
+++ b/app/assets/stylesheets/topics.scss
@@ -26,3 +26,8 @@ table.topics {
     padding: 3px;
   }
 }
+
+h1 .label {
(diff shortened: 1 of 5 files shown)


rboulton:
Code looks good - merging for checking on preview.

PR #86, opened 2015-06-11
Send top and second-level pages to content-store

jackscotti:
[Trello Ticket](https://trello.com/c/D2vjglFv/110-store-full-information-needed-to-render-first-two-columns-of-a-mainstream-browse-page-in-content-store)

## This PR is for working towards:

 - removing use of `contentapi` by collections frontend.
 - Making collections frontend simpler so we can more easily experiment with putting childcare content into it in a new way.

## Details

When collections-publisher sends a mainstream browse page document to the content store, it should include the full details needed to render the first two columns of the page, so that the collections frontend doesn't need to make a call to content-api to render these columns of a mainstream browse page.

This information consists of:

 - For `/browse`, the titles, descriptions and base_paths of all the top level mainstream browse pages.

 - For `/browse/foo`, the same as for `/browse`, and also the titles, descriptions and base_paths of all the 2nd level mainstream browse pages under `foo`.

 - For `/browse/foo/bar`, the same information as for `/browse/foo`. (Also the links to topics, if any, which we're already putting in the content store.)

### DIFF
--- a/app/presenters/mainstream_browse_page_presenter.rb
+++ b/app/presenters/mainstream_browse_page_presenter.rb
@@ -9,6 +9,28 @@ private
   def links
     super.merge(
       "related_topics" => @tag.topics.order(:title).map(&:content_id),
+      "top_level_browse_page" => top_level_browse_page_id,
+      "top_level_browse_pages" => @tag.class.sorted_parents.map(&:content_id),
+      "second_level_browse_pages" => second_level_browse_pages,
     )
   end
+
+private
+
+  def top_level_browse_page_id
+    if @tag.has_parent?
+      [@tag.parent.content_id]
+    else
+      []
+    end
+  end
+
+  def second_level_browse_pages
+    if @tag.child?
+      @tag.parent.sorted_children.map(&:content_id)
+    else
+      @tag.sorted_children.map(&:content_id)
+    end
+  end
+
 end
--- a/app/presenters/mainstream_browse_page_presenter.rb
+++ b/app/presenters/mainstream_browse_page_presenter.rb
@@ -9,7 +9,7 @@ private
   def links
     super.merge(
       "related_topics" => @tag.topics.order(:title).map(&:content_id),
-      "top_level_browse_page" => top_level_browse_page_id,
+      "active_top_level_browse_page" => top_level_browse_page_id,
       "top_level_browse_pages" => @tag.class.sorted_parents.map(&:content_id),
       "second_level_browse_pages" => second_level_browse_pages,
     )
--- a/spec/presenters/mainstream_browse_page_presenter_spec.rb
+++ b/spec/presenters/mainstream_browse_page_presenter_spec.rb
@@ -76,10 +76,12 @@ RSpec.describe MainstreamBrowsePagePresenter do
         browse_page.update_attributes!(:parent => parent_browse_page)
       end
 
-      it "sends empty array without any" do
-        expect(presented_data[:links]).to eq({
-          "related_topics" => [],
-        })
+      context "without linked topics" do
+        it "returns an empty array" do
+          expect(presented_data[:links]["related_topics"]).to eq(
+            []
+          )
+        end
       end
 
       context "with some linked topics" do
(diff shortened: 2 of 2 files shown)


app/presenters/mainstream_browse_page_presenter.rb
```diff
@@ -9,6 +9,20 @@ def format
   def links
     super.merge(
       "related_topics" => @tag.topics.order(:title).map(&:content_id),
+      "top_level_browse_page" => top_level_browse_page_id,
+      "top_level_browse_pages" => @tag.class.sorted_parents.map(&:content_id),
+      "second_level_browse_pages" => @tag.class.second_level_pages.map(&:content_id),
```
alext: I don't think this is correct.  It'll return all second-level pages that exist, it needs to only return the ones that are children of the current top-level page.

spec/presenters/mainstream_browse_page_presenter_spec.rb
```diff
@@ -102,5 +108,52 @@
...
+        bravo.children = [foxtrot]
+
+        delta.topics = [golf, hotel]
+
+        alpha.save!
+        delta.save!
+        echo.save!
+      end
+
+      it "Delta has all the possible linked pages" do
```
alext: I think these tests could be structured more clearly. It's not clear what 'Delta' is here.  It would also be good to cover this for both a top-level browse page, and a second-level page.  Maybe something like:

```ruby
context "for a top-level browse page" do
  it "includes all the top-level browse pages"
  it "includes all the child pages for the top-level page"
  it "is valid against the schema"
end

context "for a second-level browse page" do
  it "includes all the top-level browse pages"
  it "includes the parent top-level browse page"
  it "includes all the second-level pages that are children of the parent page"
  it "is valid against the schema"
end
```

spec/presenters/mainstream_browse_page_presenter_spec.rb
```diff
@@ -69,6 +69,12 @@
       expect(presented_data).to be_valid_against_schema('mainstream_browse_page')
     end
 
+    it "does not have topics" do
+      expect(presented_data[:links]["related_topics"]).to eq(
+        [],
+      )
+    end
```
alext: Why is this test added?

spec/presenters/mainstream_browse_page_presenter_spec.rb
jackscotti: Thanks, was going to do that ASA I was going to get the good to go.

app/presenters/mainstream_browse_page_presenter.rb
jackscotti: You're right. I misunderstood the requirements.

jackscotti:
@alext can you please give another look at this?

Are you happy with the newer naming convention used in the tests? I think it helps making the hierarchical structure clear.

spec/presenters/mainstream_browse_page_presenter_spec.rb
```diff
@@ -92,15 +94,96 @@
...
+      let(:top_level_page_1) { create(:mainstream_browse_page) }
+      let(:top_level_page_2) { create(:mainstream_browse_page) }
+
+      let(:second_level_page_1) { create(:mainstream_browse_page) }
+      let(:second_level_page_2) { create(:mainstream_browse_page) }
+      let(:second_level_page_3) { create(:mainstream_browse_page) }
+
+      before :each do
+        # Top level pages must have children
+        top_level_page_1.children = [second_level_page_1, second_level_page_2]
```
alext: Instead of doing this, you could assign the parents as you create the second-level pages.  ie:
```ruby
let(:second_level_page_1) { create(:mainstream_browse_page, :parent => top_level_page_1) }
```

That way you'd avoid the need for this before block entirely.  If there are issues with the lazy loading of `let` blocks, you can use `let!` instead.

alext:
The structure of the tests is really clear now.  Other than one small comment about the test setup, this looks good to me.

tijmenb:
a0a1eed fixes this branch after this change: https://github.com/alphagov/govuk-content-schemas/commit/dae4c8f3f27191465ef6adbe37f28d6dcf520df1

spec/presenters/mainstream_browse_page_presenter_spec.rb
```diff
@@ -92,15 +94,96 @@
...
+        end
+
+        it "includes all its sibling second-level pages" do
+          expect(presented_data[:links]["second_level_browse_pages"]).to eq([
+            second_level_page_1.content_id,
+            second_level_page_2.content_id,
+          ])
+        end
+
+        it "is valid against the schema" do
```
tijmenb: Also needs a `schema_test: true` tag

app/presenters/mainstream_browse_page_presenter.rb
```diff
@@ -9,6 +9,28 @@ def format
...
     )
   end
+
+private
+
+  def top_level_browse_page_id
+    if @tag.has_parent?
+      [@tag.parent.content_id]
+    else
+      []
```
alext: Did we not want to populate `active_top_level_browse_page` for a top-level page?

rboulton: Ah, yes, good point.

jackscotti: Fixed here: https://github.com/alphagov/collections-publisher/pull/87

PR #88, opened 2015-06-16
Edit #active_top_level_browse_page_id

jackscotti:
The method should never return an empty array. When @tag
is the active_top_level_browse_page, it should return its
own id.


spec/presenters/mainstream_browse_page_presenter_spec.rb
```diff
@@ -122,7 +122,7 @@
 
         it "it does not have a parent browse pages" do
```
alext: This description is no longer correct.

jackscotti: Updated.

PR #89, opened 2015-06-16
Send root `browse` page data to content store

jackscotti:
## This PR completes: 
- the work started in this PR: https://github.com/alphagov/collections-publisher/pull/86
- this [ticket](https://trello.com/c/D2vjglFv/110-store-full-information-needed-to-render-first-two-columns-of-a-mainstream-browse-page-in-content-store)

# Working towards:

 - removing use of `contentapi` by collections frontend.

 - Making collections frontend simpler so we can more easily experiment with putting childcare content into it in a new way.

# Details

When collections-publisher sends a mainstream browse page document to the content store, it should include the full details needed to render the first two columns of the page, so that the collections frontend doesn't need to make a call to content-api to render these columns of a mainstream browse page.

This information is stored in entries in the "links" field, and consists of:

 - For `/browse`: the list of all the top level mainstream browse pages.

## Browse page available in dev:

screen shot 2015-06-16 at 15 12 52

### DIFF
--- a/app/presenters/root_browse_page_presenter.rb
+++ b/app/presenters/root_browse_page_presenter.rb
@@ -1,0 +1,30 @@
+class RootBrowsePagePresenter
+
+  def render_for_publishing_api
+    {
+      content_id: "8413047e-570a-448b-b8cb-d288a12807dd",
+      format: "mainstream_browse_page",
+      title: "Browse",
+      locale: 'en',
+      public_updated_at: Time.now.iso8601,
+      publishing_app: "collections-publisher",
+      rendering_app: "collections",
+      routes: routes,
+      update_type: "major",
+      links: links,
+    }
+  end
+
+private
+
+  def routes
+    [ {path: "/browse", type: "exact"} ]
+  end
+
+  def links
+    {
+      "top_level_browse_pages" => MainstreamBrowsePage.sorted_parents.map(&:content_id)
+    }
+  end
+
+end
--- a/app/presenters/root_browse_page_presenter.rb
+++ b/app/presenters/root_browse_page_presenter.rb
@@ -6,7 +6,7 @@ class RootBrowsePagePresenter
       format: "mainstream_browse_page",
       title: "Browse",
       locale: 'en',
-      public_updated_at: Time.now.iso8601,
+      public_updated_at: public_updated_at,
       publishing_app: "collections-publisher",
       rendering_app: "collections",
       routes: routes,
@@ -17,13 +17,28 @@ class RootBrowsePagePresenter
 
 private
 
+  def top_level_browse_pages
+    MainstreamBrowsePage.sorted_parents
+  end
+
+  def public_updated_at
+    if links["top_level_browse_pages"].empty?
+      raise "Top-level pages Array can't be empty"
+    else
+      top_level_browse_pages
+        .map(&:updated_at)
+        .max
+        .iso8601
+    end
+  end
+
   def routes
     [ {path: "/browse", type: "exact"} ]
   end
 
   def links
     {
-      "top_level_browse_pages" => MainstreamBrowsePage.sorted_parents.map(&:content_id)
+      "top_level_browse_pages" => top_level_browse_pages.map(&:content_id)
     }
   end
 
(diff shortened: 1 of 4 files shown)


spec/presenters/root_browse_page_presenter_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      top_level_page_1 = create(:mainstream_browse_page)
+      top_level_page_2 = create(:mainstream_browse_page)
+
+      rendered = RootBrowsePagePresenter.new.render_for_publishing_api
+
+      expect(rendered).to be_valid_against_schema('mainstream_browse_page')
+      expect(rendered[:links]["top_level_browse_pages"]).to eq([
+        top_level_page_1.content_id,
+        top_level_page_2.content_id,
+      ])
```
alext: These 2 assertions could be split into separate tests wrapped in a context block.  The schema one could then be tagged as a schema test so that it gets included in that test run.

app/presenters/root_browse_page_presenter.rb
```diff
@@ -0,0 +1,30 @@
+class RootBrowsePagePresenter
+
+  def render_for_publishing_api
+    {
+      content_id: "8413047e-570a-448b-b8cb-d288a12807dd",
+      format: "mainstream_browse_page",
+      title: "Browse",
+      locale: 'en',
+      public_updated_at: Time.now.iso8601,
```
rboulton: I'm uncomfortable with this - republishing a page shouldn't change its public_updated_at value.

I think an ideal value to use here would be the most recent time that a top-level mainstream browse page was updated.

jackscotti: I understand, and now it make a lot of sense, thanks!

What do you think if I add a method that runs through the top-level pages, finds the most recent and sticks its updated_at time in here? 
I know this is not going to happen in a real world scenario but, what should the method return in case of total absence of top-level pages? `Time.now.iso8601`, `nil` ?

alext: I think `Time.now.utc` is a reasonable default in that case.  The content-store requires this field to be set, so we can't send `nil`.

rboulton: It's not possible for that to happen, because there's always at least one top level page (the page that's being published!).  I'd suggest raising an error in this case, rather than trying to handle it.

spec/presenters/root_browse_page_presenter_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        RootBrowsePagePresenter.new.render_for_publishing_api
+      }.to raise_error(RuntimeError)
+    end
+
+    context "with top level browse pages" do
+
+      let!(:top_level_page_1) { create(:mainstream_browse_page)}
+      let!(:top_level_page_2) { create(:mainstream_browse_page)}
+
+      setup do
```
tijmenb: We only need this setup block for `it "RootBrowsePagePresenter#updated_at == most recent #updated_time" do`, maybe it's clearer if we move it in there.

tijmenb: Also not that this is rspec - the `setup` block is actually not run here. This masks a bug here:

https://github.com/alphagov/collections-publisher/pull/89/files#diff-9ceec5e0ecaaf0a70e37d3077c9b5b80R30

app/presenters/root_browse_page_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+    MainstreamBrowsePage.sorted_parents
+  end
+
+  def public_updated_at
+    if links["top_level_browse_pages"].empty?
+      raise "Top-level pages Array can't be empty"
+    else
+      top_level_browse_pages
+        .map(&:updated_at)
+        .min
```
tijmenb: Should be `.max` - the latest/highest `updated_at`.

spec/presenters/root_browse_page_presenter_spec.rb
jackscotti: Why is this not run? :disappointed:

alext: because this is RSpec, it needs to be `before :each do`

PR #92, opened 2015-06-17
Generate CSV for router-data redirects

tijmenb:
Adds a rake task to output a CSV with redirect instructions, to feed to the `router-data` repo. It will redirect all topic pages to their new location at `/topic`.

For example:

```
/business-tax to /business-tax
```

And subtopics:

```
/business-tax/air-passenger-duty to /topic/business-tax/air-passenger-duty
/business-tax/air-passenger-duty/email-signup to /topic/business-tax/air-passenger-duty/email-signup
/business-tax/air-passenger-duty/latest to /topic/business-tax/air-passenger-duty/latest
```

Part of Trello: https://trello.com/c/EoWPp4qZ/187

### DIFF
--- a/lib/tasks/topic_redirects.rake
+++ b/lib/tasks/topic_redirects.rake
@@ -1,0 +1,16 @@
+desc "Generate a CSV for router-data redirects"
+task :topic_redirects => [:environment] do
+  topic_base_paths = Topic.published.only_parents.map(&:base_path).sort
+
+  topic_base_paths.each do |base_path|
+    puts "#{base_path},/topic#{base_path},exact"
+  end
+
+  subtopic_base_paths = Topic.published.only_children.map(&:base_path).sort
+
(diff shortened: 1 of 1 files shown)


alext:
Why are we doing this through router-data as opposed to sending redirect items to the content-store?

rboulton:
I thought router-data was still the canonical place to record redirects, and the content store was an optional extra.  Perhaps I've missed something in this chain getting more mature.

@bradleywright How are we meant to be setting up redirects these days?

tijmenb:
Closing this, I agree with @alext that adding redirects to the content store is better. Part of that logic can then also be reused to allow slug changes.

bradleywright:
Router data is only for things that don't fully integrate with the content-store.

rboulton:
Thanks for catching this @alext - my misunderstanding.

PR #94, opened 2015-06-18
Move topics to /topic namespace

tijmenb:
This PR makes collections-publisher publish Topics to the `/topic` namespace.

- Publish Topics under `/topic`. Done by prepending `/topic` to the base path of Topics.
- Sets up redirects for the old URLs. This is done by publishing "redirect" content items for the version without `/topic`.

Best reviewed per-commit and in split view.

Trello: https://trello.com/c/EoWPp4qZ/187-move-topics-to-topic

After this is deployed, we need to republish all the items.

### Screenshots

To help understand what is going on, we show the redirects for a topic if it has any:

screen shot 2015-06-18 at 14 37 47
screen shot 2015-06-18 at 14 37 53

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -39,6 +39,8 @@ class Tag < ActiveRecord::Base
   has_many :lists
   has_many :list_items, through: :lists
 
+  has_many :redirects
+
   validates :slug, :title, :content_id, presence: true
   validates :slug, uniqueness: { scope: ["parent_id"] }, format: { with: /\A[a-z0-9-]*\z/ }
   validate :parent_is_not_a_child
@@ -98,11 +100,6 @@ class Tag < ActiveRecord::Base
     end
   end
 
-  def base_path
-    base = has_parent? ? "/#{parent.slug}" : ''
-    "#{base}/#{slug}"
-  end
-
   def to_param
     content_id
   end
@@ -125,18 +122,13 @@ class Tag < ActiveRecord::Base
   def list_items_from_contentapi
     @_list_items_from_contentapi ||= begin
       CollectionsPublisher.services(:content_api)
-        .with_tag(panopticon_slug, legacy_tag_type, draft: true)
+        .with_tag(full_slug, legacy_tag_type, draft: true)
         .map { |content_blob|
           ListItem.new(title: content_blob.title, api_url: content_blob.id)
         }
     rescue GdsApi::HTTPNotFound
       []
     end
-  end
-
-  # FIXME: remove this once we're using content_id's in URLs everywhere.
-  def panopticon_slug
-    self.base_path.gsub('/browse', '')[1..-1]
   end
 
   def legacy_tag_type
@@ -154,6 +146,10 @@ class Tag < ActiveRecord::Base
     else
       @_display_curated_links
     end
+  end
+
+  def full_slug
+    @full_slug ||= [parent.try(:slug), slug].compact.join('/')
   end
 
 private
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -26,7 +26,13 @@
 class Topic < Tag
   has_many :mainstream_browse_pages, through: :reverse_tag_associations, source: :from_tag
 
+  alias subtopic? has_parent?
+
   def legacy_tag_type
     'specialist_sector'
+  end
+
+  def base_path
+    "/topic/#{full_slug}"
   end
 end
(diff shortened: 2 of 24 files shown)


alext:
I haven't reviewed yet, but before this is merged, we need to update the collections app to be able to serve these pages on both the old and the new paths.

db/migrate/20150618112012_add_old_route_to_tags.rb
```diff
@@ -0,0 +1,10 @@
+class AddOldRouteToTags < ActiveRecord::Migration
+  def up
+    add_column :tags, :has_legacy_route_without_namespace, :boolean
```
rboulton: I think this would be better disallowing NULL and with a default of false, to avoid a tri-state boolean.

app/models/topic.rb
```diff
@@ -25,7 +26,43 @@
...
+
+  # Redirects are currently hard-coded: if this topic existed before the move
+  # to the /topic namespace, we add a redirect from the old path (/oil-and-gas)
+  # to the namespaced version (/topic/oil-and-gas).
+  # For subtopics we also redirect two subroutes:
+  # /:topic/:subtopic/email-signup and /:topic/:subtopic/latest
+  def redirect_to_topic_namespace
+    return {} unless has_legacy_route_without_namespace?
+
+    old_path = base_path.gsub('/topic', '')
```
alext: Worth an anchor in this regex.  We wouldn't want this to hit any false-positives (I'm imagining a topic with a slug of "topical-events" or similar).

Also, this only needs to be `sub` not `gsub` as it should only be making a single replacement.

db/migrate/20150618112012_add_old_route_to_tags.rb
```diff
@@ -0,0 +1,10 @@
...
+    execute 'UPDATE tags SET has_legacy_route_without_namespace = true WHERE state = "published"'
```
alext: This update needs to be constrained to tags of type Topic.  We don't want to set this for browse pages.

spec/support/content_store_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+  def stub_content_store!
+    @content_store ||= FakeContentStore.new
+    allow(CollectionsPublisher).to receive(:services)
+      .with(:publishing_api)
+      .and_return(@content_store)
+
+    @content_store
+  end
+
+  alias :content_store :stub_content_store!
```
alext: Would `stubbed_content_store` be a better name for this alias to make it clear in the tests what's being referred to.

Also, the way this alias is used, it does the `allow(CollectionsPublisher)...` every time it's accessed.  Would be good to avoid that if possible.  Maybe make this a simple accessor instead of an alias.

app/models/tag.rb
```diff
@@ -130,7 +132,7 @@ def list_items_from_contentapi
 
   # FIXME: remove this once we're using content_id's in URLs everywhere.
   def panopticon_slug
-    self.base_path.gsub('/browse', '')[1..-1]
+    self.base_path.gsub('/browse', '').gsub('/topic', '')[1..-1]
```
alext: These `sub`'s need anchoring.  I also wonder whether this is worth moving to the 2 subclasses to make it clearer (avoid having to do both `sub`s)

app/notifiers/publishing_api_notifier.rb
```diff
@@ -1,13 +1,52 @@
...
+          destination: redirect.to_base_path
+        }
+      end
+
+      publishing_api.put_content_item(old_path, {
+        format: 'redirect',
+        publishing_app: 'collections-publisher',
+        update_type: 'major',
+        redirects: redirect_routes
+      })
```
alext: I wonder whether some of this belongs in a presenter?

db/migrate/20150622073727_add_redirects_for_topics_namespace.rb
```diff
@@ -0,0 +1,52 @@
+class AddRedirectsForTopicsNamespace < ActiveRecord::Migration
+  def up
+    Topic.published.only_parents.each do |topic|
+      original_topic_base_path = original_topic_base_path(topic)
```
alext: I think using the same name for the method and the local variable here makes things a little unclear.  It forces me to think about ruby's scoping rules when reading the code below.

db/migrate/20150622073727_add_redirects_for_topics_namespace.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+private
+
+  def original_topic_base_path(topic)
+    if topic.has_parent?
+      "/#{topic.parent.slug}/#{topic.slug}"
+    else
+      "/#{topic.slug}"
+    end
```
alext: This could now be simplified to use `topic.full_slug`.

alext:
This looks good to me now.

This needs the corresponding frontend changed to be deployed before this is merged though.

PR #96, opened 2015-06-23
Batch send mainstream browse page

bishboria:
Mainstream browse pages now contain entries in their links field for all the top-level browse pages, and all their sibling second-level browse pages. Whenever these lists of links are changed, these need to be updated.

https://trello.com/c/v1B0aJxW/185-update-content-store-when-list-of-browse-pages-change


tijmenb:
:+1: looks very good

app/notifiers/publishing_api_notifier.rb
```diff
@@ -1,4 +1,14 @@
+require 'sidekiq'
+
 class PublishingAPINotifier
+  def self.batch_send_to_publishing_api(tag)
+    if tag.child?
+      parent_with_children(tag).each { |item| queue_publishing_change(item.id) }
+    else
+      full_hierarchy.each { |item| queue_publishing_change(item.id) }
+    end
```
alext: This feels like knowledge that shouldn't be in this class.  I think this would be better refactored to be something like:
```ruby
def self.send_to_publishing_api(tag)
  send_to_publishing_api(tag)
  tag.dependent_tags.each { |dependent_tag| queue_publishing_change(dependent_tag.id) }
end
```
The existing `send_to_publishing_api` method can become a private method, and be renamed to something like `send_single_tag_to_publishing_api`.

That way, the knowledge of which tags depend on each other can be kept within the models, and this class is solely responsible for actually sending to the publishing-api.


This approach has the advantage that sending the tag itself remains synchronous, meaning that any errors will be reported to the user, and additionally, sending of dependent tags will only be queued if sending this tag succeeds.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -10,4 +20,26 @@ def self.send_to_publishing_api(tag)
...
+  end
+
+  private
+
+  def self.queue_publishing_change(thing)
+    QueueWorker.perform_async(thing)
+  end
+
+  def self.full_hierarchy
+    MainstreamBrowsePage.only_parents.map(&:children).flatten + MainstreamBrowsePage.only_parents
```
alext: This doesn't look correct.  How will it behave when given a topic tag?

alext:
It would be good to see something in an integration test that covers this.

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,8 @@
+# This file is overwritten on deploy
+
+Sidekiq.configure_client do |config|
+  config.redis = { :namespace => 'collections-publisher' }
```
alext: It would be good to update this to read the redis config from `config/redis.yml`.  See publisher for example.

spec/notifiers/publishing_api_notifier_spec.rb
```diff
@@ -1,10 +1,61 @@
 require "rails_helper"
 
 RSpec.describe PublishingAPINotifier do
+  include GdsApi::TestHelpers::PublishingApi
```
alext: This isn't used anywhere below any more.

alext:
The sidekiq addition also needs a `Procfile` added to the route with a `worker:` line in it.  That's what's used by the init scripts on the servers to run the worker.  See publisher for an example.

config/initializers/sidekiq.rb
bishboria: done

spec/notifiers/publishing_api_notifier_spec.rb
bishboria: done

bishboria:
`Procfile` added too.

PR #98, opened 2015-06-23
Enable republishing of dirty tags

tijmenb:
This PR enables us to publish tags to the content-store, without also publishing the curated lists.

Trello:

> Currently, when a topic or mainstream browse page is curated, but the changes have not yet been published, we cannot republish the page without also publishing the changes.
> So that we don't have to republish changes in this case, record the contents of the `details` hash field sent to publishing-api in a column in the Tags table.  When republishing, just used the stored value.  When publishing changes to the curated lists, first update the stored value, and then send it.
> We'll populate the initial values in the columns by reading from the content store (and we'll need to be careful to ensure that this doesn't conflict with any publish actions which happen at the same time).
> This is a blocker for adding the /topic/ namespace, and also for the batch republishing of browse pages when new browse pages are created.

https://trello.com/c/lBeQH9bz/218-enable-republishing-of-dirty-tags

Note that we have chosen to save the `groups` array from the `details` hash, because the details hash also contains the `beta` label. Including the entire `details` hash would expose us to the exact same problem when changing the beta-label.

One time task to run: `rake populate_published_groups`

Review per-commit.


db/schema.rb
```diff
@@ -43,17 +43,18 @@
...
     t.datetime "updated_at"
-    t.string   "content_id",  limit: 255,                 null: false
-    t.string   "state",       limit: 255,                 null: false
-    t.boolean  "dirty",       limit: 1,   default: false, null: false
-    t.boolean  "beta",        limit: 1,   default: false
+    t.string   "content_id",       limit: 255,                   null: false
+    t.string   "state",            limit: 255,                   null: false
+    t.boolean  "dirty",            limit: 1,     default: false, null: false
+    t.boolean  "beta",             limit: 1,     default: false
+    t.text     "published_groups", limit: 65535
```
alext: Is 64k big enough to cover all cases?

app/services/list_publisher.rb
```diff
@@ -0,0 +1,13 @@
+class ListPublisher
+  attr_reader :tag
+
+  def initialize(tag)
+    @tag = tag
+  end
+
+  def perform
+    groups = TopicPresenter.new(tag).build_groups
```
alext: This needs to use `TagPresenter.for(tag)` so that it will also support mainstream browse pages.

lib/tasks/populate_published_groups.rake
```diff
@@ -0,0 +1,18 @@
+desc 'One-time task to populate Tag#published_groups'
+task populate_published_groups: [:environment] do
+  Tag.only_children.each do |tag|
+    item = CollectionsPublisher.services(:content_store).content_item(tag.base_path)
+    unless item
+      puts "#{tag.base_path} is missing from the content-store."
+      next
+    end
+
+    tag.update!(published_groups: item['details']['groups'])
```
alext: It would be good for this to handle the case where the content item doesn't have a groups key.  We should still be populating `published_groups` with empty array in this case.

lib/tasks/populate_published_groups.rake
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  Tag.only_children.where(dirty: false).each do |tag|
+    actual_groups = TopicPresenter.new(tag).build_groups.map(&:stringify_keys)
+    published_groups = tag.published_groups
+    raise "Non-dirty tag #{tag.id} is actually dirty."
```
alext: Instead of raising here, it's might be worth creating a list of tags that are dirty, and then raising an error with the full list.

app/models/tag.rb
```diff
@@ -49,6 +50,13 @@ class Tag < ActiveRecord::Base
   scope :only_children, -> { where('parent_id IS NOT NULL') }
   scope :in_alphabetical_order, -> { order('title ASC') }
 
+  # The links last sent to the content-store.
+  serialize :published_groups, JSON
+
+  after_initialize do
+    self.published_groups ||= []
+  end
```
alext: Instead of using `after_initialize` here (which can get quite expensive), we could set the default value in the schema to be the serialized representation of empty array.

alext:
A few small points, otherwise this looks good to me.

lib/tasks/populate_published_groups.rake
tijmenb: We want to record whatever has been published, so nil would be the correct value here.

lib/tasks/populate_published_groups.rake
tijmenb: Good point, but the code here is more of a written proof that the content is in a consistent state than an actual test / error reporting mechanism.

app/services/list_publisher.rb
tijmenb: Will fix this by extracting a `GroupsPresenter`, since the `build_groups` method doesn't have anything to do with `TagPresenter` anymore.

PR #99, opened 2015-06-23
Include topic children in links

alext:
So that we can render the topic pages entirely from the content-store.


alext:
Tests will fail until https://github.com/alphagov/govuk-content-schemas/pull/88 is merged.

PR #100, opened 2015-06-24
Remove data migration that have been run

tijmenb:
These data migrations have been deployed to production and have run. They can be safely removed.

Removing them prevents problems where updates to the code base require updates to never-used migrations.


db/migrate/20150608145958_remove_unused_topic_record.rb
```diff
@@ -1,36 +0,0 @@
...
-      "publishing_app" => "collections-publisher",
-      "update_type" => "major",
-      "redirects" => [
-        {
-          "path" => "/immigration-operational-guidance/european-casework-instructions",
-          "type" => "exact",
-          "destination" => "/immigration-operational-guidance",
-        },
-      ],
-    }
```
alext: I'm not sure we should remove this migration just yet.  It's the only place that this redirect is recorded.  In the future we will likely support archiving/renaming tags, at which point this can be modled in the database.

PR #104, opened 2015-06-25
Add Sidekiq monitoring

tijmenb:
This commit exposes the sidekiq-web monitoring app on `/sidekiq`.

Sidekiq was added in https://github.com/alphagov/collections-publisher/pull/96. Users are authenticated via the normal GDS::SSO mechanism. They need the `Sidekiq Monitoring` permission to see the app.

Related to Trello: https://trello.com/c/v1B0aJxW/185

Note that `/sidekiq` is currently blocked by nginx.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,8 @@ gem 'select2-rails', '~> 3.5.9'
 
 gem 'unicorn', '~> 4.9.0'
 
+# sidekiq-web depends on sinatra
+gem 'sinatra', require: nil
 gem 'sidekiq', '~> 2.17.2'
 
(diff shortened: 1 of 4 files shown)


bishboria:
:+1:

bradleywright:
We have this: https://github.com/alphagov/sidekiq-monitoring, which is a central place to do what you're doing here.

alext:
@bradleywright As I recall, we went that route because we couldn't get this to work behind signon auth.  This PR does put the monitoring behind signon auth, and therefore may be a better approach.

This might be worth a discussion offline...

PR #109, opened 2015-06-26
Send topics to rummager after update

tijmenb:
Send topics to rummager after update. Currently, the panopticon app sends data to Rummager for indexing.

Collections publisher owns the documents, so is the natural place to do this.

_This is a re-commit of 6dc69bc, which was reverted in 28c4d08 because it was committed straight to master._

### DIFF
--- a/config/initializers/services.rb
+++ b/config/initializers/services.rb
@@ -17,7 +17,7 @@ module CollectionsPublisher
   class ServiceNotRegisteredException < Exception; end
 end
 
-# Only used by rake task `populate_published_groups` 
+# Only used by rake task `populate_published_groups`
 require 'gds_api/content_store'
 CollectionsPublisher.services(:content_store, GdsApi::ContentStore.new(Plek.new.find('content-store')))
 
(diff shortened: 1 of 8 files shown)


config/initializers/services.rb
```diff
@@ -32,3 +32,6 @@ class ServiceNotRegisteredException < Exception; end
   :panopticon,
   GdsApi::Panopticon.new(Plek.new.find('panopticon'),
                          bearer_token: ENV['PANOPTICON_BEARER_TOKEN'] || 'example'))
+
+require 'gds_api/rummager'
+CollectionsPublisher.services(:rummager, GdsApi::Rummager.new(Plek.new.find('search')))
```
alext: The plek call should be `Plek.new.find('rummager')`.  When the app was renamed, the search vhost was left in place to avoid breaking everything, but should be considered deprecated.

alext:
:+1:  Looks good.

We should hold off merging this until the corresponding panopticon PR is ready to go at the same time.  Otherwise we'll end up with duplicated results in the search index.

rboulton:
:+1:

PR #112, opened 2015-06-29
Send topic slug to rummager.

alext:
This will enable us to avoid the mess here:
https://github.com/alphagov/rummager/blob/5ac805b/lib/registry.rb#L93-L107


rboulton:
Good call.

alext:
Upcoming PR on rummager will fix the services & information pages...

PR #113, opened 2015-07-02
Bump GDS API adapters to 20.1.1

elliotcm:
This changes the user agent to include the app
name in API requests.

https://trello.com/c/qkHdtob4/247-bump-gds-api-adapters-everywhere-it-s-used-to-include-user-agent-information

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'plek', '~> 1.10.0'
 gem 'airbrake', '~> 4.2.0'
 
 gem 'gds-sso', '~> 11.0.0'
-gem 'gds-api-adapters', '~> 18.10.0'
+gem 'gds-api-adapters', '20.1.1'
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -7,7 +7,7 @@ gem 'plek', '~> 1.10.0'
 gem 'airbrake', '~> 4.2.0'
 
 gem 'gds-sso', '~> 11.0.0'
-gem 'gds-api-adapters', '~> 18.10.0'
+gem 'gds-api-adapters', '~> 20.1.0'
```
alext: Your commit says 20.1.1, but this is 20.1.0.  Probably worth pinning to 20.1.1 given 20.1.0 has a bug.

elliotcm: It's in the lock file, but fair enough.  Done.

PR #115, opened 2015-07-07
Allow publishers to reorder mainstream pages

jackscotti:
This PR is the first step towards allowing publishers to re-order mainstream pages.

Added a page where publishers can select the type of child ordering of a page (alphabetical or curated).
In the same page, the user will be able to input the order in which s/he wants to display them (in case of curated list).

Next step: When updating the child ordering, collections-publisher should send to content-store the parent page and the children in the publisher's wanted order.

NB: Cosmetic work will follow. This is absolutely not what it should and will look like.

Ticket: https://trello.com/c/Tpgjz0HI/167-collections-publisher-can-control-order-of-mainstream-browse-columns

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -15,6 +15,8 @@
 #  dirty            :boolean          default(FALSE), not null
 #  beta             :boolean          default(FALSE)
 #  published_groups :text(16777215)
+#  child_ordering   :string(255)      default("alphabetical"), not null
+#  index            :integer          default(0), not null
 #
 # Indexes
 #
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -52,7 +52,8 @@ class Tag < ActiveRecord::Base
 
   scope :only_parents, -> { where('parent_id IS NULL') }
   scope :only_children, -> { where('parent_id IS NOT NULL') }
-  scope :in_alphabetical_order, -> { order('title ASC') }
+  scope :in_alphabetical_order, -> { order(:title) }
+  scope :in_curated_order, -> { order(:index) }
 
   # The links last sent to the content-store.
   serialize :published_groups, JSON
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -92,7 +92,11 @@ class Tag < ActiveRecord::Base
   end
 
   def sorted_children
-    children.sort_by(&:title)
+    if self.child_ordering == "alphabetical"
+      children.in_alphabetical_order
+    else
+      children.in_curated_order
+    end
   end
 
   def title_including_parent
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -59,7 +59,7 @@ class Tag < ActiveRecord::Base
   serialize :published_groups, JSON
 
   # after_initialize is expensive, but MySQL doesn't support default values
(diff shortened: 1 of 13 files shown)


app/models/tag.rb
```diff
@@ -45,12 +48,14 @@ class Tag < ActiveRecord::Base
   validates :slug, uniqueness: { scope: ["parent_id"] }, format: { with: /\A[a-z0-9-]*\z/ }
   validate :parent_is_not_a_child
   validate :cannot_change_slug
+  validate :ordering_type_must_be_valid
```
alext: You don't need to use a custom validator here, you can do:
```ruby
validates :ordering_type, inclusion: {in: ORDERING_TYPES}
```

app/views/mainstream_browse_pages/manage_child_ordering.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= tag_header @browse_page, 'Manage child ordering' %>
+
+<% if @browse_page.sorted_children %>
+  <ul>
+    <%= form_for @browse_page do |f| %>
+      <%= f.select :child_ordering, Tag::ORDERING_TYPES %>
+      </br>
+      <%= f.fields_for :children, @browse_page.children do |c| %>
```
alext: It would be good to use `sorted_children` here...

app/views/mainstream_browse_pages/show.html.erb
```diff
@@ -2,6 +2,11 @@
   <%= link_to "Edit page",
     edit_mainstream_browse_page_path(@browse_page) %>
 
+  <% if @browse_page.has_children? %>
```
alext: We probably want this to be `@browse_page.can_have_children?`.  We might want to be able to set the `child_ordering` field, even when no children have yet been created.

app/views/mainstream_browse_pages/manage_child_ordering.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= tag_header @browse_page, 'Manage child ordering' %>
+
+<% if @browse_page.sorted_children %>
```
alext: I don't think we need this conditional.

app/views/mainstream_browse_pages/manage_child_ordering.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  <ul>
+    <%= form_for @browse_page do |f| %>
+      <%= f.select :child_ordering, Tag::ORDERING_TYPES %>
+      </br>
+      <%= f.fields_for :children, @browse_page.children do |c| %>
+        <%= c.text_field :index, label: c.object.slug %>
+      <% end %>
+      <%= f.submit 'Save', class: 'btn-submit' %>
+    <% end %>
+  </ul>
```
alext: This `<ul>` doesn't contain any list items.

app/models/tag.rb
alext: Also, there are no unit tests around this.  Would be good to add some.

jackscotti: that's great, did not know that. Will do.

jackscotti:
@alext , I should have addressed all. Thank you for the good feedback.

PR #116, opened 2015-07-09
Test the order in which child pages are rendered

jackscotti:
Review by commit advised.

Adds tests around this work:
https://github.com/alphagov/collections-publisher/pull/115
Completes:
https://trello.com/c/Tpgjz0HI/167-collections-publisher-can-control-order-of-mainstream-browse-columns

### DIFF
--- a/spec/presenters/mainstream_browse_page_presenter_spec.rb
+++ b/spec/presenters/mainstream_browse_page_presenter_spec.rb
@@ -118,19 +118,19 @@ RSpec.describe MainstreamBrowsePagePresenter do
         :title => "Top-level page 2",
       )}
 
-      let!(:second_level_page_1) { create(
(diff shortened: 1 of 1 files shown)


alext:
With the tests setup the way they currently are, if the presenter just output the 2nd-level pages in database order, I think the tests would still pass.  I think it would be good to re-order the creation of the various 2nd-level pages to ensure these tests can't accidentally pass.

jackscotti:
Yep, I did exactly the opposite of what I was supposed to do. 
Now A and B that are supposed to be displayed alphabetically are created in non alphabetical order.
C and D that are supposed to be displayed in non-alphabetical order are now created in alphabetical order. These changes ensure that these tests can't accidentally pass.

PR #121, opened 2015-07-14
Fix N+1 on topic and browse page index pages

tijmenb:
A N+1 query was introduced recently because for each topic/browse page on the index page, we are making a query for the subtopics with either alphabetical or curated ordering. This caused a fair amount of extra database traffic.

To remedy this, this commit moves the sorting of the parents to ruby-land by using `sort_by`. This way all `children` for all tags are eager-loaded and we just have to execute one query.

Before: ActiveRecord: 159.5ms
After: ActiveRecord: 7.5ms

Not that [removing the .includes in _children.html.erb](https://github.com/alphagov/collections-publisher/commit/815dedf70821073f21a5902e36d29f34f4a38161#diff-21926eabfcc0366fe8005ef12ca6d26eL17) introduces a new N+1 query when loading the show page for mainstream browse pages with associated topics. These pages are less frequently accessed and cause less queries (@alext).

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -54,8 +54,6 @@ class Tag < ActiveRecord::Base
 
   scope :only_parents, -> { where('parent_id IS NULL') }
   scope :only_children, -> { where('parent_id IS NOT NULL') }
-  scope :in_alphabetical_order, -> { order(:title) }
-  scope :in_curated_order, -> { order(:index) }
 
   # The links last sent to the content-store.
   serialize :published_groups, JSON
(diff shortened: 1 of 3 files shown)


app/models/tag.rb
```diff
@@ -94,10 +92,10 @@ def self.sorted_parents
   end
 
   def sorted_children
-    if self.child_ordering == "alphabetical"
-      children.in_alphabetical_order
+    if child_ordering == "alphabetical"
+      children.sort_by(&:title)
```
alext: This will now be a case-sensitive sort.  I'm not sure if that matters or not.

rboulton: This is not desirable - we have had examples (eg, "IR35") where the ordering was wrong due to a case sensitive sort being used.

PR #122, opened 2015-07-14
Import router-data redirects for browse pages and topics

alext:
Now that we can model these here it makes sense to do so and reduce the volume of redirects in router-data.

This PR also adds a uniqueness constraint to the Redirects table to help avoid duplicates.

### DIFF
--- a/app/models/redirect.rb
+++ b/app/models/redirect.rb
@@ -12,7 +12,8 @@
 #
 # Indexes
 #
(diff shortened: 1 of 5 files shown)


alext:
This should be reviewed alongside https://github.gds/gds/router-data/pull/450.

Additionally https://github.gds/gds/router-data/pull/444 also removed some redirects that are included here.

PR #123, opened 2015-07-16
Delete unused mainstream browse pages

jackscotti:
Two childcare mainstream browse pages were created,
but then the content was updated and they're no longer needed.
They were never published, and never should be,
but need deleting to tidy things up.

Ticket:
https://trello.com/c/drh7LqiI/249-delete-two-unused-childcare-mainstream-browse-page-tags


alext:
You need to run the migration and include `db/schema.rb` in the PR (the serial number will be updated).

jackscotti:
Gotcha, thanks!

PR #125, opened 2015-07-23
Set descriptions for topics

tijmenb:
The frontend app is [currently generating fallback descriptions for topics](https://github.com/alphagov/frontend/blob/2b64f4989fcb7089d8e2c7255c78170ee7c624da/app/presenters/search_result.rb#L65). This migration adds the same description to the database, so that we don't need special code in `frontend` or `rummager`.

We need to run `rake rummager:index_topics` after deploying this change.

## Before

screen shot 2015-07-23 at 12 23 36

## After

screen shot 2015-07-23 at 12 23 29

Trello: https://trello.com/c/AM4LQ2i3

### DIFF
--- a/app/views/topics/edit.html.erb
+++ b/app/views/topics/edit.html.erb
@@ -4,6 +4,8 @@
   <%= f.text_field :slug, :disabled => true %>
   <%= f.text_field :title %>
   <%= f.text_field :description %>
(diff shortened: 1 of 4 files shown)


alext:
This needs collections to be deployed before it can be merged.

Otherwise looks good to me.

PR #127, opened 2015-07-23
Send browse pages to Rummager

tijmenb:
We'd like to use the mainstream browse pages in rummager. This PR makes collections-publisher send the content of browse pages to rummager in the same way as happened with topics.

Trello: https://trello.com/c/6vXKBd5d

PR on rummager to then hide the browse pages from search results: https://github.com/alphagov/rummager/pull/471

### DIFF
--- a/spec/controllers/mainstream_browse_pages_controller_spec.rb
+++ b/spec/controllers/mainstream_browse_pages_controller_spec.rb
@@ -46,8 +46,9 @@ RSpec.describe MainstreamBrowsePagesController do
       create(:mainstream_browse_page)
     }
 
-    it 'notifies panopticon' do
+    it 'notifies panopticon and rummager' do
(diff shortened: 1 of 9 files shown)


spec/controllers/mainstream_browse_pages_controller_spec.rb
```diff
@@ -48,6 +48,7 @@
 
     it 'notifies panopticon' do
```
alext: This description needs to be updated.

tijmenb: updated

tijmenb:
Branch updated after offline conversation with @alext. Instead of separate classes for the Rummager payload, we re-use the `TagPresenter` and subclasses.

alext:
:+1: This looks good to me.

alphagov/rummager#471 should be deployed to prod before this is merged.

PR #128, opened 2015-07-28
Fetch tagged documents from rummager

tijmenb:
Currently this application fetches documents tagged to a topic or browse page from the content-api.

This commit changes that to request the tagged documents from rummager instead. This removes our dependency on content-api completely.

Trello: https://trello.com/c/K4JU8Ovz/263-collections-publisher-reads-content-from-rummager


spec/support/rummager_helper.rb
```diff
@@ -0,0 +1,17 @@
+module RummagerHelper
+  def stub_rummager_linked_content_call
+    stub_request(:get, %r[https://rummager.test.gov.uk/.]).
```
alext: It would be better to use `Plek.find('rummager')` here to avoid hardcoding this domain etc.

alext: Nitpick: Also, this regex could use the `.`s escaping, and an anchor at the start.

alext:
Looks good to me.

We should hold off merging until we've validated that the data in rummager is complete.

rboulton:
We've validated that the data in rummager is *NOT* complete, so I'm closing this for now to avoid accidental merging.  We'll definitely want to do this at some point soon though, once we've fixed rummager (and we have the trello ticket linking to this PR to remind us).

alext:
Re-opened.  The data in rummager for topics is now complete (pending some work to put topics for manuals in).  There are 13 browse items that are missing from rummager, but since the only curated browse pages are the childcare ones (which don't include the missing documents), this is not a problem.

Merging this will allow the HMRC manuals to be added to curated topics.

PR #80, opened 2015-06-08
Migration to remove unused subtopic

binaryberry:
All content from this subtopic has been moved to other pages. This
migration removes the Topic associated with the slug
immigration-operational-guidance/european-casework-instructions

cc @alext @rboulton 

Pivotal Tracker: https://www.pivotaltracker.com/story/show/91699292

PR #130, opened 2015-08-06
Redirect old topic to new location

tijmenb:
This redirect was first created in PR #80 (fd69069).

However, the original owner wants a new location for the URL.

Trello: https://trello.com/c/Bw0pk6dI/277-add-redirect-for-european-casework-instructions
Zendesk: https://govuk.zendesk.com/agent/tickets/972973


alext:
Is there not a corresponding redirect record in the DB?  If not, we should create one.

The original PR was created before we were storing these redirects here, and the redirect was therefore also captured in router-data.

PR #132, opened 2015-08-07
Rake task to relocate detailed guides to new urls

boffbowsh:
For all published detailed guides, finds any List Items that point to
it and prefix the `base_path` with `/guidance` due to the relocation of
all detailed guides from the root namespace.

Once this has been completed, it iterates over all `Topic`s and
republishes their lists to the Publishing API via Sidekiq.

This relies on all Detailed Guides being reslugged in Rummager first,
as the `GroupsPresenter` used by `ListPresenter` relies on the content
also being tagged correctly in Rummager.

### DIFF
--- a/lib/tasks/detailed_guides.rake
+++ b/lib/tasks/detailed_guides.rake
@@ -1,0 +1,32 @@
+namespace :detailed_guides do
+  desc "Relocate detailed guides referenced in Topics"
+  task :relocate_to_guidance => :environment do
+    panopticon = CollectionsPublisher.services(:panopticon)
+
+    page = 1
+    while true
+      puts "Fetching page #{page}"
+      response = panopticon.get_json("#{panopticon.send(:base_url)}.json?kind=detailed_guide&state=published&page=#{page}")
+      if response.count == 0
+        break
+      else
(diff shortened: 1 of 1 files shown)


lib/tasks/detailed_guides.rake
```diff
@@ -0,0 +1,32 @@
+namespace :detailed_guides do
+  desc "Relocate detailed guides referenced in Topics"
+  task :relocate_to_guidance => :environment do
+    panopticon = CollectionsPublisher.services(:panopticon)
+
+    page = 1
+    while true
```
alext: This pagination stuff should be unnecessary.  There's already pagination support built into gds-api-adapters (See [`ListResponse#with_subsequent_pages`](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/list_response.rb#L79))

boffbowsh: If I use `get_list!` instead of `get_json`, I get this error:

```
NoMethodError: undefined method `grouped_results' for #<Array:0x007f173c1960b8>
/var/govuk/collections-publisher/lib/tasks/detailed_guides.rake:8:in `each'
```

Which I guess means the request I'm making isn't compatible with `get_list!`?

PR #137, opened 2015-09-01
Fix the comparison task

rboulton:
The task for comparing the results from rummager and content_api for
listing documents by tag was broken, because the content_api service has
been removed from the app everywhere.

It's still needed by the comparison task, so import and configure it
locally.

### DIFF
--- a/lib/tasks/compare.rake
+++ b/lib/tasks/compare.rake
@@ -3,6 +3,8 @@ namespace :comparison do
   desc "report on differences in items tagged as reported by rummager and contentapi"
   task :run => :environment do
     require 'csv'
+    require 'gds_api/content_api'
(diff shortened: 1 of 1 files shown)


lib/tasks/compare.rake
```diff
@@ -3,6 +3,8 @@ namespace :comparison do
   desc "report on differences in items tagged as reported by rummager and contentapi"
   task :run => :environment do
     require 'csv'
+    require 'gds_api/content_api'
+    CollectionsPublisher.services(:content_api, GdsApi::ContentApi.new(Plek.new.find('content_api')))
```
alext: I think that'll need to be `Plek.new.find('contentapi')` (without the underscore...).  The argument to Plek becomes the hostname in the returned URL.

PR #138, opened 2015-09-04
Add functionality to remove a topic

tijmenb:
We're regularly getting requests to delete (old) topics. This is time consuming work, because we have to do this manually and in data migrations. For example: https://trello.com/c/ybdnGMhl, https://trello.com/c/dTPgvWYL and https://trello.com/c/7UEYsTr8. It is also error prone, as evidenced by https://github.com/alphagov/collections-publisher/pull/130 and https://github.com/alphagov/collections-publisher/pull/131.

This PR adds a `TagArchiver` class that archives tags by:

1. Setting a `archived` boolean to indicate that the topic is no longer in use
2. Adding a redirect in the content-store
3. Removing the topic from rummager

For draft tags, there's a `DraftTagRemover` class that _removes_ tags by:

1. Removing it from the database
2. Adding a "gone" item in the content-store

## Screenshots

Draft subtopics have a "Remove" button:

screen shot 2015-09-07 at 15 50 38

And can be removed by the user, who has to confirm with a javascript-confirm dialog.

screen shot 2015-09-07 at 15 51 27

After which it is removed from the database.

---

Published topics have a "Archive topic" button:
 
screen shot 2015-09-07 at 15 56 30

Which leads to the page where the user has to select a successor to redirect to:

screen shot 2015-09-07 at 16 00 06

Submitting this will cause the page to redirect and the topic to be archived.

screen shot 2015-09-07 at 16 01 05

### Discussion

- [x] Should we add a panopticon endpoint that removes the tag, so it can't be used for tagging? Code already exists [in a Rake task] ((https://github.com/alphagov/panopticon/blob/master/lib/tasks/specialist_sector_cleanup.rake). _(answer: yes)_
- [x] How do we deal with existing redirects? For example, most topics have redirects from their own non-namespaced variant to the `/topic/` namespace. These are not being redirected currently. _(answer: that's OK)_
- [x] Can we automate the untagging of documents, so we don't need the check for `tag.tagged_documents.any?` _(answer: not right now)_
- [x] Do we treat draft topics differently? Do they still need redirects? _(answer: don't redirect)_
- [x] How do we prevent updates to these topics? Must we do more to prevent republishing?

### Todo

- [x] Add panopticon endpoint to remove empty tag
- [x] Do this in a transaction
- [x] Take care of draft topics
- [x] Prevent topic updates
- [x] Add redirects for suffix-routes
- [x] Add UI for removing the topic

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,7 +87,7 @@ GEM
       railties (>= 3.0.0)
     faraday (0.9.1)
       multipart-post (>= 1.2, < 3)
-    gds-api-adapters (20.1.1)
+    gds-api-adapters (24.3.0)
       link_header
       lrucache (~> 0.1.1)
       null_logger
@@ -298,7 +298,7 @@ DEPENDENCIES
   capybara (~> 2.4.1)
   database_cleaner (~> 1.3.0)
   factory_girl_rails (~> 4.5.0)
-  gds-api-adapters (= 20.1.1)
+  gds-api-adapters (= 24.3.0)
   gds-sso (~> 11.0.0)
   generic_form_builder (~> 0.9.0)
   govuk-content-schema-test-helpers (~> 1.3)
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -91,6 +91,7 @@ class Tag < ActiveRecord::Base
   end
 
   alias child? has_parent?
+  alias parent? can_have_children?
 
   def self.sorted_parents
     only_parents.includes(children: [:lists]).order(:title)
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -67,6 +67,11 @@ class Tag < ActiveRecord::Base
   aasm column: :state, no_direct_assignment: true do
     state :draft, initial: true
     state :published
+    state :archived
+
+    event :move_to_archive do
+      transitions from: :published, to: :archived
+    end
 
     event :publish do
       transitions from: :draft, to: :published
--- a/app/notifiers/publishing_api_notifier.rb
+++ b/app/notifiers/publishing_api_notifier.rb
@@ -13,7 +13,10 @@ class PublishingAPINotifier
 
   def send_single_tag_to_publishing_api
     if tag.published?
-      publishing_api.put_content_item(presenter.base_path, presenter.render_for_publishing_api)
+      unless tag.archived?
+        publishing_api.put_content_item(presenter.base_path, presenter.render_for_publishing_api)
+      end
+
       add_redirects
     else
       publishing_api.put_draft_content_item(presenter.base_path, presenter.render_for_publishing_api)
--- a/app/notifiers/publishing_api_notifier.rb
+++ b/app/notifiers/publishing_api_notifier.rb
@@ -39,7 +39,7 @@ private
   end
 
   def publishing_api
-    @publishing_api ||= CollectionsPublisher.services(:publishing_api)
+    Services.publishing_api
   end
 
   def self.publish_root_page(tag)
@@ -65,14 +65,14 @@ private
   class RootBrowsePageWorker
     include Sidekiq::Worker
     def perform
-      CollectionsPublisher.services(:publishing_api).put_content_item("/browse", RootBrowsePagePresenter.new.render_for_publishing_api)
+      Services.publishing_api.put_content_item("/browse", RootBrowsePagePresenter.new.render_for_publishing_api)
     end
   end
 
   class RootTopicWorker
     include Sidekiq::Worker
     def perform
-      CollectionsPublisher.services(:publishing_api).put_content_item("/topic", RootTopicPresenter.new.render_for_publishing_api)
+      Services.publishing_api.put_content_item("/topic", RootTopicPresenter.new.render_for_publishing_api)
     end
   end
 end
--- a/app/notifiers/publishing_api_notifier.rb
+++ b/app/notifiers/publishing_api_notifier.rb
@@ -13,12 +13,11 @@ class PublishingAPINotifier
 
   def send_single_tag_to_publishing_api
     if tag.published?
-      unless tag.archived?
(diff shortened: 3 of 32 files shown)


app/services/tag_archiver.rb
```diff
@@ -0,0 +1,44 @@
...
+  def setup_redirects
+    # The redirect will live at the original path of the tag, so the original
+    # item in the content store will be replaced by it. The parent topic will
+    # no longer expand the item in the `links/children` field because this
+    # item will be of the type redirect.
+    tag.redirects.create!(
+      original_tag_base_path: tag.base_path,
+      from_base_path: tag.base_path,
+      to_base_path: successor.base_path,
+    )
```
alext: This doesn't account for the extra routes that topics can have (latest feed, and email-signup)

alext:
Thoughts around the questions:

> Should we add a panopticon endpoint that removes the tag, so it can't be used for tagging? Code already exists in a Rake task.

This could be something to do.  We'd need to do a similar thing for Whitehall as well I think.

> How do we deal with existing redirects? For example, most topics have redirects from their own non-namespaced variant to the /topic/ namespace. These are not being redirected currently.

These will end up as double-redirects. I don't think that's a terrible thing.  We could dereference these, but that would make it harder if we decide to allow republishing an archived tag.

> Can we automate the untagging of documents, so we don't need the check for tag.tagged_documents.any?

It might be possible to do this, but with all the upcoming changes to how tagging information is stored, this is quite a moving target.

> Do we treat draft topics differently? Do they still need redirects?

I don't think it's worth adding redirects for these. They've never been live on the site, therefore the URLs won't be in use anywhere. We should probably send a draft gone item for them so that they get removed from the draft site.

> How do we prevent updates to these topics? Must we do more to prevent republishing?

At least initially, I think we should prevent them from being republished. If there's a need to allow this, we should implement that feature fully (including removing the corresponding redirect etc...).

rboulton:
I think we should add an endpoint for deleting a topic from panopticon.  To keep things simple, this endpoint could return an error if there are any documents tagged to the topic (automatically untagging them all is fiddly right now, because for some formats other apps are the 'source of truth' for the tagging).

If I remember right, we don't need an equivalent thing for whitehall, because whitehall gets the list of all available tags from contentapi (it only stores taggings, not tags).

Agree with all @alext's other answers.  Basically, focus on making it possible to delete an unused topic.

tijmenb:
Thanks, updated the description with new checklist.

app/services/tag_archiver.rb
tijmenb: Addressed in https://github.com/alphagov/collections-publisher/commit/ecfcdbcabd0708f5cbcd4316ddea8cc77e2777cb

tijmenb:
@alext @rboulton this is now ready for full review. 

I've not done anything with panopticon here because that would introduce a deploy dependency. For now we have to remember to delete a panopticon tag before/after archiving the tag here.

rboulton:
I think I'd like to see this not be merged before the panopticon stuff is merged too: I don't want to deploy this archive button without the panopticon side being done too, because we're too likely to end up with "orphaned" stuff in panopticon.  Since there's a pretty good PR for panopticon, I don't think this should end up being a blocker.

tijmenb:
Updated with the extra call to remove the tag from panopticon: https://github.com/alphagov/collections-publisher/commit/3329ca793f135075b82df41bef93f2051db0543a

rboulton:
This looks good to me: I'd like @alext to take a look at it though.

Gemfile.lock
```diff
@@ -318,6 +318,3 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   unicorn (~> 4.9.0)
   webmock (~> 1.21.0)
-
-BUNDLED WITH
-   1.10.6
```
alext: I don't think you meant to commit this change... (collections is now on ruby 2.2.3, which uses bundler 1.10.6 in our stack)

app/services/draft_tag_remover.rb
```diff
@@ -0,0 +1,36 @@
...
+  def initialize(tag)
+    @tag = tag
+  end
+
+  def remove
+    return if tag.published? || tag.parent? || tag.tagged_documents.any?
+
+    remove_tag_from_panopticon
+    add_gone_item
+    tag.destroy!
```
alext: Worth a transaction here for consistency?

tijmenb: Oh definitely, thanks.

app/models/tag.rb
```diff
@@ -17,6 +17,7 @@
 #  published_groups :text(16777215)
 #  child_ordering   :string(255)      default("alphabetical"), not null
 #  index            :integer          default(0), not null
+#  archived         :boolean          default(FALSE), not null
```
alext: I'm not convinced this is the best way to model this.  We already have a state column, and a state machine. I think we should add archived as an additional state for this. Otherwise we'd end up with tags that are both published and archived at the same time.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -13,7 +13,10 @@ def initialize(tag)
 
   def send_single_tag_to_publishing_api
     if tag.published?
-      publishing_api.put_content_item(presenter.base_path, presenter.render_for_publishing_api)
+      unless tag.archived?
```
alext: This logic would probably be clearer as a case statement now that there are more than 2 possible states.

Gemfile.lock
tijmenb: Removed & rebased. Still a mystery what removed this.

tijmenb:
@alext replaced the data migration with the state machine and added a commit that makes it use that.

app/services/draft_tag_remover.rb
tijmenb: Added.

tijmenb:
@alext updated to use the exception from here: https://github.com/alphagov/gds-api-adapters/pull/353

PR #140, opened 2015-09-07
Wrap publishing in a transaction

tijmenb:
During the publishing of a tag, we currently update the database and then try to send the document to the external services. When a downstream service like whitehall fails this will result in the item being `published?`, but the item will not have be sent to external services.

Calling `publish` on a published item will then result in `AASM:​:Inva​lidTr​ansit​ion: ​Event​ 'pub​lish'​ cann​ot tr​ansit​ion f​rom '​publi​shed'` exceptions. This has happened in production a couple of times, and happens often in development because a dependent app isn't running.

screen shot 2015-09-07 at 15 29 18


rboulton:
Thanks, this is great.

PR #145, opened 2015-09-24
Republish all tags to panopticon

tijmenb:
The current rake task only republishes published tags. Tags that are `draft` (but not `archived`) have entries in panopticon too, so there's no reason for this behaviour. It was added here: https://github.com/alphagov/collections-publisher/commit/d7db8d31d0fc3f5c4a5c5b7b1df6e0697966c76e

Trello: https://trello.com/c/IgYwT6yM

### DIFF
--- a/lib/tasks/panopticon.rake
+++ b/lib/tasks/panopticon.rake
@@ -1,7 +1,7 @@
 namespace :panopticon do
   desc "Republish all tags to panopticon"
   task :republish_tags => [:environment] do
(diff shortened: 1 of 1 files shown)


rboulton:
Does this correctly pass the state of the tag to panopticon?  Is there a need for a test about that?

tijmenb:
`PanopticonNotifier#update_tag` is used by the controllers for draft and published alike, so I'm assuming that works as excpected.

PR #146, opened 2015-09-29
Bump govuk_admin_template to 3.0.0 for GA fix

boffbowsh:
Trello: https://trello.com/c/1dS5tgTg/125-fix-analytics-in-all-16-admin-tools

Ensures the Analytics domain is set to the 
`GOVUK_APP_DOMAIN` env variable rather than 
being hardcoded.


Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'airbrake', '~> 4.2.0'
 gem 'gds-sso', '~> 11.0.0'
 gem 'gds-api-adapters', '24.3.0'
 
-gem 'govuk_admin_template', '~> 2.3.1'
+gem 'govuk_admin_template', '3.0.0'
```
tijmenb: Can you maintain the pessimistic versioning here?

boffbowsh: Almost every other app has the version pinned for this gem.

tijmenb: Almost every other app is wrong. 

https://github.com/alphagov/styleguides/blob/bebfb81551f900dc6be623f508f2b19fa8032ce1/ruby.md#gemfile

PR #148, opened 2015-10-01
Eager load Services to fix circular dependency error

tijmenb:
We've gotten a few errors on production where Rails can't find the `Services` module and crashes with a circular dependency error.

The "Circular dependency detected while autoloading constant Services" is raised when Rails thinks it will encounter an infinite loop. In this case, when first seeing the constant `Services`, it loads a file named
`services.rb` to load the `Services` module.

However, if Rails has already loaded a file named `services.rb` and the constant is still missing, it assumes that there is a circular dependency.

My guess in this case is that Rails is incorrect in thinking it has already loaded _our_ `services.rb`, but is trying to load some other file. By explicitly loading our `services.rb` on boot time, we skip the
autoloading step and avoid the error.

It's still worth investigating why this error occurs, this mitigation is a temporary step.

Trello: https://trello.com/c/aBe4IMSt

### DIFF
--- a/config/initializers/eager_load.rb
+++ b/config/initializers/eager_load.rb
@@ -1,0 +1,2 @@
+# Eager load Services to prevent intermittent circular dependency errors.
+require 'services'


jamiecobbett:
Needs a rebase.

We have a similar problem that I tried to fix in Whitehall https://github.com/alphagov/whitehall/pull/2303 and then reverted: https://github.com/alphagov/whitehall/pull/2318

tijmenb:
Ah! Thanks. I'm closing this to find a structural fix then.

jamiecobbett:
I think this is a perfectly sensible pragmatic solution as-is, though there might be a systematic fix.

tijmenb:
Hmm, can't reopen this PR. Probably because I rebased and force-pushed it.

PR #154, opened 2015-10-06
Allow users to redirect archived topics to non-topic pages

jackscotti:
Currently, topics which are archived can only be redirected to topics / subtopics.  It would sometimes be more appropriate to just redirect to a piece of content e.g. the subtopic 'export refunds' seemed to best redirect to the guidance 'Export refunds'.

This work makes possible to redirect topics to any piece of content (on GOV.UK) when archiving them.

Notes from content designer:
- OK to use 'URL' as long as we clarify that it needs to be the relative path. (This is the pattern used elsewhere eg the short URL tool.)

How the updated form will look like:

screen shot 2015-10-06 at 14 47 02

This work already went through Design and Content review.
Ticket: https://trello.com/b/Kg3JL5Cd/finding-things


app/controllers/topics_controller.rb
```diff
@@ -61,17 +61,29 @@ def propose_archive
...
+        TagArchiver.new(topic, successor).archive
+        redirect_to topic_path(topic), notice: 'The topic has been archived.'
+      else
+        DraftTagRemover.new(topic).remove
+        redirect_to topics_path, notice: 'The topic has been removed.'
+      end
+    rescue GdsApi::HTTPConflict
+      flash[:error] = "The tag could not be deleted because there are documents tagged to it"
+      redirect_to :back
+    rescue GdsApi::HTTPServerError
```
jackscotti: This is too general and needs to be replaced by something more specific.

jackscotti:
@bishboria that would make sense too but what would happen if a user would select a topic and input a path too? This approach (2 forms) has been suggested and approved by @alextea 🌈

tijmenb:
To recap offline discussion: this all looks good, but we're waiting to see if we can catch a more specific error when publishing-api rejects the content item.

jackscotti:
Yup! Exactly.

KushalP:
CI appears to have got stuck here?

tijmenb:
@KushalP force-pushed, CI ran again.

app/views/topics/propose_archive.html.erb
```diff
@@ -1,10 +1,24 @@
 <%= tag_header @archival.tag, 'Archive topic' %>
```
rboulton: In the PR description you say that the content designers said: "'Delete' rather than 'Archive' because we won't keep a copy of the topic page".  Should this (and other visible buttons, etc) be updated to say Delete, therefore?

app/validators/valid_govuk_path_validator.rb
```diff
@@ -0,0 +1,8 @@
+class ValidGovukPathValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    raise URI::InvalidURIError unless value.starts_with?('/')
+    Services.content_store.content_item!(value)
+  rescue GdsApi::ContentStore::ItemNotFound, URI::InvalidURIError
+    record.errors[attribute] << "This URL doesn't exist on GOV.UK."
```
rboulton: This message isn't quite right - things like attachments exist on GOV.UK, though we don't want to allow redirecting to them.  Perhaps "This URL isn't a valid target for a redirect on GOV.UK"?

rboulton:
A couple of copy concerns (an error message, and that you're saying "Archive" instead of "Delete").

app/views/topics/propose_archive.html.erb
jackscotti: Archive is the correct word, the description needs to be updated. We're not really deleting them because they're still accessible from the publisher UI.

jackscotti: Updated.

jackscotti:
@rboulton addressed your comments and force-pushed.

PR #157, opened 2015-11-09
Migrate to normalised table for redirects

tijmenb:
A "redirect" can have multiple redirect routes. For example, the "redirect" for `/animal-welfare/pets` has the routes:

`/animal-welfare/pets` -> `/topic/animal-welfare/pets`
`/animal-welfare/pets/email-signup` -> `/topic/animal-welfare/pets/email-signup`

Currently the `redirects` holds both the redirect item as well as the redirect routes. Because we want to add `content_id` to the redirects (but not to the routes), we need to normalise the table.

We do the migration by using a different `table_name` for the old redirect table and copying the data. In a later PR, the `redirects` table will be removed and `newest_redirects` will be renamed to
`redirects`.

Trello: https://trello.com/c/0pcDOcMd

As follow-up PR to add content_ids to the redirects will be up after this is merged.


rboulton:
I think the example in the PR description is misleading - top level topics don't have email-signup pages currently, and /animal-welfare actually redirects to a detailed guide (it was added as a topic after the namespace changed to /topic/*)

Other than that, the intention described in this PR sounds good - I've not yet looked at the code.

tijmenb:
@rboulton You're right - I've amended the description!

PR #158, opened 2015-11-10
Add content_id to redirects

tijmenb:
Redirects need their own `content_id` so that they can live in the publishing-api.

We also add `replaced_content_item` to the item. This was added in alphagov/govuk-content-schemas#142. It allows publishing-api to add special behaviour for putting back redirects.

This change was made easy by https://github.com/alphagov/collections-publisher/pull/157.

Trello: https://trello.com/c/0pcDOcMd

/cc @elliotcm

### DIFF
--- a/app/models/redirect.rb
+++ b/app/models/redirect.rb
@@ -3,4 +3,12 @@ class Redirect < ActiveRecord::Base
 
   belongs_to :tag
   has_many :redirect_routes
+
+  before_save :ensure_content_id_is_present
+
+private
+
+  def ensure_content_id_is_present
+    self.content_id ||= SecureRandom.uuid
+  end
 end
--- a/db/migrate/20151110150858_add_content_id_to_redirect.rb
+++ b/db/migrate/20151110150858_add_content_id_to_redirect.rb
@@ -1,0 +1,16 @@
+class AddContentIdToRedirect < ActiveRecord::Migration
+  def up
+    add_column :newest_redirects, :content_id, :string
+
+    Redirect.find_each do |redirect|
+      redirect.update_column :content_id, SecureRandom.uuid
+    end
+
+    change_column_null :newest_redirects, :content_id, false
+    add_index :newest_redirects, :content_id, unique: true
+  end
+
+  def down
+    remove_column :newest_redirects, :content_id
+  end
+end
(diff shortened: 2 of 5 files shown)


app/models/redirect.rb
```diff
@@ -3,4 +3,12 @@ class Redirect < ActiveRecord::Base
 
   belongs_to :tag
   has_many :redirect_routes
+
+  before_save :add_content_id
```
elliotcm: `ensure_content_id_is_present` or similar might be more accurate.

db/migrate/20151110150858_add_content_id_to_redirect.rb
```diff
@@ -0,0 +1,12 @@
+class AddContentIdToRedirect < ActiveRecord::Migration
+  def up
+    add_column :newest_redirects, :content_id, :string
+    execute 'UPDATE newest_redirects SET content_id = UUID()'
```
elliotcm: We can't use SQL `UUID()` for this because the command is replicated down to the secondary servers in the cluster, and they generate their own UUIDs which are different to the primary (and to each other).

We need to generate these in Ruby so that they all get sent along statically through the replication tree.

db/migrate/20151110150858_add_content_id_to_redirect.rb
```diff
@@ -0,0 +1,12 @@
...
+    change_column :newest_redirects, :content_id, :string, null: false
```
elliotcm: You can use `change_column_null` for this to provide a clearer indication of intent: http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null

db/migrate/20151110152255_make_tag_id_not_null.rb
```diff
@@ -0,0 +1,5 @@
+class MakeTagIdNotNull < ActiveRecord::Migration
+  def change
+    change_column :redirects, :tag_id, :integer, null: false
```
elliotcm: Likewise can use `change_column_null`: http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null

elliotcm:
:+1: A few small changes needed/requested.

tijmenb:
Thanks for that feedback! `change_column_null` is nice. 

Updated branch.

PR #163, opened 2015-11-26
Remove Redirect model, create association between Tag and RedirectRoute

stephen-richards:
Trello: https://trello.com/c/yG9xsgvX/447-new-architecture-for-tag-redirects

This is to fix a bug which caused a duplicate key violation when a tag was archived.
The old code attempted to create a redirect item with the same base path as the content item
that was being archived.  This code will write a redirect item with the same content id, thereby
replacing the old content_item.

Summary of changes:
- removal of Redirect model
- creation of direct has_many redirect_routes association on Tag
- the addition of a redirect will write details of the redirects in the content item, not as a sparate content item
- RedirectPresenter renamed as ArchivedTagPresenter to better describe its new function
- mark archived children as such in the GUI


spec/presenters/archived_tag_presenter_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'rails_helper'
+
+RSpec.describe ArchivedTagPresenter do
+  
+  let(:parent)   { FactoryGirl.create :topic,
```
tijmenb: You don't need `FactoryGirl.` here.

spec/presenters/archived_tag_presenter_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+                                      slug: 'parent',
```
tijmenb: We generally don't align the indentation, as they make for noisy diffs.

spec/presenters/archived_tag_presenter_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      ]
+    }
+  }
+
+ 
+
+
+  describe 'render_for_publishing_api' do
+    before(:each) do
+      %w{ child-1 child-1/latest child-1/email-signup }.each do |route|
```
tijmenb: If this block is only used here, I'd prefer it to be inlined into the `it` block. Would that make it more easily readable?

spec/presenters/archived_tag_presenter_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+          type:         "exact"
+        },
+      ]
+    }
+  }
+
+ 
+
+
+  describe 'render_for_publishing_api' do
```
tijmenb: Use `#render_for_publishing_api` for instance methods,`.render_for_publishing_api` for class methods.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -12,31 +12,17 @@ def initialize(tag)
...
-      publishing_api.put_content_item(
-        redirect_presenter.base_path,
-        redirect_presenter.render_for_publishing_api
-      )
-    end
-  end
-
   def presenter
-    @presenter ||= TagPresenter.presenter_for(tag)
+    @presenter ||= tag.archived? ? ArchivedTagPresenter.new(tag) : TagPresenter.presenter_for(tag)
```
tijmenb: Could move to `TagPresenter.presenter_for`

PR #165, opened 2015-11-30
Use V2 for the publishing-api to write content to the content store

stephen-richards:
This change modifies collections-publisher to use V2 of the publishing API (see
https://gov-uk.atlassian.net/wiki/display/TECH/Publishing+Platform?focusedCommentId=42303704#comment-42303704 for details).

The usage of the API is abstracted by use of the gds-api-adapters gem.  The main differences between v1 and v2 are:
 - a content_id is the main identifier of content being written to the content store, rather than base path
 - a simple put_content will write the content item to the draft content store
 - a subsequent publish is required to write the cotnetn item to the live content store
 - archiving a topic is achieved by replacing the existing topic with a redirect item with the same content_id, and then publishing it.


Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'plek', '~> 1.11.0'
 gem 'airbrake', '~> 4.3.1'
 
 gem 'gds-sso', '~> 11.0.0'
-gem 'gds-api-adapters', '~> 24.4.0'
+gem 'gds-api-adapters', '~> 25.1'
```
jamiecobbett: We made some [breaking changes and improvements](https://github.com/alphagov/gds-api-adapters/pull/393) to the test helper in 26.0.0, so its probably worth upgrading to that instead.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -1,22 +1,26 @@
...
   end
 
+  def self.publish(tag)
+    new(tag).write_content
+    tag.dependent_tags.each { |item|
+      QueueWorker.perform_async(item.id)
+    }
+    publish_root_page(tag)
+  end
+
```
stephen-richards: change to one-liner or do..end

app/presenters/tag_presenter.rb
```diff
@@ -31,8 +40,13 @@ def render_for_rummager
     }
   end
 
+  def update_previous_version(gdsapi_response)
```
tijmenb: This isn't used anywhere.

app/notifiers/publishing_api_notifier.rb
```diff
@@ -11,12 +11,8 @@ def initialize(tag)
     @tag = tag
   end
 
-  def send_single_tag_to_publishing_api
-    if tag.draft?
-      publishing_api.put_draft_content_item(presenter.base_path, presenter.render_for_publishing_api)
-    else
-      publishing_api.put_content_item(presenter.base_path, presenter.render_for_publishing_api)
-    end
+  def write_content
```
tijmenb: Should we keep the name `send_single_tag_to_publishing_api` here?

app/presenters/archived_tag_presenter.rb
```diff
@@ -3,18 +3,34 @@ def initialize(tag)
     @tag = tag
   end
 
+  def draft?
```
tijmenb: Perhaps we could use `delegate` like in the `TagPresenter` for the methods that call `@tag`.

spec/features/mainstream_browse_pages_spec.rb
```diff
@@ -126,11 +142,17 @@
...
-    assert_publishing_api_put_item('/browse/citizenship', {
+    assert_publishing_api_put_item(citizenship.content_id, {
       "title" => "Citizenship in the UK",
       "format" => "mainstream_browse_page",
     })
 
+    # And the links should have been sent to publishing api
+    assert_publishing_api_put_links(citizenship.content_id)
+
+    # but it should have been published
```
tijmenb: This should probably read `and it should`

tijmenb:
:+1: looking good. Few minor comments left. Can you add some context to the commit message and PR?

app/notifiers/publishing_api_notifier.rb
stephen-richards: I don't think that adds anything here - it's called in the pattern:
PublishingAPINotifier.new(tag).write content, or as part of PublishingAPINotifier.send_to_publishing_api(tag), where it's obviosu that it's doing other stuff too like dependent tags and root topics/pages.

app/presenters/archived_tag_presenter.rb
stephen-richards: I'm easy, but the reason I didn't is because you need to defing a tag method (or attr_reader) for this to work.

spec/features/mainstream_browse_pages_spec.rb
stephen-richards: Done

app/presenters/tag_presenter.rb
stephen-richards: removed

tijmenb:
:+1: :fireworks:

PR #173, opened 2015-12-18
Use stubbing from gds-api-adapters

tijmenb:
Many stubs used here are not necessary anymore since gds-api-adapters has learned new methods:

https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb

This will make it easier for developers to switch between projects.


tijmenb:
cc @stephen-richards

PR #181, opened 2016-02-05
Tidy up: Removes unused tables and column

Davidslv:
Removes table redirects
Removes related_tag_id column from redirect_items
Removes newest_redirects table

### DIFF
--- a/db/migrate/20160205141721_remove_newest_redirects_table.rb
+++ b/db/migrate/20160205141721_remove_newest_redirects_table.rb
@@ -1,0 +1,9 @@
+class RemoveNewestRedirectsTable < ActiveRecord::Migration
+  def up
+    drop_table :newest_redirects
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration, "newest_redirect table can't be recreated"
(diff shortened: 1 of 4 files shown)


db/migrate/20160205141721_remove_newest_redirects_table.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveNewestRedirectsTable < ActiveRecord::Migration
+  def up
+    drop_table :newest_redirects
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration, "newest_redirect table can't be recreated"
```
Davidslv: I would like to know if this is standard way of doing this kind of migrations here.
I'm happy to change to whatever it should be.

benlovell: This looks good to me and consistent with the general rails guidance. :+1: 

Although your second argument should pluralise the table name.

Davidslv: Thanks, will correct it

PR #182, opened 2016-02-08
Downgrade Airbrake

tijmenb:
4.3.5 looks compatible with Sidekiq 3.0. We can't yet upgrade to Airbrake 5.X, because it requires a `project_id` to be set, which errbit doesn't seem to provide.

@mobaig


Gemfile.lock
```diff
@@ -361,3 +361,4 @@ DEPENDENCIES
 
 BUNDLED WITH
    1.10.6
+   1.11.2
```
mobaig: bit of weirdness here.

PR #184, opened 2016-02-09
Use "minor" update type for tag publishing

tijmenb:
Setting the update type to minor means we won't send email alerts for changes to a topic title or description. This will become relevant if topics will be tagged to other topics (we have no plans to do this, but we need to be careful).

Since there is only ever one update type, all methods have been replaced by a constant.

The only change is in the `RedirectItemPresenter` which previously was sent with a "republish" update type. This was incorrect, as the app is publishing new data, [not republishing old data](https://github.com/alphagov/publishing-api/blob/d5f63c13497147d7f366ef3c02e0439c4cf18300/doc/semantics-of-the-publishing-api.md#update_type).

Related: https://github.com/alphagov/content-tagger/pull/25
Trello: https://trello.com/c/GRt7NSbL

### DIFF
--- a/app/forms/taxon_form.rb
+++ b/app/forms/taxon_form.rb
@@ -36,7 +36,7 @@ class TaxonForm
       ]
     )
 
-    Services.publishing_api.publish(content_id, "major")
+    Services.publishing_api.publish(content_id, "minor")
 
     Services.publishing_api.put_links(
       content_id,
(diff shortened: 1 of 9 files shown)


rboulton:
I approve of the changes.  It would be better if the app could send a "major" update for brand new topics, simply because this would be closer to "truth" - a new topic is a major update for that item.  However, ensuring that we don't send out emails incorrectly is more important than fiddling with these semantics, and it would be a lot of work for no real gain to try and make the publisher aware of whether the topic was being newly created or being updated.

PR #188, opened 2016-03-09
Republish content items with `republish` update_type

tijmenb:
When republishing all content from this application, we should be using the republish `update_type`. This semantically correct and will prevent email alerts from being sent. It also works around a thing in publishing-api where a bug occurs when trying to publish a redirect with update_type `minor`.

Includes a bunch of refactorings to make the flow more obvious. Best reviewed per-commit.


bishboria:
:+1:

PR #190, opened 2016-03-18
Stop gds-api-adapters caching

tijmenb:
This caching is almost never wanted because this is a publishing application and we want to most up-to-date information possible.

@mobaig


MatMoore:
Looks good to me. I think if there is any need for caching in the publisher it should be decided per request instead.

PR #191, opened 2016-04-06
Redirect to taxon list after updating a taxon

MatMoore:
When I edit a taxon, :back sends me back to the form itself, so
its not obvious whether anything happened. This change makes
create and update do the same thing, i.e. redirect back to the list.

This also makes it easier to edit multiple taxons at once.


tijmenb:
:+1: very nice!

PR #194, opened 2016-04-26
Request tracing

tuzz:
https://trello.com/c/Eddv2AiS/688-for-migrated-formats-update-for-request-tracing

This change makes it so that a request_id is passed into Sidekiq jobs so that we can send this as a header downstream, via GdsApi::GovukHeaders. Currently, this header is lost when the Sidekiq jobs run because they run in a separate process to the one in which the request was processed.

This is based on this pull request: alphagov/whitehall#2567

### DIFF
--- a/app/workers/worker_base.rb
+++ b/app/workers/worker_base.rb
@@ -1,0 +1,20 @@
+class WorkerBase
+  include Sidekiq::Worker
+
+  def self.perform_async(*args)
+    args << { request_id: GdsApi::GovukHeaders.headers[:govuk_request_id] }
+    super(*args)
+  end
+
+  def perform(*args)
+    last_arg = args.last
+
+    if last_arg.is_a?(Hash) && last_arg.keys == ["request_id"]
+      args.pop
+      request_id = last_arg["request_id"]
+      GdsApi::GovukHeaders.set_header(:govuk_request_id, request_id)
+    end
+
+    call(*args)
+  end
+end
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -6,6 +6,7 @@
 require "action_controller/railtie"
 require "action_view/railtie"
 require "sprockets/railtie"
+require "tilt/erubis"
```
tijmenb: What is this?

app/workers/worker_base.rb
```diff
@@ -0,0 +1,25 @@
+class WorkerBase
+  include Sidekiq::Worker
+
+  def self.perform_async(*args)
+    args << request_id_argument
+    super(*args)
+  end
+
+  def self.request_id_argument
```
tijmenb: Any reason not to inline this?

tijmenb:
Out of interest, have any alternatives been looked at to doing this? It looks a little bit like a hack here.

config/application.rb
tuzz: This fixes the following warning:

```
WARN: tilt autoloading 'tilt/erubis' in a non thread-safe way; explicit require 'tilt/erubis' suggested.
```

tuzz:
As discussed offline, an RFC to perhaps add a govuk-sidekiq gem may be in order.

tuzz:
Hi @tijmenb, are you happy for this to be merged given we have an RFC to address your "shotgun surgery" concern as a follow-up action? Thanks

PR #196, opened 2016-05-04
Fix curated list item titles

benhyland:
Item titles for display in collections-publisher are currently looked up from data
persisted to the mysql db at list curation time.
This is not consistent with the collections frontend, which looks up titles from rummager.

It is possible for an item to appear in a curated list but to since have had its tag removed
or been deleted. In this case the collections frontend doesn't display the item.

Also, it is possible for a content item to have been renamed since it was included in a
curated list. It can then be added again under its new title.
In this case, collections-publisher will display two items which, while pointing to the
same base_path, appear different becase the persisted titles are different.
Collections frontend will display both items using the most recent title from rummager,
giving the appearance of duplicates which cannot be found in collections-publisher.

See https://trello.com/c/gfzHlBUn/620-collections-publisher-duplicates-in-list-editor-issue

Finally, it is possible for the persisted title to be null, in which case
collections-publisher can display an empty `<li>` element.

Once all taggable content is being published via publishing-api, we can dispense with
the rummager lookup in favour of publishing-api.

In the meantime, this commit fixes the display issues in collections-publisher by using
the rummager title where it exists, and falling back to the persisted title or placeholder
text where it does not.

@benhyland and @tijmenb

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -108,6 +108,10 @@ class Tag < ActiveRecord::Base
     @_tagged_documents ||= TaggedDocuments.new(self)
   end
 
+  def tagged_document_for_base_path(base_path)
+    tagged_documents.find { |document| document.base_path == base_path }
+  end
+
   def legacy_tag_type
     nil
   end
--- a/spec/models/list_item_spec.rb
+++ b/spec/models/list_item_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe ListItem do
+  describe "#display_title" do
+    it "shows the rummager title by default" do
+      stub_any_call_to_rummager_with_documents([{ link: "/some-link", title: "The Title" },
+                                                { link: "/some-other-link", title: "Another Title" }])
+
+      tag = create(:topic)
+      list_item = create(:list_item, base_path: "/some-link", list: create(:list, tag: tag))
+
+      expect(list_item.display_title).to eql('The Title')
+    end
+
+    it "falls back to the cached title" do
+      stub_any_call_to_rummager_with_no_documents
+
+      tag = create(:topic, slug: "my-tag-slug")
+      list_item = create(:list_item, base_path: "/some-link", title: "My persisted title", list: create(:list, tag: tag))
+
+      expect(list_item.display_title).to eql('My persisted title')
+    end
+  end
+end
--- a/app/models/list_item.rb
+++ b/app/models/list_item.rb
@@ -5,4 +5,8 @@ class ListItem < ActiveRecord::Base
 
   attr_accessor :tagged
   alias :tagged? :tagged
+
+  def display_title
+    list.tag.tagged_document_for_base_path(base_path).try(:title) || title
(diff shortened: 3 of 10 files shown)


app/models/tag.rb
```diff
@@ -108,6 +108,10 @@ def tagged_documents
     @_tagged_documents ||= TaggedDocuments.new(self)
   end
 
+  def tagged_document_for_base_path(base_path)
+    tagged_documents.find { |document| document.base_path == base_path }
```
Davidslv: if you trying to find the first match use `detect` instead of `find`

> Passes each entry in enum to block. Returns the first for which block is not false. If no object matches, calls ifnone and returns its result when it is specified, or returns nil otherwise.
> If no block is given, an enumerator is returned instead.
```
(1..10).detect   { |i| i % 5 == 0 and i % 7 == 0 }   #=> nil
(1..100).find    { |i| i % 5 == 0 and i % 7 == 0 }   #=> 35
```

spec/models/list_item_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe ListItem do
+  describe "#display_title" do
+    it "shows the rummager title by default" do
+      stub_request(:get, RummagerHelper::SEARCH_ENDPOINT)
```
Davidslv: Is there any test helper for this (out of curiosity, not a blocker) ?

app/models/tag.rb
benhyland: The docs suggest `find` and `detect` are synonyms - can you explain the difference?
http://ruby-doc.org/core-2.3.1/Enumerable.html

spec/models/list_item_spec.rb
benhyland: What is a test helper and how would I check? The term is not easily googleable - it sounds like a term of art in Ruby or Rails world?

Davidslv: It seems to have been something invented here, it's part of gds-api-adapters, you can find them here https://github.com/alphagov/gds-api-adapters/tree/master/lib/gds_api/test_helpers

tijmenb: We could extract the method into a module. https://github.com/alphagov/collections-publisher/blob/fix-curated-list-item-titles/spec/support/rummager_helper.rb would be the place for that probably.

app/models/tag.rb
Davidslv: I took the quote from the here: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-find

The example seems to speak for itself that they are not synonyms

tijmenb: Yeah, they should be equivalent. @Davidslv perhaps you're thinking of `select` / `filter`?

benhyland: We figured this one out in person - the docs are really not easy to read ><

benhyland:
Addressed the comment re test helper - there was an existing helper with some useful methods already, tyvm!

PR #197, opened 2016-05-10
Use `document_type` and `schema_name`

tijmenb:
The `format` field is deprecated and should not be used anymore.

Includes some new schema testing for the taxons.


benhyland:
👍 getting too stale to not merge

PR #199, opened 2016-05-31
Fix schema tests 

tijmenb:
When a new PR is made on govuk-content-schemas, the Jenkins task runs `./jenkins-schema.sh`. This test task's purpose is to run a subset of all tests that interact with the schemas. This is done so that govuk-content-schema tests run faster.

The problem is that for this to happen, the developer needs to tag schema tests with `schema_test: true`. This is error prone, and we currently have multiple schema tests that aren't triggered by jenkins.
This causes an annoying thing where a govuk-content-schemas PR is merged that makes the tests (on master) for this repo fail.

To prevent this from happening again, we choose here to run _all_ the tests. This causes wasted CPU cycles, but I'm of the opinion that that's an acceptable loss compared to potential bugs not being caught
by CI. Especially now that publishing-api is working on validating the incoming requests against the schemas.

### DIFF
--- a/app/presenters/root_browse_page_presenter.rb
+++ b/app/presenters/root_browse_page_presenter.rb
@@ -26,6 +26,7 @@ class RootBrowsePagePresenter
       publishing_app: "collections-publisher",
       rendering_app: "collections",
       routes: routes,
+      details: {},
     }
   end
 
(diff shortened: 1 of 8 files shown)


mobaig:
👍

PR #201, opened 2016-06-09
Update Nokogiri to 1.6.8

Davidslv:
This is a security update https://github.com/sparklemotion/nokogiri/issues/1473

Even though it's only used by capybara to run our test suite we don't
want to see the alerts in our security audit.


mobaig:
👍

PR #202, opened 2016-06-17
Remove RummagerNotifier

benhyland:
We are turning on indexing of mainstream browse pages,
topics and organisations from notifications sent by publishing-api.
collections-publisher no longer needs to send these notifications itself.

Since the only tags collections-publisher handles are mainstream browse pages,
topics and archived topics, we can simply eliminate all notifications to Rummager.


jackscotti:
> We are turning on indexing of mainstream browse pages,
can you provide a link to a ticket/PR ?

tijmenb:
Sorry, the ticket was unclear, added a comment here: https://trello.com/c/LbZA0i15

PR #203, opened 2016-06-21
Use the govuk_sidekiq gem

mgrassotti:
https://trello.com/c/JQREOuvh/424-use-sidekiq-gem

Use the govuk gem to setup and initialize sidekiq.
The headers inclusion in the worker args is no longer required as it is
already performed inside the gem.


tijmenb:
Cool!

I think there's a bunch more to remove though - https://github.com/alphagov/collections-publisher/pull/194 introduced the base class that we don't need anymore. Perhaps reverting https://github.com/alphagov/collections-publisher/pull/194/commits/bf1e09f0dc5ef35755701c07fc953a695b3e519f is an option.

mgrassotti:
👍 thanks!

Davidslv:
awesome 👍

PR #204, opened 2016-06-29
Remove redundant argument

tijmenb:
The usage of `govuk_headers` was removed in: https://github.com/alphagov/collections-publisher/pull/203

To keep this worker backwards compatible, this argument was kept. Now that we've deployed collections-publisher to production we can remove it.


tijmenb:
The failing tests are related to these problems: https://github.com/alphagov/whitehall/pull/2633#issuecomment-229607776

tijmenb:
Don't merge this, https://github.com/alphagov/govuk_sidekiq/pull/10 should provide a more general solution.

tijmenb:
Updated with the new version of `govuk_sidekiq`.

PR #206, opened 2016-07-07
Send 2nd level browse pages in the details hash

jackscotti:
content-store does not retain the order of the items that are sent in the links
hash, therefore in order to allow custom 2nd level browse pages ordering we
have to read the links order from the details hash.

Deploy and run rake task to republish all items pages/tags before merging and deploy:
https://github.com/alphagov/collections/compare/sort-2nd-level-browse-pages

Depends on: https://github.com/alphagov/govuk-content-schemas/pull/343

paired on it with @rubenarakelyan

### DIFF
--- a/app/presenters/mainstream_browse_page_presenter.rb
+++ b/app/presenters/mainstream_browse_page_presenter.rb
@@ -20,7 +20,8 @@ private
 
   def details
     super.merge(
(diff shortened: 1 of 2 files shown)


tijmenb:
👍 excellent

PR #207, opened 2016-07-11
Print rake task logs

jackscotti:
This is so that I do not have to ssh into a machine to read the logs, the change will allow me to read them on Jenkins or via the console from which I am running the rake task


rubenarakelyan:
👍  Nice and simple :)

PR #210, opened 2016-07-14
Taxons with multiple parents

carvil:
We would like Taxons to have multiple parents instead of just one. At the moment, there is a link type called `parent` which has been used for this purpose. Another purpose of `parent` is to help out building breadcrumbs on pages.

The idea of this change is to allow us to setup multiple parents to a taxon. In doing so, we are adding the multiple parents to `parent` until we create a new link type called `taxon_parents` in the schema for taxons.

This PR will need an extra commit to fix those link calls from `parent` to `taxon_parent` after the schema is changed.

This way, we leave `parent` untouched so we don't interfere with the breadcrumb logic.

https://trello.com/c/nImPlcNS/19-allow-taxons-to-have-multiple-parents

This PR was created by me and @Davidslv


tijmenb:
Can you add a screenshot for the UI changes?

carvil:
@tijmenb Yes, please find them below:

## Edit page

screen shot 2016-07-14 at 12 42 19

## New page

screen shot 2016-07-14 at 12 42 48

## Show page (now including parent information)

screen shot 2016-07-14 at 12 43 42

## Index page

Stays the same for now as discussed.

spec/features/taxonomy_manager_spec.rb
```diff
@@ -3,6 +3,8 @@
 RSpec.feature "Managing taxonomies" do
   before do
     stub_user.permissions << "Edit Taxonomy"
+    @item_1 = { title: "I Am A Taxon", content_id: "ID-1", base_path: "/foo" }
```
tijmenb: Perhaps `@taxon_1` & `@taxon_2` would be more self-explanatory.

carvil:
FYI, we need to run a rake task after deploy.

app/forms/taxon_form.rb
```diff
@@ -1,5 +1,5 @@
 class TaxonForm
-  attr_accessor :title, :parent, :content_id, :base_path
+  attr_accessor :title, :taxon_parents, :content_id, :base_path
```
mobaig: Minor thing, but the naming feels a little redundant here. This may as well be `parents`. If this is about naming confusion with the existing `parent` field, perhaps `parent_taxons` would be better still. This way you're explicitly stating that the parent items are taxons themselves.

app/services/taxonomy/taxon_parents_initializer.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def assign_existing_parents_to_taxon_parents
+      @taxon_parents = links.parent if links.present? && links.parent.present?
+    end
+
+    def taxon_parents_present?
+      links.present? && links.taxon_parents.present?
+    end
+
```
mobaig: whitespace

app/services/taxonomy/taxon_parents_initializer.rb
```diff
@@ -0,0 +1,57 @@
+module Taxonomy
+  # Copies the parent links to the new field taxon_parents
+  class TaxonParentsInitializer
```
mobaig: Need a different name I reckon - 'initializer' is a little overloaded in Rails land. 

Could call this something like `PopulateTaxonParents` (or `PopulateParentTaxons` if we go with my naming suggestion above 😃 ).

app/forms/taxon_form.rb
carvil: It's a good point. What we wanted to communicate here is that this taxon has a list of parents (multiple ones, not just one) that are all taxons too.

The reason for this name here was to follow a bit what comes from the schemas (PRL: https://github.com/alphagov/govuk-content-schemas/pull/355/files) so we have consistency in the naming. However, I do think that `parent_taxons` sounds better...

I will change this both here and in the schema. 👍

app/services/taxonomy/taxon_parents_initializer.rb
carvil: 👍 we will fix that

app/services/taxonomy/taxon_parents_initializer.rb
carvil: Yup, it could get a bit confusing. We will change that name.

mobaig:
👍  nice

PR #211, opened 2016-07-14
Update linter version

carvil:
In trying to run the linter in this repo in another branch, I got the following error:

```
$ bundle exec govuk-lint-ruby --diff --cached --format clang app test lib
Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.
(obsolete configuration found in /usr/lib/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/govuk-lint-0.5.0/configs/rubocop/gds-ruby-styleguide.yml, please update it)
```

I couldn't just update `govuk-lint` because of rake:

```
Bundler could not find compatible versions for gem "rake":
  In snapshot (Gemfile.lock):
    rake (= 11.2.2)

  In Gemfile:
    rails (= 4.2.7) was resolved to 4.2.7, which depends on
      railties (= 4.2.7) was resolved to 4.2.7, which depends on
        rake (>= 0.8.7)

    govuk-lint (= 1.1.0) was resolved to 1.1.0, which depends on
      scss_lint (~> 0.44.0) was resolved to 0.44.0, which depends on
        rake (~> 10.0)
```

I have set the rake version to the older `10.0` so we can upgrade the linter here. That's not ideal, but it will get us going.

I was also thinking about upgrading `rake` and `scss_lint` in `govuk-lint` so we don't need to use an older version of `rake` here, but I wasn't sure if that would break anything else.

What do you think?


Gemfile
```diff
@@ -32,7 +33,7 @@ group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
   gem 'quiet_assets', '~> 1.1.0'
-  gem 'govuk-lint', '0.5.0'
+  gem 'govuk-lint'
```
jackscotti: We might want to pin to a version here

carvil:
Please check https://github.com/alphagov/govuk-lint/pull/52

After that's merged, we can remove the `upgrade rake` commit and only upgrade govuk-lint.

Gemfile
carvil: I was thinking that without the pinned version, we could upgrade this at any time with `bundle update govuk-lint` (and double check the current version in Gemfile.lock). What do you think?

jackscotti: Those are preferences I believe. Having it pinned in the Gemfile we will know the version by looking at the Gemfile and will not require us to look at the Gemfile.lock file. 

I have no strong preference, but it seems that we prefer to pin down gem versions in this project.

carvil: Yup, I understand where you are coming from on this.

The way I have been looking at this so far is: `Gemfile.lock` already remembers versions, so I have reserved pinned versions in the Gemfile for 2 reasons:

- for gems that we have a hard dependency on (e.g. if I bump my version of gem X nothing will work and we can't commit the time to fix that right now);
- rails (pinned to minor version only so we get patch version upgrades without touching the Gemfile).

This way, we can run `bundle update my-awesome-gem` without having to edit the Gemfile (or find out what the latest version is). If all goes well, we only commit the `Gemfile.lock`.

What do you think?

Davidslv:
👍

PR #213, opened 2016-07-21
Use alphagov/govuk-ci-scripts

tijmenb:
This commits replaces our jenkins scripts with the canonical version from https://github.com/alphagov/govuk-ci-scripts.


rubenarakelyan:
👍

PR #216, opened 2016-08-02
Add SECRET_KEY_BASE for production

tijmenb:
We will use this later, so that we don't have to overwrite secrets.yml from the alphagov-deployment repo.

https://trello.com/c/dKfYzDNx


rubenarakelyan:
👍

PR #219, opened 2016-08-12
Add browse page archiving

rubenarakelyan:
Currently, Collections Publisher allows users to archive or delete topics that are no longer required, but does not allow the same functionality with mainstream browse pages. The underlying functionality to do this exists in the Publishing API but is not exposed here.

This commit adds the user-facing functionality to allow mainstream browse pages to be deleted (if they have not yet been published) or archived and redirected (if they have already been published). The user experience is identical to archiving/deleting topics.

Paired with @jackscotti 

Trello: https://trello.com/c/olZEBaef/67-make-it-possible-to-redirect-browse-pages-in-collections-publisher

Unpublished mainstream browse page has a "remove" button:
screen shot 2016-08-12 at 15 25 48

Published mainstream browse page has an "archive" button:
screen shot 2016-08-12 at 15 25 57

Archiving a mainstream browse page can redirect it either to another browse page or any path on GOV.UK:
screen shot 2016-08-12 at 15 26 07

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -34,3 +34,8 @@ button {
   @extend .btn;
   @extend .btn-default;
 }
+
+#new_mainstream_browse_page_archival_form,
+#new_topic_archival_form {
+  padding-bottom: 2em;
+}
--- a/app/forms/mainstream_browse_page_archival_form.rb
+++ b/app/forms/mainstream_browse_page_archival_form.rb
@@ -1,11 +1,11 @@
-class ArchivalForm
+class MainstreamBrowsePageArchivalForm
   include ActiveModel::Model
   attr_accessor :tag, :successor, :successor_path
 
   validates :successor_path, presence: true, valid_govuk_path: true, if: :redirecting_to_path?
 
-  def topics
-    published_topics - [tag]
+  def browse_pages
+    published_browse_pages - [tag]
   end
 
   def archive_or_remove
@@ -30,14 +30,15 @@ private
 
   def successor_object
     if redirecting_to_path?
-      OpenStruct.new(base_path: successor_path, subroutes: [])
+      Struct.new("RedirectToPath", :base_path, :subroutes)
+      Struct::RedirectToPath.new(successor_path, [])
     else
-      Topic.find_by_id(successor)
+      MainstreamBrowsePage.find_by_id(successor)
     end
   end
 
-  def published_topics
-    Topic.includes(:parent).published.sort_by(&:title_including_parent)
+  def published_browse_pages
+    MainstreamBrowsePage.includes(:parent).published.sort_by(&:title_including_parent)
   end
 
   def redirecting_to_path?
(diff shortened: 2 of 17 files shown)


jackscotti:
This adds a bit of duplication with what is done for Topics, maybe we can try to refactor the two archivers into a more general one? 
Also we should use shared view partials.

spec/features/archiving_tags_for_mainstream_browse_pages_spec.rb
```diff
@@ -0,0 +1,147 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/content_store'
+
+RSpec.feature "Archiving tags for mainstream browse pages" do
+  include PublishingApiHelpers
+  include CommonFeatureSteps
+  include GdsApi::TestHelpers::ContentStore
+
+  before do
+    # Stub rummager so that topics do not have links.
```
jackscotti: # Stub rummager so that topics do not have links. 

topics => Mainstream Browse Pages ?

rubenarakelyan: Thanks, fixed :)

app/assets/stylesheets/forms.scss
```diff
@@ -34,3 +34,8 @@ button {
   @extend .btn;
   @extend .btn-default;
 }
+
+#new_mainstream_browse_page_archival_form,
```
mobaig: Worth double checking in the relevant styleguide, but I think css selectors are hyphen separated.

rubenarakelyan: @mobaig These two classes are auto-generated by Rails based on the views, so they always use underscores.

spec/features/archiving_tags_for_mainstream_browse_pages_spec.rb
```diff
@@ -0,0 +1,147 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/content_store'
+
+RSpec.feature "Archiving tags for mainstream browse pages" do
```
mobaig: Small thing, but I think this would be clearer if worded "Archiving mainstream browse page tags".

Same with "Archiving topic tags", plus the corresponding filenames.

app/forms/mainstream_browse_page_archival_form.rb
```diff
@@ -0,0 +1,46 @@
...
+  rescue GdsApi::HTTPClientError
+    errors.add :base, "The tag could not be deleted because of an error."
+    false
+  end
+
+private
+
+  def successor_object
+    if redirecting_to_path?
+      OpenStruct.new(base_path: successor_path, subroutes: [])
```
mobaig: Let's replace this with a Struct. You'll get a proper NoMethodError if a client object mishandles method calls on the successor_object, rather than the `nil` that OpenStruct returns.

mobaig:
Couple of small bits, otherwise looks good.

spec/features/archiving_tags_for_mainstream_browse_pages_spec.rb
rubenarakelyan: Done

app/forms/mainstream_browse_page_archival_form.rb
rubenarakelyan: Sounds good, done

PR #220, opened 2016-08-12
Remove comparison rake task

tijmenb:
This task was used to compare content-api and publishing-api, and is no longer used. It can be safely removed.

https://trello.com/c/HcMgeRfi


rubenarakelyan:
👍

PR #222, opened 2016-08-26
Link tagged documents to Content Tagger

jackscotti:
To make it easier to remove the tags from the content.

Feel free to highlight how bad my HTML is, might learn something new!
https://trello.com/c/j0AgxRd5/74-collections-publisher-should-link-to-content-tagger-for-topic-links

## Before:
screen shot 2016-08-30 at 11 27 24

## After:
screen shot 2016-08-30 at 11 25 12
## Links to:
screen shot 2016-08-30 at 11 25 51

### DIFF
--- a/app/views/shared/tags/_curated_links_preview.html.erb
+++ b/app/views/shared/tags/_curated_links_preview.html.erb
@@ -4,7 +4,9 @@
 
 <h2>Links</h2>
 
-<p>Links for this tag have been curated into lists.</p>
+<p>
+  Links for this tag have been curated into lists. Click on the title for editing on Content Tagger.
+</p>
 
 <% tag.lists.each do |list| %>
   <h4><%= list.name %></h4>
@@ -14,10 +16,14 @@
   <ul>
   <% list.list_items_with_tagging_status.each do |list_item| %>
     <li>
-      <% unless list_item.tagged? %>
+      <% if list_item.tagged? %>
+        <a href=<%= content_tagger_url + '/taggings/' + list_item.content_id %>>
+          <%= raw list_item.display_title || "<em>Unknown title (#{list_item.base_path})</em>" %>
+        </a>
+      <% else %>
         <p class='label label-warning'>Tag was removed</p>
+        <%= raw list_item.display_title || "<em>Unknown title (#{list_item.base_path})</em>" %>
       <% end %>
-      <%= raw list_item.display_title || "<em>Unknown title (#{list_item.base_path})</em>" %>
     </li>
   <% end %>
   </ul>
@@ -26,6 +32,10 @@
 <h4>Uncurated items <em>(not shown to user)</em></h4>
 <ul>
 <% tag.uncurated_tagged_documents.each do |item| %>
-  <li><%= item.title %></li>
+  <li class="tagged-document">
+    <a href=<%= content_tagger_url + '/taggings/lookup' + item.content_id %>>
+      <%= item.title %>
+    </a>
+  </li>
 <% end %>
 </ul>
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,4 +2,8 @@ module ApplicationHelper
   def website_url(base_path)
     Plek.new.website_root + base_path
   end
+
+  def content_tagger_url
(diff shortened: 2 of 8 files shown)


app/views/shared/tags/_curated_links_preview.html.erb
```diff
@@ -17,7 +19,13 @@
       <% unless list_item.tagged? %>
         <p class='label label-warning'>Tag was removed</p>
       <% end %>
-      <%= raw list_item.display_title || "<em>Unknown title (#{list_item.base_path})</em>" %>
+      <% if list_item.base_path %>
+        <a href=<%= Plek.new.find('content-tagger') + '/taggings/lookup' + list_item.base_path %>>
+          <%= raw list_item.display_title || "<em>Unknown title (#{list_item.base_path})</em>" %>
```
MatMoore: Perhaps this should be an explicit if condition?

```
<% if ... %>
<em>Unknown...</em>
<% else %>
list_item.display_title
<% end %>
```

I'm not too familiar with the template language but it looks like this could have XSS problems if the title contained special characters.

app/views/shared/tags/_curated_links_preview.html.erb
```diff
@@ -26,6 +34,14 @@
 <h4>Uncurated items <em>(not shown to user)</em></h4>
 <ul>
 <% tag.uncurated_tagged_documents.each do |item| %>
-  <li><%= item.title %></li>
+  <li class="tagged-document">
+    <% if item.base_path %>
+      <a href=<%= Plek.new.find('content-tagger') + '/taggings/lookup' + item.base_path %>>
```
MatMoore: Could you do the `Plek.new.find` in the controller and then pass it into the view? It feels more like business logic than presentation logic.

app/views/shared/tags/_curated_links_preview.html.erb
carolinegreen: @jackscotti Think it's probably better to use <strong> unless you are using <em> for some particular reason?

jackscotti: I left `<em>` becuse I've found it there. I can surely replace it with `strong`.

app/views/shared/tags/_curated_links_preview.html.erb
tijmenb: Could we link directly to the tagging page with the content_id?

app/views/shared/tags/_curated_links_preview.html.erb
carolinegreen: No, think that's okay looking at it today and not on a Friday afternoon.  Could you add screenshots for what has changed?

app/views/shared/tags/_curated_links_preview.html.erb
jackscotti: If we linked to the tagging page without content-id we would not need to link each document but we could just add a single link in the page. With Mat we thought it would have been really useful to be able to click on a link in collections-publisher and be linked straightaway to its page on content-tagger. What do you think?

jackscotti:
@tijmenb @carolinegreen updated.

app/views/shared/tags/_curated_links_preview.html.erb
tijmenb: Yeah, but this ^ just redirects to the page. Since we already have the content_id of the thing we can link to it directly right?

https://content-tagger.integration.publishing.service.gov.uk/taggings/ffe22070-123b-4390-8cc4-51f9d5b5cc74

jackscotti: Oh, I did not know about this route in the content-tagger. Updating!

lib/services.rb
```diff
@@ -33,6 +33,10 @@ def self.rummager
       disable_cache: true
     )
   end
+
+  def self.content_tagger
```
tijmenb: As this is not a service/API, this might be a little confusing that it returns a URL.

jackscotti:
@tijmenb updated to use the content_id instead of the content_path. I will squash the commits if the changes are accepted.

tijmenb:
Yeah excellent, squash away!

PR #223, opened 2016-09-05
Send browse page parent tag to publishing-api

tijmenb:
This application currently does not send the `parent` in the links for browse pages, but it does for topics. This commit moves the parent (or children) code up to the superclass so topic and browse pages act the same.

Having the parent in the links will allow us to render the breadcrumb from the content-store in collections.

https://trello.com/c/xvM9vb5L

After this is deployed we need to run the rake task that re-sends all content to the publishing-api.

### DIFF
--- a/app/presenters/tag_presenter.rb
+++ b/app/presenters/tag_presenter.rb
@@ -114,9 +114,16 @@ private
     tag.published_groups
   end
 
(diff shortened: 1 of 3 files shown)


rubenarakelyan:
👍

PR #226, opened 2016-11-03
Remove interaction with panopticon

tijmenb:
This app currently sends tags to panopticon, so that panopticon can save them in the shared database and we can use them from the content-api.

The content-api doesn't need the tags anymore because everything is doing via the content-store and publishing-api.

This commit removes all interaction with panopticon. As an added bonus, we can now allow removal of browse pages and topics even if they have documents tagged to it. This previously wasn't possible because panopticon would get confused otherwise.

https://trello.com/c/KfgeIRtP


rubenarakelyan:
👍

PR #229, opened 2016-12-12
Fix all Rubocop linter warnings

suzannehamilton:
This is in preparation for migrating the build to Jenkins 2. Jenkins 2 will require us to add new build scripts which could optionally include Rubocop linting, so this seems like a convenient point to add linting to this project.

All the Rubocop errors were trivial, and most could be fixed automatically.

I'll comment on the two manual fixes in case you want to look at those in more detail.

### DIFF
--- a/app/workers/dependent_tag_publish_worker.rb
+++ b/app/workers/dependent_tag_publish_worker.rb
@@ -4,20 +4,20 @@ class DependentTagPublishWorker
   def perform(tag_id)
     tag = Tag.find(tag_id)
 
-    tag.dependent_tags.each do |tag|
-      presenter = TagPresenter.presenter_for(tag)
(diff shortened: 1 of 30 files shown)


app/workers/dependent_tag_publish_worker.rb
```diff
@@ -4,20 +4,20 @@ class DependentTagPublishWorker
   def perform(tag_id)
     tag = Tag.find(tag_id)
 
-    tag.dependent_tags.each do |tag|
-      presenter = TagPresenter.presenter_for(tag)
+    tag.dependent_tags.each do |dependent_tag|
```
suzannehamilton: I fixed the variable shadowing of `tag` here.

PR #231, opened 2016-12-19
Only build 'released-to-production' branch to test content schemas

suzannehamilton:
Unlike dev branches, the 'released-to-production' branch should not be built when the branch is updated. The only reason to build it is to test it against changes to the content schemas.

This change adds Jenkins configuration which distinguish between updates to 'released-to-production' (in which case Jenkins should do nothing) and builds triggered by the content schema build (in which case, run the build as normal, but check out the correct content schema branch.

The govuk-content-schemas Jenkins config will be updated to trigger this build, and it will provide an 'IS_SCHEMA_TEST' parameter to identify it as a schema test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,9 +19,33 @@ node {
       paramsToUseForLimit: 'collections-publisher',
       throttleEnabled: true,
       throttleOption: 'category'],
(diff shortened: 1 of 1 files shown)


tijmenb:
👍 I hope we can extract this to the jenkins lib later!

PR #235, opened 2017-01-17
Delete all 'beta' messages from topics

suzannehamilton:
Topic pages are no longer in beta, and they will be deprecated soon when the new taxonomy-based navigation is added. This means that the 'beta' tag is out-of-date, and can be deleted.

The 'beta' tag is not relevant to browse pages either, so all references have been removed from the code.

New topics will be created with default values for properties related to 'beta' content. This currently means that they will be created in the 'live' phase, and with no 'beta' property in 'details'.

https://trello.com/c/oW7gSMoB/328-remove-beta-tags-from-topics-in-collections

A few things I'm not sure of:

- Is it OK that if you update an existing beta topic, it removes the `beta` property in the `details` section? Or do we need to preserve those for existing topics?
- There's still a `beta` column in the collections-publisher DB, which is no longer referenced by the code. Should I remove it? It should be equivalent to the 'beta' value in the content store, so we shouldn't lose any data by removing it. (It would just make it hard to roll this change back if we needed to.)
- And should we remove the `beta` property from the content schema for topics?

Once deployed:

- [ ] Merge #236 
- [ ] Remove `beta` from the content schema for topics

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -69,7 +69,7 @@ class TopicsController < ApplicationController
 private
 
   def topic_params
-    params.require(:topic).permit(:slug, :title, :description, :parent_id, :beta)
+    params.require(:topic).permit(:slug, :title, :description, :parent_id)
   end
 
   def find_topic
--- a/app/helpers/status_helper.rb
+++ b/app/helpers/status_helper.rb
@@ -12,14 +12,9 @@ module StatusHelper
   def labels_for_tag(tag)
     labels = []
     labels << draft_tag if tag.draft?
-    labels << beta_tag if tag.beta?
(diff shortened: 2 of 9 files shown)


tijmenb:
> Is it OK that if you update an existing beta topic, it removes the beta property in the details section? Or do we need to preserve those for existing topics?

Yes, okay to remove. No need to preserve them.

> There's still a beta column in the collections-publisher DB, which is no longer referenced by the code. Should I remove it? It should be equivalent to the 'beta' value in the content store, so we shouldn't lose any data by removing it. (It would just make it hard to roll this change back if we needed to.)

Yes, can be removed. The data is fine to delete as well.

> And should we remove the beta property from the content schema for topics?

Yep, good catch!

rubenarakelyan:
👍

PR #236, opened 2017-01-17
Delete 'beta' flag from tags in DB

suzannehamilton:
The 'beta' flag was removed from the rest of the codebase in a previous commit. This cleans up the now-unused field from the DB.

The only kinds of content edited in collections-publisher are topics and browse pages. Topics are no longer in beta and browse pages never were, so it's safe to remove the beta flag data entirely.

https://trello.com/c/oW7gSMoB/328-remove-beta-tags-from-topics-in-collections


suzannehamilton:
@tijmenb Since this change deletes data, is it worth waiting until #235 has been released, so that we know it works in production?

tijmenb:
Good idea 👍

PR #238, opened 2017-02-06
Add content_store_document_type to be sent to rummager.

Davidslv:
Finding Things has been working into adding a new field in rummager
called `content_store_document_type`.

This commit addds that field into rummager and also updates the spec to
test the output to stdout and avoid having it displayed when tests run.

Trello:
https://trello.com/c/iMotNdUv/461-add-more-document-type-to-search


spec/notifiers/rummager_notifier_spec.rb
```diff
@@ -34,6 +34,7 @@
       expect(rummager).to have_received(:add_document)
         .with("edition", "/topic/a-test-topic", content_id: '28ac662c-09cf-4baa-9e7c-98339a2a3bcd',
           format: 'specialist_sector',
+          content_store_document_type: "specialist_sector",
```
tijmenb: This should be `topic` (like it is in the content store).

PR #240, opened 2017-02-13
Add publishing_app & rendering_app to rummager payload

tijmenb:
This will make it possible to search for anything rendered or publishing by a certain app from search.

https://trello.com/c/iMotNdUv


carvil:
👍

PR #245, opened 2017-04-21
Upgrade to Rails 5.0.2

andrewgarner:
Follows the Rails Guide [documented approach](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html). Also upgrades all Ruby gem dependencies.

[See Trello](https://trello.com/c/UvadVLm8)


Gemfile
```diff
@@ -1,49 +1,41 @@
...
-gem 'gds-sso', '~> 11.0.0'
-gem 'gds-api-adapters', '~> 30.2.0'
-
-gem 'govuk_admin_template', '~> 4.1'
-gem 'generic_form_builder', '~> 0.13.0'
-gem 'aasm', '~> 4.3.0'
+gem 'rails', '~> 5.0'
 
+gem 'aasm', '~> 4.0'
+gem 'airbrake', '~> 6.0'
```
whoojemaflip: I have a suspicion that `airbrake` > `4.x` is API-incompatible with the version of Errbit we're running.

app/controllers/tags_controller.rb
```diff
@@ -1,9 +1,9 @@
 class TagsController < ApplicationController
-  before_filter :find_tag
-  before_filter :require_gds_editor_permissions_to_edit_browse_pages!
+  before_action :find_tag
+  before_action :require_gds_editor_permissions_to_edit_browse_pages!
 
   def publish_lists
     ListPublisher.new(@tag).perform
-    redirect_to :back
+    redirect_back(fallback_location: root_path)
```
tijmenb: TIL

Gemfile
andrewgarner: 5.5 is working in Publishing API – https://github.com/alphagov/publishing-api/blob/master/config/initializers/airbrake.rb

PR #252, opened 2017-11-06
Update task list page description

sihugh:
We've specified a meta description on the [task list page](https://github.com/alphagov/collections/blob/d2e95a7c052be9982b8d07eff0ea7de6384f1510/app/views/tasklist/show.html.erb#L4), so we want to make this consistent.

https://trello.com/c/wAMac6Dz/282-update-task-list-description-in-publishing-api

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -22,7 +22,7 @@ namespace :publishing_api do
       schema_name: "generic",
       document_type: "task_list",
       title: "Learn to drive a car: step by step",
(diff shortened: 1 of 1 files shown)


sihugh:
Yep, already is: https://github.com/alphagov/collections/blob/cf68a35642653cbc1714cc2c75359a1c847875eb/app/views/tasklist/show.html.erb#L4 👍

tijmenb:
Sorry, I meant fetching it from the content item instead of having it hardcoded there too.

PR #254, opened 2017-11-07
Add Dockerfile

RyanMacG:
As part of a project expanding on the existing alphagov/publishing-e2e-tests we are adding a basic Dockerfile for use from that repository.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM ruby:2.3.0
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
+
+ENV DATABASE_URL mysql2://root:root@mysql/collections_publisher_development
+ENV GOVUK_APP_NAME collections-publisher
+ENV PORT 3071
+ENV RAILS_ENV development
+ENV REDIS_HOST redis
+
+ENV APP_HOME /app
+RUN mkdir $APP_HOME
+
+WORKDIR $APP_HOME
+ADD Gemfile* .ruby-version $APP_HOME/
+RUN bundle install
+ADD . $APP_HOME
+
+CMD bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p $PORT -b '0.0.0.0'"
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,10 +7,12 @@ default: &default
 development:
   <<: *default
   database: collections_publisher_development
+  url: <%= ENV["DATABASE_URL"]%>
 
 test: &test
   <<: *default
   database: collections_publisher_test
+  url: <%= ENV["DATABASE_URL"].try(:sub, /(development)?$/, 'test')%>
 
 cucumber:
   <<: *test
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,14 @@
+if User.where(name: "Test user").present?
+  puts "Skipping because user already exists"
+  exit
+end
+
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ruby:2.3.1
```
kevindew: Looks like this project runs on [ruby 2.3.0](https://github.com/alphagov/collections-publisher/blob/master/.ruby-version)

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
```
kevindew: Does this need nodejs installed?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# for capybara-webkit
+RUN apt-get update -qq && apt-get install -y libqt4-webkit libqt4-dev xvfb
```
kevindew: I don't think we have capybara-webkit in this project (I can just see capybara alone)

config/database.yml
```diff
@@ -3,6 +3,7 @@ default: &default
   encoding: utf8
   username: collections_pub
   password: collections_publisher
+  host: <%= ENV['DATABASE_HOST'] || 'localhost' %>
```
kevindew: Can we use database_url instead of database_host? see: http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database

Otherwise we'd only be able to configure one MySQL database in the docker-compose rather than having one per application

Dockerfile
RyanMacG: The gemfile has `uglifier` as a dependency so requires a JS runtime

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ruby:2.3.0
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
+
+ENV DATABASE_URL mysql2://collections_pub@mysql/collections_publisher_development
```
kevindew: I see this is tied into: https://github.com/alphagov/publishing-e2e-tests/blob/add-collections/docker-compose.yml#L20-L22 which seems to be a bit more specific than other apps e.g: https://github.com/alphagov/publishing-api/blob/master/Dockerfile#L4 

And I wonder if that will make it more difficult for a second mysql app?

Is it possible to just run the mysql as a default mysql user and not need those constants in the docker-compose.yml?

RyanMacG: From what I've seen while adding this is it tries to create a database for the URL provided, I could try passing the user instead of as an ENV variable and see if the root user can work around these problems. Then we'd just need to set the root password instead.

kevindew: Yeah part of `db:setup` is the creating of the database(s).

So with postgres image we have a default user (postgres) with an empty password and we can proceed with that pretty simply. If mysql is the same we might be able to just have `mysql2://mysql@mysql/collections_publisher` and not need the mysql env vars?

config/database.yml
```diff
@@ -1,16 +1,18 @@
 default: &default
   adapter: mysql2
   encoding: utf8
-  username: collections_pub
-  password: collections_publisher
+  username: <%= ENV['DATABASE_USER'] || 'collections_pub' %>
+  password: <%= ENV['DATABASE_PASSWORD'] || 'collections_publisher' %>
```
kevindew: Do these need to be configurable? I don't think anything is using them and if I remember correctly DATABASE_URL takes precedence over them?

PR #258, opened 2017-11-25
Update Ruby to 2.4.2 and bump Gemfile dependancies

RoryMacDonald:
Ahead of enabling Dependabot on the repository, we want to be using the latest
Ruby version and gem dependancies, to minimise the onslaught of incremental PRs.


RoryMacDonald:
👋  @kevindew - I'm seeing a DB error in Jenkins when running tests: `Original message: Mysql2::Error: Can't create table 'collections_publisher_test.#sql-1419_2f9f9' (errno: 150)`. 

Am I imagining this, or did you say you had seen something similar in another app + that somebody was looking into this? If so, could you point me in the right direction?

kevindew:
@RoryMacDonald I don't have access right now to CI to confirm this - but I wonder if this is a consequence of CI running out of disk space yesterday? might just want to try it again

tijmenb:
Reran the tests and got:

```
ActiveRecord::MismatchedForeignKey: Column `list_id` on table `list_items` has a type of `int(11)`.
This does not match column `id` on `lists`, which has type `bigint(20)`.
To resolve this issue, change the type of the `list_id` column on `list_items` to be :integer. (For example `t.integer list_id`).

Original message: Mysql2::Error: Can't create table 'collections_publisher_test.#sql-29b0_5e12e' (errno: 150): ALTER TABLE `list_items` ADD CONSTRAINT `list_items_list_id_fk`
FOREIGN KEY (`list_id`)
  REFERENCES `lists` (`id`)
 ON DELETE CASCADE
```

kevindew:
@tijmenb that looks like a Rails 5 (or 5.1) upgrade issue. The db/schema.rb doesn't specify the id type of tables and will default to bigint whereas db/schema.rb does [specify](https://github.com/alphagov/collections-publisher/blob/master/db/schema.rb#L19) the type of foreign key as integer

kevindew:
@danielburnley do you want to generate the schema file from `db:schema:dump` I've tried on my machine and it comes out quite different to yours. Which will probably mean this is a headache first time someone does a migration.

PR #305, opened 2018-01-31
Update learn-to-drive-a-car content

sihugh:
This is the [learn-to-drive-a-car step_nav content](https://github.com/alphagov/collections/blob/ca108ab0ba73ed571e7a3697f36942758a353f1c/config/tasklists/learn-to-drive-a-car.json#L12-L219) updated to match the schema in https://github.com/alphagov/govuk-content-schemas/pull/709.

I've also moved each of the step_nav tasks into separate files for each step by step navigation journey in order to make reviewing them easier, as they're quite long.  We're starting work on our publishing tool so hopefully these won't hang around for too long.


lib/tasks/step_nav/learn_to_drive_a_car.rake
```diff
@@ -0,0 +1,226 @@
...
+  desc "Publish /learn-to-drive-a-car step navigation to publishing api"
+  task publish_learn_to_drive_a_car: :environment do
+    content_id = "e01e924b-9c7c-4c71-8241-66a575c2f61f"
+    params = {
+      base_path: "/learn-to-drive-a-car",
+      publishing_app: "collections-publisher",
+      rendering_app: "collections",
+      public_updated_at: Time.zone.now.iso8601,
+      update_type: "major",
+      change_note: "updating the content to match the new step_by_step_nav schema",
```
sihugh: @deborahchua I've just added this as I realised we were causing problems down the line by not providing a change note for a major update.

deborahchua: Ah good point

PR #317, opened 2018-02-19
 Set up access permissions for Step by Step pages

deborahchua:
The Modelling Services team are working on implementing the Step by Step pages into the publisher.
We want to be able to develop features and deploy them without affecting the workflow already in place.

Users with GDS editor permissions should be able to see the `Step by step pages` link next to `Topics`.

Trello Card: [Step by Step Page behind access permissions](https://trello.com/c/LjXNfJg1/504-step-by-step-page-behind-access-permissions)

screen shot 2018-02-19 at 11 23 06

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,8 +3,7 @@ Rails.application.routes.draw do
 
   root to: redirect('/topics', status: 302)
 
-  resources :step_by_step_pages, path: 'step-by-step-pages' do
-  end
+  get :"step-by-step-pages", to: 'step_by_step_pages#index'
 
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -3,6 +3,9 @@
 
   root to: redirect('/topics', status: 302)
 
+  resources :step_by_step_pages, path: 'step-by-step-pages' do
```
sihugh: This generates a bunch of routes that we don't need yet.  Maybe limit this just to a get of the index page?

deborahchua:
Thanks @sihugh , I have updated the routes now.

PR #321, opened 2018-02-22
Update StepByStep validation

deborahchua:
The current format regex for the `base_path` does not accept a single word. This has now been updated.

This is part of https://github.com/alphagov/collections-publisher/pull/320


sihugh:
I don't think this was a major problem, as I would expect most of the slugs would be of the form `<verb>-a-<thing>`. Good catch though 👍

PR #324, opened 2018-02-26
Add new Step model

deborahchua:
The Modelling Services team are implementing the Step by step publishing tool into Collections Publisher.

The `StepByStepPage` model has already been created in this pr: https://github.com/alphagov/collections-publisher/pull/320

This PR creates the `Step` model, which has a `belongs_to` relationship with `StepByStepPage`.

Trello card: [Allow Step by Step Pages and Steps to be stored in Collections Publisher](https://trello.com/c/oZX9zHNf/506-allow-step-by-step-pages-and-steps-to-be-stored-in-collections-publisher)

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -1,4 +1,6 @@
 class StepByStepPage < ApplicationRecord
+  has_many :steps, dependent: :destroy
+
   validates :title, :base_path, presence: true
   validates :base_path, format: { with: /\A([a-z0-9]+-)*[a-z0-9]+\z/ }
   validates :base_path, uniqueness: true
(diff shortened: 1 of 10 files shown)


app/models/step_by_step_page.rb
```diff
@@ -1,4 +1,6 @@
 class StepByStepPage < ApplicationRecord
+  has_many :steps, dependent: :destroy
```
sihugh: We might want to use something like `has_many :steps, -> { order(position: :asc) }` to ensure they come back in the right order.  We'd want to add a test for that too, so I'm happy for this to happen soon

deborahchua: Thanks @sihugh , I've made a note to remember to add this after 👍

PR #327, opened 2018-02-27
Build step by step pages

deborahchua:
https://trello.com/c/I7BkEdQz/522-build-pages-for-step-by-step-in-collections-publisher

### DIFF
--- a/app/views/step_by_step_pages/index.html.erb
+++ b/app/views/step_by_step_pages/index.html.erb
@@ -1,2 +1,46 @@
-<h1>StepByStepPages#index</h1>
-<p>Find me in app/views/step_by_step_pages/index.html.erb</p>
+<%
+  links = [
+    {
+      text: 'Step by step publisher',
+    }
+  ]
+%>
+
+<%= render 'shared/steps/step_breadcrumb', links: links %>
+
+<%= render 'shared/steps/form_notice', message: notice, status: "success" %>
+
+<h1>Step by step navigation publisher</h1>
+
+<p class="step-actions">
+  <%= link_to 'Create new', new_step_by_step_page_path, class: "btn btn-primary" %>
+</p>
+
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th>Title</th>
+      <th>Actions</th>
+      <th>Slug</th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% @step_by_step_pages.each do |step_by_step_page| %>
+      <tr>
+        <th>
+          <%= step_by_step_page.title %>
+        </th>
+        <td>
+          <ul class="list-inline">
+            <li><%= link_to 'Edit content', step_by_step_page %></li>
+            <li><%= link_to 'Delete', step_by_step_page, method: :delete, data: { confirm: 'Are you sure?' } %></li>
+          </ul>
+        </td>
+        <td>
+          <%= step_by_step_page.base_path %>
+        </td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
--- a/app/views/step_by_step_pages/show.html.erb
+++ b/app/views/step_by_step_pages/show.html.erb
@@ -17,5 +17,48 @@
 <h1><%= @step_by_step_page.title %></h1>
 
 <ul class="list-inline step-actions">
+  <li><%= link_to 'Add a new step', new_step_by_step_page_step_path(@step_by_step_page), class: "btn btn-primary" %></li>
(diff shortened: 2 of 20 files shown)


app/views/step_by_step_pages/index.html.erb
```diff
@@ -1,2 +1,59 @@
...
+    <% @step_by_step_pages.each do |step_by_step_page| %>
+      <tr>
+        <td>
+          <%= step_by_step_page.title %>
+        </td>
+        <td>
+          <%= link_to 'Edit content', step_by_step_page %>
+        </td>
+        <td>
+          Edit rules
```
selfthinker: I don't think we should add anything that isn't functional yet. I would simply remove any references to the rules.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -1,2 +1,59 @@
...
+        </td>
+        <td>
+          <%= link_to("Preview", step_by_step_page_preview_path(step_by_step_page)) %>
+        </td>
+        <td>
+          Download JSON
```
selfthinker: Same here. I would simply remove this as it isn't implemented yet.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -1,2 +1,59 @@
...
+        </td>
+        <td>
+          <%= link_to 'Delete', step_by_step_page, method: :delete, data: { confirm: 'Are you sure?' } %>
+        </td>
+        <td>
+          <span class="label label-success">Status</span>
```
selfthinker: Same here.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -1,2 +1,59 @@
...
+        </td>
+      <% if false %>
```
selfthinker: I guess this is the equivalent of a comment, which I would probably remove until needed.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -0,0 +1,55 @@
...
+
+<%= @step_by_step_page.introduction %>
+
+<p>Meta description: <%= @step_by_step_page.description %></p>
+
+<p class="text-right">
+  <%= link_to("Preview", step_by_step_page_preview_path(@step_by_step_page), class: "btn btn-info") %>
+  <%= link_to 'Edit title, intro or slug', edit_step_by_step_page_path(@step_by_step_page), class: "btn btn-primary" %>
+  <%= link_to 'Add a new step', new_step_by_step_page_step_path(@step_by_step_page), class: "btn btn-primary" %>
+  [Reorder steps]
```
selfthinker: Same here.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -0,0 +1,55 @@
...
+<%= render 'partials/form_notice', message: notice, status: "success" %>
+
+<h1><%= @step_by_step_page.title %></h1>
+
+<%= @step_by_step_page.introduction %>
+
+<p>Meta description: <%= @step_by_step_page.description %></p>
+
+<p class="text-right">
+  <%= link_to("Preview", step_by_step_page_preview_path(@step_by_step_page), class: "btn btn-info") %>
```
selfthinker: I wonder if we should even remove the preview? The current implementation is only very basic and has non-functional elements as well (mainly the "share preview link" link).

selfthinker:
I've just added the content changes Jon and me did yesterday.

I still cannot save the base path (now renamed "slug" in the UI), I always get an error that it's invalid. I tried several variations (starting or ending with slash or not). Is there any specific rule I need to follow? If yes, that should be made explicit in the UI.

selfthinker:
I have removed every reference to anything which wasn't implemented yet or wasn't implemented properly (like the preview). I couldn't remove ae648e17921f764dce0d6a8f26cab6851097cc36 from the git history because that contains other things than just adding the preview page, despite its commit message.

selfthinker:
I noticed that adding or editing a nav and editing a step all show a success feedback message, but adding a new step doesn't. Is that intentional?

selfthinker:
I've added a feedback message when a new step gets created (in the "add step pages" commit).

sihugh:
I've fixed the tests and squashed a few of the last commits

deborahchua:
Looks good @sihugh , thanks for fixing those tests!

selfthinker:
I've just added the latest design changes as well and cleaned up the git history so everything that belongs together is in one commit.

selfthinker:
I've also just moved the 'partials' from the partials folder into the existing 'shared' folder into a new 'steps' subfolder. I think it makes more sense as that is the place where there already are a few shared partials.

I'm happy with all the frontend stuff.

PR #331, opened 2018-03-02
Update cost of provisional driving

sihugh:
We list the fee for applying for a provisional as £34. This is the fee to apply online; the fee to apply by post is £43.

https://trello.com/c/H77MJZRD/477-change-fee-for-apply-for-a-provisional-licence


sihugh:
🚅

PR #337, opened 2018-03-08
Step nav frontend improvements

selfthinker:
These are some smaller frontend improvements to the step by step navigation publisher:

- Make titles and breadcrumbs more consistent
- Add unique page titles to every page
- Improve content of Markdown example snippet
- Make format of slugs clearer
- Add defaults for likely step options
- Hide optional step help section, show when needed
- Improve guidance links for step nav
- Fix breadcrumbs showing when empty
- Remove duplication and unnecessary complexity from form errors

### DIFF
--- a/app/views/shared/steps/_markdown_help.html.erb
+++ b/app/views/shared/steps/_markdown_help.html.erb
@@ -1,3 +1,5 @@
+<p><%= link_to "Content design guide", "https://gov-uk.atlassian.net/wiki/spaces/MS/pages/268140618/Content+design+principles+for+step+by+step+navigation+Q4" %> for step by step navigation</p>
+
 <%# intentionally not indented properly because the PRE formats all whitespace %>
 <pre class="small markdown-example">Paragraphs are separated by empty lines.
 Use non-bulleted lists for tasks. Add any costs after the link:
--- a/app/views/shared/steps/_markdown_help.html.erb
+++ b/app/views/shared/steps/_markdown_help.html.erb
@@ -1,14 +1,13 @@
-<%# sensible formatting not included because the PRE formats all whitespace %>
-<pre class="small">Paragraphs are separated by empty lines.
+<%# intentionally not indented properly because the PRE formats all whitespace %>
+<pre class="small markdown-example">Paragraphs are separated by empty lines.
+Use non-bulleted lists for tasks. Add any costs after the link:
 
-This is a list of bulleted links...
+[task name](/link) £10 to £20
+[task name](/link)
 
-* [Link text](/link/href/value)
-* [Link text](/link/href/value)
+Only use bullets to show when a task has a number of options to choose from:
 
-This is a list of non-bulleted links...
-
-- [Link text](/link/href/value)
-- [Link text](/link/href/value)
-
+- [download form option 1](/link)
+- [download form option 2](/link)
+- bullet with no link
 </pre>
--- a/app/views/shared/steps/_page_title.html.erb
+++ b/app/views/shared/steps/_page_title.html.erb
@@ -1,0 +1,7 @@
+<% page_title = '' %>
+
+<% links.each do |link| %>
+  <% page_title = "#{link[:text]} | #{page_title}" %>
+<% end %>
+
+<% content_for :page_title, page_title %>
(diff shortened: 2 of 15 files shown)


app/views/shared/steps/_markdown_help.html.erb
```diff
@@ -1,14 +1,14 @@
...
-This is a list of non-bulleted links...
+Non-bulleted links are used for sequential tasks.
 
 - [Link text](/link/href/value)
 - [Link text](/link/href/value)
 
+Bulleted links are used when users have to choose from a list of options:
+
+* [Link text](/link/href/value)
+* [Link text](/link/href/value)
```
Davidslv: can you add an example that has context please?
and also one that has a bullet but no link `- Hello World` `* Hello World`

app/views/shared/steps/_page_title.html.erb
```diff
@@ -0,0 +1,9 @@
+<% page_title = '' %>
+
+<% if links.length %>
```
Davidslv: please don't do this, it will always evaluate to true, what you looking for is `if links.any?`

app/views/steps/_form.html.erb
```diff
@@ -84,30 +87,48 @@
   </div>
 </div>
 
-<h3>"Get help completing this step" section</h3>
+<% has_help_section = !(step.optional_heading.blank? or step.optional_contents.blank?) %>
```
Davidslv: this should be a helper method inside `app/helpers/step_helpers.rb` 

```ruby
def has_help_section?(step)
  !(step.optional_heading.blank? || step.optional_contents.blank?
end
```

app/views/step_by_step_pages/_form.html.erb
```diff
@@ -6,7 +6,10 @@
 <%= form_for(@step_by_step_page) do |form| %>
   <% if step_by_step_page.errors.any? %>
     <% error_explanation = capture do %>
-      <h2 class="h4">There are <%= pluralize(step_by_step_page.errors.count, "error") %></h2>
+      <h2 class="h4">
+        There <%= step_by_step_page.errors.count == 1 ? "is" : "are" %>
```
Davidslv: You can extract this to a helper method and use it in both places where you need to show error messages

app/views/shared/steps/_page_title.html.erb
selfthinker: Actually, in this case the `if` isn't necessary at all. I will remove it.
I have copied this from the existing `_step_breadcrumb.html.erb`. I can change it in there.

app/views/step_by_step_pages/_form.html.erb
selfthinker: Do you think it's worth it extracting just that line?
I was thinking of rather creating a partial for the whole snippet to reduce duplication.

app/views/shared/steps/_markdown_help.html.erb
selfthinker: I've added the new content-reviewed version.

app/views/step_by_step_pages/_form.html.erb
selfthinker: In which helper would that need to be? Because it is used both for `step_by_step_pages` as well as `steps`. Should it be in the `application_helper`?
I wonder if it makes more sense to rewrite the heading so that wouldn't be necessary... ;-)

selfthinker: I have now fixed this by rewriting the heading, but also moved everything into a partial.
We never use headings as specific as "There are 2 errors" elsewhere. Our guidance on theses error summary headings is not very clear but something like "There is a problem" (to which I have changed this heading) is most commonly used, as far as I know.

selfthinker:
Thanks @Davidslv. I have now fixed all the things and added one more content change ("Improve guidance links for step nav") and retrofitted more content changes to the Markdown example snippet.

app/views/step_by_step_pages/_form.html.erb
```diff
@@ -4,22 +4,7 @@
...
-      <% end %>
-      </ul>
-    <% end %>
-
-    <%= render "shared/steps/form_notice",
-      status: "danger",
-      message: error_explanation
-    %>
-  <% end %>
+  <%= render "shared/steps/form_errors", page: step_by_step_page %>
```
Davidslv: I would call it a (maybe) a `resource` instead of a page as neither `step_by_step_page` and `step` are pages but models/resources.

selfthinker: That makes sense. Renamed.

sihugh:
I'm dismissing @Davidslv 's review as I think all the comments have been addressed and we want to get it merged.

PR #338, opened 2018-03-09
Handle bullets in the publisher tool markdown to align it with govspeak

Davidslv:
We want to accommodate content designers to continue writing their markdown
the same way was they have so far with Govspeak.

This means that both `-` and `*` can be used to indicate bulleted links and also accommodates when publishers want to use bullet points without any links (just text).

It also removes LINK_CAPTURE_REGEX constant to
enable taking advantage of named capture groups in regular expressions
to use named backreferences. This is possible when the regular
expression is on the left hand side of the operator `=~`, by doing so it
assigns the captured text to local variables with the corresponding
names. [Regex Docs](https://ruby-doc.org/core-2.1.1/Regexp.html)

Trello: https://trello.com/c/2zhfNYfj


sihugh:
Text only list items require this to be merged: https://github.com/alphagov/govuk-content-schemas/pull/735

PR #344, opened 2018-03-13
Add SlugValidator to StepByStepPage model

deborahchua:
The Modelling Services team have been working on implementing the `Step by step` publishing tool.

To prevent users from creating slugs that have already been taken, we are using the SlugValidator as part of our validations on the `StepByStepPage` model. 

Trello card: [Add SlugValidator in Collections Publisher](https://trello.com/c/daxHC9M9/549-add-slugvalidator-in-collections-publisher)

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -2,9 +2,7 @@ class StepByStepPage < ApplicationRecord
   has_many :steps, -> { order(position: :asc) }, dependent: :destroy
 
   validates :title, :slug, :introduction, :description, presence: true
-  validates :slug, format: { with: /\A([a-z0-9]+-)*[a-z0-9]+\z/ }
-  validates :slug, uniqueness: true
-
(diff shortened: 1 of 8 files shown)


app/models/step_by_step_page.rb
```diff
@@ -5,6 +5,9 @@ class StepByStepPage < ApplicationRecord
   validates :slug, format: { with: /\A([a-z0-9]+-)*[a-z0-9]+\z/ }
   validates :slug, uniqueness: true
 
+  include ActiveModel::Validations
+  validates_with SlugValidator
```
sihugh: I think we can replace the separate slug validation lines with something like:
`validates :slug, format: { with: /\A([a-z0-9]+-)*[a-z0-9]+\z/ }, uniqueness: true, slug: true`

The `slug: true` bit works out that there's a validator called `SlugValidator`

Davidslv:
@sihugh I believe I've addressed your comment.

PR #356, opened 2018-03-21
Steps use position field to reorder steps

deborahchua:
The Modelling Services team are implementing the Step by step publishing tool into Collections Publisher.

We want to be able to reorder the `steps` of a `Step by step page` using the `position` field.

When more than one step is created and they are reordered, some javascript updates the order of the steps and populates a hidden field in the form containing an array of step `id` and `position` values. The `position` for each step is then updated in the database.

Trello card: [Page to reorder the steps](https://trello.com/c/IXGb9WJG/548-page-to-reorder-the-steps)

screen shot 2018-03-21 at 12 25 54
screen shot 2018-03-21 at 12 24 24


app/views/step_by_step_pages/reorder.html.erb
```diff
@@ -0,0 +1,67 @@
...
+
+<div class="row">
+  <div class="col-md-7">
+    <p>Use the up/down controls on the right to reorder steps. Step numbers will be automatically updated on save.</p>
+
+    <%
+      step_number = 1
+      steps_length = @step_by_step_page.steps.length - 1
+    %>
+    <% @step_by_step_page.steps.each_with_index do |step, step_index| %>
```
sihugh: Could we add a message here if you have no steps?  It looks a bit weird otherwise.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -23,6 +23,7 @@
   <li><%= link_to 'Add a new step', new_step_by_step_page_step_path(@step_by_step_page), class: "btn btn-primary" %></li>
   <li><%= link_to 'Preview', preview_url(@step_by_step_page.slug), class: "btn btn-primary" %></li>
   <li><%= link_to 'Edit title, intro or slug', edit_step_by_step_page_path(@step_by_step_page), class: "btn btn-default" %></li>
+  <li><%= link_to("Reorder steps", step_by_step_page_reorder_path(@step_by_step_page), class: "btn btn-default") %></li>
```
sihugh: Could we hide this if you don't have any steps?

deborahchua:
Thanks @sihugh, I have addressed these comments now 😄

sihugh:
Maybe we should hold this until we're back in the office now @deborahchua

PR #365, opened 2018-03-27
Prevent long titles in step nav wrapping buttons

selfthinker:
Before: 
screen shot 2018-03-27 at 12 07 19

After: 
screen shot 2018-03-27 at 12 07 39

[Trello card](https://trello.com/c/pAa4JrrZ)

### DIFF
--- a/app/views/step_by_step_pages/show.html.erb
+++ b/app/views/step_by_step_pages/show.html.erb
@@ -59,8 +59,8 @@
             </th>
 
             <td class="text-right text-nowrap">
-            <%= link_to("Edit", edit_step_by_step_page_step_path(@step_by_step_page.id, step.id), class: "btn btn-primary btn-sm") %>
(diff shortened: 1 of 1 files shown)


Davidslv:
Maybe a rebase will help get rid of that failure and we can merge it after.

selfthinker:
Shouldn't it be fine to just run it again? (I cannot do that right now as I'm working from home.)

selfthinker:
The main reason for a narrower table is because there are not many columns. Bootstrap always stretches the table to full width and when there are titles that are not that long, it can cause issues with screen magnifier users because they could miss that there is anything on the right-hand side.
It generally makes sense to keep an action as close to their "label" as possible. I guess it depends on how often we'll have long vs short titles. But so far it looks like short titles rule.

sihugh:
👍

PR #368, opened 2018-03-28
[DO NOT MERGE] Rules page frontend

Davidslv:
This commit saves the option of showing or not showing the step pages on
the sidebar of a content item.

- add routes for navigation-rules
- Improves readability of page_title shared component
- Add button for navigation rules
- Form for navigation rules
- Improve copy and design of step nav rules
- Saves navigation rules

The explanation for the hidden field with the same name as the check_box can be [read here in the rails documentation][rails-documentation].

Also the best way at the moment to bypass the "GenericFormBuilder" gem is to explicitly pass it as an option of `form_for` like so:

```
  form_for(..., builder: ActionView::Helpers::FormBuilder)
```
[rails-documentation]: https://github.com/rails/rails/blob/44a9aedd7b8d65517b15bbbb7729f3f16991e23f/actionpack/lib/action_view/helpers/form_helper.rb#L788

It sorts the content_items alphabetically in the navigation rules page.

Add two scopes into NavigationRule model
 - base_paths_part_of_step_nav
 - base_paths_related_to_step_nav

base_paths_related_to_step_nav: These are the content_items that are related to the StepByStepPage but are not showing it on the sidebar of their pages, it will show the related items instead.

base_paths_part_of_step_nav: These are the content_items that will have the StepByStepPage showing on the sidebar.

The reason to have both keys is to keep a relationship between the content_items and the StepByStepPage without having to affect the sidebar when not wanted.

Trello: https://trello.com/c/0DUVntqs

### DIFF
--- a/app/controllers/navigation_rules_controller.rb
+++ b/app/controllers/navigation_rules_controller.rb
@@ -23,6 +23,7 @@ class NavigationRulesController < ApplicationController
   end
 
 private
+
   def set_step_by_step_page
     @step_page = StepByStepPage.find(params[:step_by_step_page_id])
   end
--- a/app/controllers/navigation_rules_controller.rb
+++ b/app/controllers/navigation_rules_controller.rb
@@ -1,0 +1,29 @@
+class NavigationRulesController < ApplicationController
+  before_action :require_gds_editor_permissions!
+  before_action :set_step_by_step_page, only: %i(edit update)
+
+  def edit
+    @step_page.navigation_rules
+  end
+
+  def update
+    navigation_rules = params.delete(:navigation_rules)
+
+    navigation_rules.each_pair do |content_id, value|
+      value = %w(true false).include?(value) ? value : false
+
+      rule = @step_page.navigation_rules.find_by(content_id: content_id)
+      rule.update_attribute(:include_in_links, value) if rule
+    end
+
+    redirect_to(
+      step_by_step_page_navigation_rules_path(@step_page),
+      notice: 'Your choices have been saved.'
+    )
+  end
+
+private
+  def set_step_by_step_page
+    @step_page = StepByStepPage.find(params[:step_by_step_page_id])
+  end
+end
--- a/app/presenters/step_nav_presenter.rb
+++ b/app/presenters/step_nav_presenter.rb
@@ -73,15 +73,31 @@ private
 
   def edition_links
     {
-      "pages_part_of_step_nav": parsed_edition_links
+      "pages_part_of_step_nav": pages_part_of_step_nav,
+      "pages_related_to_step_nav": pages_related_to_step_nav,
     }
   end
 
-  def parsed_edition_links
-    StepNavPublisher.lookup_content_ids(parsed_base_paths).values
+  def base_paths_pages_part_of_step_nav
+    @base_paths_pages_part_of_step_nav ||=
+      parsed_base_paths - step_nav.navigation_rules.base_paths_part_of_step_nav
+  end
+
+  def base_paths_pages_related_to_step_nav
+    @base_paths_pages_related_to_step_nav ||=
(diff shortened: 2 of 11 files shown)


selfthinker:
I've been through this with our content designer Jon and made some more improvements to the content (which I added to the "Navigation Rules Frontend" commit).

app/presenters/step_nav_presenter.rb
```diff
@@ -73,15 +73,31 @@ def steps
...
-  def parsed_edition_links
-    StepNavPublisher.lookup_content_ids(parsed_base_paths).values
+  def base_paths_pages_part_of_step_nav
+    @base_paths_pages_part_of_step_nav ||=
+      parsed_base_paths - step_nav.navigation_rules.base_paths_part_of_step_nav
+  end
+
+  def base_paths_pages_related_to_step_nav
+    @base_paths_pages_related_to_step_nav ||=
+      (parsed_base_paths - step_nav.navigation_rules.base_paths_related_to_step_nav) - base_paths_pages_part_of_step_nav
```
sihugh: Can this be just `step_nav.navigation_rules.base_paths_related_to_step_nav`?

app/presenters/step_nav_presenter.rb
```diff
@@ -73,15 +73,31 @@ def steps
...
+      "pages_part_of_step_nav": pages_part_of_step_nav,
+      "pages_related_to_step_nav": pages_related_to_step_nav,
     }
   end
 
-  def parsed_edition_links
-    StepNavPublisher.lookup_content_ids(parsed_base_paths).values
+  def base_paths_pages_part_of_step_nav
+    @base_paths_pages_part_of_step_nav ||=
+      parsed_base_paths - step_nav.navigation_rules.base_paths_part_of_step_nav
```
sihugh: Should this be `parsed_base_paths - step_nav.navigation_rules.base_paths_related_to_step_nav`?

app/views/step_by_step_pages/index.html.erb
```diff
@@ -35,12 +35,13 @@
...
-            <li><%= link_to 'Edit content', step_by_step_page %></li>
+            <li><%= link_to 'Edit', step_by_step_page %></li>
+            <li><%= link_to 'Choose navigation', step_by_step_page_navigation_rules_path(step_by_step_page) %></li>
             <% if step_by_step_page.has_draft? %>
-              <li><%= link_to 'Preview', preview_url(step_by_step_page.slug) %></li>
+                <li><%= link_to 'Preview', preview_url(step_by_step_page.slug) %></li>
             <% end %>
             <% unless step_by_step_page.has_been_published? %>
-              <li><%= link_to 'Delete', step_by_step_page, method: :delete, data: { confirm: 'Delete this step by step page?' } %></li>
+                <li><%= link_to 'Delete', step_by_step_page, method: :delete, data: { confirm: 'Are you sure?' } %></li>
```
sihugh: I wonder whether we should keep `Delete this step by step page?` as we want people to be doubly sure when they're doing a destructive thing. (I know this was a merge thing and not actually you :-) )

Davidslv: I don't know the answer on this one

app/presenters/step_nav_presenter.rb
Davidslv: Good point

app/controllers/navigation_rules_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      rule = @step_page.navigation_rules.find_by(content_id: content_id)
+      rule.update_attribute(:include_in_links, value) if rule
+    end
+
+    redirect_to(
+      step_by_step_page_navigation_rules_path(@step_page),
+      notice: 'Your choices have been saved.'
+    )
+  end
```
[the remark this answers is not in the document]
Davidslv: Ah! I was thinking it would be done on publishing, I see!

sihugh:
Closing as this was absorbed into #379 .  Thanks @Davidslv

PR #426, opened 2018-07-16
Allow publishers to view draft step-by-steps without signon

leenagupte:
Trello: https://trello.com/c/9ZMVHe0u

## Motivation
Nearly all of the step-by-steps that are created need to be reviewed by people in departments who don't have signon access.  Content designers have been getting around this limitation by following a complicated process to create a preview app to send to factcheckers, but this process will become harder to support as more and more step-by-steps are created.

There is already a pattern used by Mainstream and Whitehall publishers that allows a list of `auth_bypass_ids` to be passed into the payload sent to publishing-api. This id is used by [authenticating-proxy](https://github.com/alphagov/authenticating-proxy) to decode a query string passed into the URL of the draft step-by-step. If the ids match, the user can view the content, otherwise, they will be asked to sign in.

## Tech debt
The code used to create the auth_bypass_id and JWT token passed into the query string are duplicated from Publisher:

* https://github.com/alphagov/publisher/blob/master/app/helpers/paths_helper.rb#L30
* https://github.com/alphagov/publisher/blob/master/app/models/edition.rb

A [card](https://trello.com/c/GcmcoPY0) has been added to the backlog to address this.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,7 +7,24 @@ module ApplicationHelper
     Plek.new.external_url_for('content-tagger')
   end
 
-  def preview_url(slug)
-    "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
+  def preview_url(slug, auth_bypass_id: nil)
+    url = "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
+
+    if auth_bypass_id.present?
+      token = jwt_token(auth_bypass_id)
+      url << "?token=#{token}"
+    end
+
+    url
+  end
+
+private
+
+  def jwt_token(auth_bypass_id)
+    JWT.encode({ 'sub' => auth_bypass_id }, jwt_auth_secret, 'HS256')
+  end
+
+  def jwt_auth_secret
+    Rails.application.config.jwt_auth_secret
   end
 end
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -41,6 +41,21 @@ class StepByStepPage < ApplicationRecord
     redirect_url =~ regex
   end
 
+  # This function first generates a hash from the object id,
+  # then unpacks it to integers,
+  # and then finally interpolates them into a string.
+  # The purpose is to generate a deterministic hash
+  # It mostly rips off SecureRandom.uuid
+  # see http://ruby-doc.org/stdlib-1.9.3/libdoc/securerandom/rdoc/SecureRandom.html#uuid-method
+  def auth_bypass_id
+    @_auth_bypass_id ||= begin
+      ary = Digest::SHA256.hexdigest(content_id.to_s).unpack('NnnnnN')
+      ary[2] = (ary[2] & 0x0fff) | 0x4000
+      ary[3] = (ary[3] & 0x3fff) | 0x8000
+      "%08x-%04x-%04x-%04x-%04x%08x" % ary
+    end
+  end
+
 private
 
   def generate_content_id
--- a/app/presenters/step_nav_presenter.rb
+++ b/app/presenters/step_nav_presenter.rb
@@ -6,6 +6,7 @@ class StepNavPresenter
 
   def render_for_publishing_api(publish_intent = PublishIntent.minor_update)
     payload = required_fields
+    payload.merge!(optional_fields)
     payload.merge(publish_intent.present)
   end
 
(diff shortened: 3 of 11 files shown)


app/models/step_by_step_page.rb
```diff
@@ -41,6 +41,21 @@ def self.validate_redirect(redirect_url)
...
 
+  # This function first generates a hash from the object id,
+  # then unpacks it to integers,
+  # and then finally interpolates them into a string.
+  # The purpose is to generate a deterministic hash
+  # It mostly rips off SecureRandom.uuid
+  # see http://ruby-doc.org/stdlib-1.9.3/libdoc/securerandom/rdoc/SecureRandom.html#uuid-method
+  def auth_bypass_id
+    @_auth_bypass_id ||= begin
+      ary = Digest::SHA256.hexdigest(id.to_s).unpack('NnnnnN')
```
tijmenb: I wonder about `id.to_s` - since this is just a numeric ID, might there be clashes with other preview URLs? Say that another app also has a record with `17` as ID?

leenagupte: Yeah, that occurred to me too. I wondered whether it would still work with a content_id?

tijmenb: To be honest, I don't really understand this part. If `content_id` would work, why would this have the complicated `z"%08x-%04x-%04x-%04x-%04x%08x" % ary` stuff at the bottom?

Could it be just `SecureRandom.uuid` instead?

tijmenb: No Tijmen, you can't use `SecureRandom.uuid` because that'll change every time. This method is built to generate a _predictable_ string that is the same every time. So `content_id` will work as well.

leenagupte: I'll change it to use the content_id.

leenagupte: @tijmenb Updated and fixed-up

sihugh:
Publisher's approach is hooked to the edition ID which should mean that the token is different for a new draft of a published version.

I think this will generate a single auth token which will be the same for all drafts of each step by step (as it's based on the content ID).  Is this OK?

leenagupte:
@sihugh Do step-by-steps have an equivalent to an Edition?

sihugh:
Not yet! ;-).  I'm fine with it if you are, it's just worth being aware
that this is the behaviour

app/helpers/application_helper.rb
```diff
@@ -7,7 +7,24 @@ def content_tagger_url
     Plek.new.external_url_for('content-tagger')
   end
 
-  def preview_url(slug)
-    "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
+  def preview_url(slug, auth_bypass_id: nil)
```
tijmenb: Could we can extract the JWT generation into a separate class?

app/models/step_by_step_page.rb
```diff
@@ -41,6 +41,21 @@ def self.validate_redirect(redirect_url)
     redirect_url =~ regex
   end
 
+  # This function first generates a hash from the object id,
```
tijmenb: This method can perhaps also be extracted into a separate class. Would be good to document our discussion about what the input for `Digest::SHA256.hexdigest()` is.

app/helpers/application_helper.rb
```diff
@@ -7,7 +7,24 @@ def content_tagger_url
...
+      url << "?token=#{token}"
+    end
+
+    url
+  end
+
+private
+
+  def jwt_token(auth_bypass_id)
+    JWT.encode({ 'sub' => auth_bypass_id }, jwt_auth_secret, 'HS256')
```
tijmenb: The `jwt_auth_secret` uses a bit of indirection. Would it be simpler to use the ENV variable here directly?

app/models/step_by_step_page.rb
leenagupte: I can put our discussion into the commit message.

tijmenb: I think it would be better to rewrite the comment to be clearer, "The purpose is to generate a deterministic hash" isn't enough because it doesn't explain why.

app/helpers/application_helper.rb
tijmenb: OK, but I meant more that `Rails.application.config.jwt_auth_secret` is a bit of an indirection. It's two layers of environment specific configuration. If you move this into a class I'd consider dropping some.

leenagupte:
@sihugh @tijmenb I've addressed all of your suggestions in https://github.com/alphagov/collections-publisher/pull/427

PR #427, opened 2018-07-17
Allow publishers to view draft step-by-steps without signon

leenagupte:
Trello: https://trello.com/c/9ZMVHe0u

## Motivation
Nearly all of the step-by-steps that are created need to be reviewed by people in departments who don't have signon access.  Content designers have been getting around this limitation by following a complicated process to create a preview app to send to factcheckers, but this process will become harder to support as more and more step-by-steps are created.

There is already a pattern used by Mainstream and Whitehall publishers that allows a list of `auth_bypass_ids` to be passed into the payload sent to publishing-api. This id is used by [authenticating-proxy](https://github.com/alphagov/authenticating-proxy) to decode a query string passed into the URL of the draft step-by-step. If the ids match, the user can view the content, otherwise, they will be asked to sign in.

## Tech debt
The code used to create the auth_bypass_id and JWT token passed into the query string are duplicated from Publisher:

* https://github.com/alphagov/publisher/blob/master/app/helpers/paths_helper.rb#L30
* https://github.com/alphagov/publisher/blob/master/app/models/edition.rb

A [card](https://trello.com/c/GcmcoPY0) has been added to the backlog to address this.

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -16,4 +16,8 @@ module Services
       disable_cache: true
     )
   end
+
+  def self.draft_origin
+    @draft_origin ||= Plek.new.external_url_for('draft-origin')
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,6 +8,6 @@ module ApplicationHelper
   end
 
   def preview_url(slug)
-    "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
+    "#{Services.draft_origin}/#{slug.sub(/^\//, '')}"
   end
 end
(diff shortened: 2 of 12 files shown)


app/lib/services.rb
```diff
@@ -16,4 +16,8 @@ def self.content_store
       disable_cache: true
     )
   end
+
+  def self.draft_origin
+    @draft_origin ||= Plek.new.external_url_for('draft-origin')
```
sihugh: This feels a little weird.  The other services are things to interact with rather than pieces of configuration.  I'd be happier with the inline Plek call in `preview_url` tbh.

leenagupte:
@tijmenb @sihugh Finally attempt here: https://github.com/alphagov/collections-publisher/pull/428

It's an amalgam of PRs #426 and #427 🤞

app/lib/services.rb
leenagupte: The next set of changes in #428 removes the need for this.

PR #433, opened 2018-07-26
Add status to step by step index page

andysellick:
- adds a status for a step by step navigation to the step by step navigation publishing interface
- status can be draft, published, or unpublished changes (i.e. has been published and since edited)

screen shot 2018-07-26 at 11 26 27

Trello cards: 

- https://trello.com/c/lgxjhPT6/750-display-status-of-step-by-step-on-index-page
- https://trello.com/c/1hRIjqY9/578-improve-search-and-filtering-functionality

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -41,6 +41,12 @@ class StepByStepPage < ApplicationRecord
     redirect_url =~ regex
   end
 
+  def status
+    return unpublished_status if has_draft? && has_been_published?
+    return live_status if has_been_published?
+    draft_status
+  end
+
   # Create a deterministic, but unique token that will be used to give one-time
   # access to a piece of draft content.
   # This token is created by using an id that should be unique so that there is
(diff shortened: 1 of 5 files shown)


app/views/step_by_step_pages/index.html.erb
```diff
@@ -55,6 +56,20 @@
         <td>
           <%= step_by_step_page.slug %>
         </td>
+        <td class="text-right">
+          <%
+            status = "Draft"
+            status = "Live" if step_by_step_page.has_been_published?
+            status = "Unpublished changes" if step_by_step_page.has_draft? and step_by_step_page.has_been_published?
```
leenagupte: Perhaps at the least extract this line to a method.  e.g. `unpublished_changes`

app/models/step_by_step_page.rb
```diff
@@ -41,6 +41,12 @@ def self.validate_redirect(redirect_url)
     redirect_url =~ regex
   end
 
+  def status
+    return unpublished_status if has_draft? && has_been_published?
```
leenagupte: This looks good 💯 

I would fix-up the bits of this commit that get and show the status to the previous commit. That way this commit only has changes related to adding the filter.

It will make the code easier to trace later on when we'll invariably have to update it.

app/models/step_by_step_page.rb
```diff
@@ -61,4 +67,28 @@ def auth_bypass_id
...
+    {
+      name: "live",
+      text: "Live",
+      label_class: "label-success"
+    }
+  end
+
+  def unpublished_status
+    {
+      name: "unpublished",
```
leenagupte: This name could be confusing. We have an "unpublish" button which means something different.

PR #445, opened 2018-08-07
Add linting

leenagupte:
https://trello.com/c/hAI18YOD

### DIFF
--- a/app/assets/stylesheets/curated_lists.scss
+++ b/app/assets/stylesheets/curated_lists.scss
@@ -1,5 +1,14 @@
+$light-grey: #f1f1f1;
+$grey: #ccc;
+$dark-grey: #333;
+$white: #fff;
+$light-yellow: #ffff99;
+
 // Hide the new item form.
-html.js {
+.js {
+  // scss-lint:disable SelectorFormat
+  // to fix [W] SelectorFormat: Selector `new_list_item` should be written in hyphenated BEM (Block Element Modifier) format
+  // @TODO: Investigate and remove element if element not used.
   .new_list_item {
     display: none;
   }
@@ -9,7 +18,7 @@ html.js {
   }
 }
 
-#all-list-items {
+.all-list-items {
   // Remove-buttons have no use in the uncategorized section, but we would like
   // to have them appear when dragged to a list.
   .removal-button form {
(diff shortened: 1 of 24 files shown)


app/assets/stylesheets/curated_lists.scss
```diff
@@ -1,5 +1,14 @@
...
+$white: #fff;
+$light-yellow: #ffff99;
+
 // Hide the new item form.
-html.js {
+.js {
+  // scss-lint:disable SelectorFormat
+  // to fix [W] SelectorFormat: Selector `new_list_item` should be written in hyphenated BEM (Block Element Modifier) format
+  // @TODO: Investigate and remove element if element not used.
   .new_list_item {
```
andysellick: It's in [curated_lists.js](https://github.com/alphagov/collections-publisher/blob/master/app/assets/javascripts/curated_lists.js#L70) but I can't see it anywhere else. It seems like that JS would fail if it can't find it, but I'm not sure if that JS is even running.

app/assets/stylesheets/views/mainstream_browse_pages.scss
```diff
@@ -1,12 +1,17 @@
-/* import govuk_admin_template to get bootstrap styles */
+// import govuk_admin_template to get bootstrap styles
 @import 'govuk_admin_template';
 
-table.tags-list {
+$light-grey: #eee;
+$grey: #888;
```
andysellick: Is it confusing that in one file we declare `$grey` as one colour, but declare it as another colour in another file?

app/assets/stylesheets/views/mainstream_browse_pages.scss
```diff
@@ -65,6 +72,7 @@ h1 .type {
 }
 
 .attributes {
+  // scss-lint:disable PlaceholderInExtend
   @extend .panel;
   @extend .panel-default;
```
andysellick: Whenever you call `scss-lint:disable (rule)` I think that means that particular rule is disabled from that line until either the end of the file or if you add `scss-lint:enable (rule)`. As general good practice we should isolate these disables by wrapping only the specific exception in a disable/enable block.

app/assets/stylesheets/views/mainstream_browse_pages.scss
leenagupte: I'm happy for this to go as is to set-up linting, but create another task to consolidation the colour scheme.

app/assets/stylesheets/views/mainstream_browse_pages.scss
DilwoarH: because this is on the block it only disables it on that block nothing else. https://github.com/brigade/scss-lint#disabling-linters-via-source

andysellick: That's a good point - I didn't realise that. However I think there's a few other examples in this file where we should be closing the lint disabler after we've opened it, because it's applying to more than just a small chunk like this.

DilwoarH:
@andysellick - I've updated it to disable linting line by line rather than on a block level

app/assets/stylesheets/forms.scss
```diff
@@ -1,14 +1,15 @@
...
+.new-list {
   margin-bottom: 20px;
 }
 
 .btn-submit {
-  @extend .btn;
-  @extend .btn-lg;
-  @extend .btn-success;
+  // @TODO: remove extending classes and just add class to element
+  @extend .btn; // scss-lint:disable PlaceholderInExtend
```
andysellick: Can you clarify what this does? Does it disable PlaceholderInExtend for the duration of this line (line 10), everything that follows this line, or does it apply to everything that is pulled in from the @extend?

DilwoarH: Hey - it will only disable the linting for that specific line, it will not disable linting for everything pulled in from the `@extend`. 

Ref: https://github.com/brigade/scss-lint#disabling-linters-via-source

PR #446, opened 2018-08-07
Changes schema to include LinkReport model

jonodrew:
This is a pull request that forms part of a link-checking feature for collections-publisher. The aim is for publishers to be able to check that links in their step-by-step guides are still alive. We're going to batch up all the links in a Step and send them to link-checker-api, collect the returned batch_id and store it against the Step, and then retrieve that batch by calling the API again when the batch is complete..
This pull request changes the schema and adds a new table to the database, called LinkReports. It also adds a relationship to the Step table. It sets up the branch for application changes.

### DIFF
--- a/app/models/link_report.rb
+++ b/app/models/link_report.rb
@@ -1,0 +1,3 @@
+class LinkReport < ApplicationRecord
+  belongs_to :step
+end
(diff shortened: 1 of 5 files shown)


leenagupte:
@jonodrew Can you change the title to of the commit to something shorter so that it fits on one line?

The convention is that the title should be no more than 50 characters, and each line no more than 72 characters.

PR #447, opened 2018-08-07
Display summary of broken links on edit steps page

jonodrew:
This commit adds a method to the Step model. It retrieves the most recent batch_id associated with the Step. Then it retrieves the full batch report from the link-checker-api service. Any broken links are returned to the frontend as an array of Hashes with the same structure as a [`BatchReport`](https://github.com/alphagov/link-checker-api/blob/master/docs/api.md#batchreport-entity)

We're only saving the batch_id for efficiency. There's a tradeoff between storing the whole batch locally against fetching it over the wire, but I've tried to keep the data in one place rather than replicating it in link-checker-api and in collections-publisher.

[Ticket](https://trello.com/c/yZLz7LTB/765-display-the-broken-link-report-in-the-edit-step-page)

### DIFF
--- a/spec/models/step_spec.rb
+++ b/spec/models/step_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
+require 'gds_api/test_helpers/link_checker_api'
 
 RSpec.describe Step do
+  include GdsApi::TestHelpers::LinkCheckerApi
   before do
     allow(Services.publishing_api).to receive(:lookup_content_id)
   end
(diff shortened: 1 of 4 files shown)


spec/models/step_spec.rb
```diff
@@ -38,4 +40,52 @@
       expect(step_item.errors).to have_key(:logic)
     end
   end
+  describe 'link reports' do
```
leenagupte: This should be `broken_links`, because that's what is is being tested.

spec/models/step_spec.rb
```diff
@@ -38,4 +40,52 @@
...
+    it 'should return nothing if there are no link reports yet' do
+      expect(step_item.broken_links).to be_nil
+    end
+
+    it 'should contain one item if there are link reports and at least one is broken' do
```
leenagupte: Can you add a test that checks that `nil` is returned if a broken link report has been created, but there aren't any broken links.

You only need to create one link report to test this.

spec/models/step_spec.rb
```diff
@@ -38,4 +40,52 @@
...
+            "suggested_fix": "null"
+          }
+        ]
+        )
+      create(:link_report, batch_id: 1, step_id: step_item.id)
+      create(:link_report, completed: Time.now, batch_id: 2, step_id: step_item.id)
+      broken_link_report = step_item.broken_links.first
+      expect(broken_link_report.fetch('uri')).to eq "https://www.gov.uk/404"
+      expect(broken_link_report.fetch('problem_summary')).to eq "404 error (page not found)"
+      expect(step_item.broken_links.length).to eql 1
```
leenagupte: I think these assertions could be split up amongst multiple tests, so it's clear what piece of test data relates to an expectation.

For example, why does the second link report have a completed time? How is that being used/tested?
I think this test only needs to check that the most recent link report is being queried if more than one link report exists. 

You could check the values of the broken links being returned in another test. For that, you only need to stub one link report.

spec/models/step_spec.rb
```diff
@@ -38,4 +40,59 @@
...
+            "warnings": [],
+            "problem_summary": "null",
+            "suggested_fix": "null"
+          }
+        ]
+      )
+      expect(step_item.broken_links).to eql []
+    end
+
+    it 'should contain one item if there are link reports and at least one is broken' do
```
leenagupte: Is this two tests hiding as one? One test to check that broken links are being returned and one test to check that the most recent link report is being returned?

PR #452, opened 2018-08-10
Adds webhook to receive batch

jonodrew:
This commit adds a webhook that receives a batch from the link-checker-api and updates that batch's completed time. This enables us to search for the most recent batch_id and retrieve that from the link-checker-api. We've implemented a webhook to avoid spamming the link-checker-api.

[Ticket](https://trello.com/c/eK2bYYNM/769-accept-responses-from-link-checker-api-about-the-status-of-a-link-report-s-m)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -47,6 +47,8 @@ Rails.application.routes.draw do
     "/tags/#{params[:tag_id]}/lists"
   }
 
+  patch 'link_report/:id', to: 'link_report#update'
+
   mount GovukAdminTemplate::Engine, at: "/style-guide"
 
   class SidekiqAccessContraint
--- a/spec/controllers/link_report_controller_spec.rb
+++ b/spec/controllers/link_report_controller_spec.rb
@@ -1,0 +1,20 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/link_checker_api'
+
+RSpec.describe LinkReportController do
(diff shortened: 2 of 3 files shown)


config/routes.rb
```diff
@@ -42,6 +42,9 @@
     end
   end
 
+  resources :link_report do
```
leenagupte: I think this should just be a single POST (or PATCH) route, rather than using `resources` and creating seven routes. You can see all of the routes that created by running `rake routes`.

spec/controllers/link_report_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/link_checker_api'
+
+RSpec.describe LinkReportController do
+  include GdsApi::TestHelpers::LinkCheckerApi
+  describe '.update' do
+    it 'should update an existing LinkReport' do
```
leenagupte: What happens if you receive a patch request for a batch_id that doesn't exist?

spec/models/step_spec.rb
```diff
@@ -40,7 +40,7 @@
       expect(step_item.errors).to have_key(:logic)
     end
   end
-  describe 'broken_links' do
+  describe '.broken_links' do
```
leenagupte: Not really needed, but if you do want to edit the test name, can you split it out because it's not related to the other changes in this commit.

PR #454, opened 2018-08-14
[DO NOT MERGE]Batch links

jonodrew:
This PR adds code that parses Step content for links, batches it, and sends it to the link-checker-api. This will allow users to trigger a request for the links in their Step-By-Step to be checked.

[Ticket](https://trello.com/c/nX8IgZ6z/766-request-a-broken-link-report-m)


spec/models/link_report_spec.rb
```diff
@@ -1,5 +1,41 @@
 require 'rails_helper'
+require 'gds_api/test_helpers/link_checker_api'
 
-RSpec.describe LinkReport, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+RSpec.describe LinkReport do
+  include GdsApi::TestHelpers::LinkCheckerApi
+  before do
+    allow(Services.publishing_api).to receive(:lookup_content_id)
+    # allow(Services.link_checker_api).to receive(:create_batch)
```
leenagupte: Can you delete this line if it's not needed?

spec/models/link_report_spec.rb
```diff
@@ -5,7 +5,6 @@
   include GdsApi::TestHelpers::LinkCheckerApi
   before do
     allow(Services.publishing_api).to receive(:lookup_content_id)
-    # allow(Services.link_checker_api).to receive(:create_batch)
   end
```
leenagupte: Can you squash this deletion up to the previous commit?

app/models/link_report.rb
```diff
@@ -17,7 +17,7 @@ def create_batch
 private
   
   def list_contents(list_contents)
-    list_contents.map { |content_item| content_item.fetch(:href) }
+    list_contents.map { |content_item| content_item.fetch(:href) } unless list_contents.nil?
```
leenagupte: Is this change related to adding `request_broken_links`? If not, can you squash it up to the previous commit?

app/models/link_report.rb
```diff
@@ -29,6 +29,7 @@ def generate_links(parsed_content)
     parsed_content.each do |content|
       links << list_contents(content.fetch(:contents)) unless content.fetch(:type, nil) != 'list'
     end
-    links.flatten!
+    links.flatten! unless links.empty?
+    links
```
leenagupte: Is this change related to adding `request_broken_links`? If not, can you squash it up to the previous commit?

spec/models/step_spec.rb
```diff
@@ -124,4 +124,41 @@
...
+      it 'should have an associated LinkReport with a batch id' do
+        link_checker_api_create_batch(
+          uris: [
+            'https://www.gov.uk/good/stuff',
+            "https://www.gov.uk/also/good/stuff",
+            "https://www.gov.uk/not/as/great",
+            "http://example.com/good"
+          ],
+          id: 1234,
+          webhook_uri: "https://collections-publisher.test.gov.uk/link_report"
```
leenagupte: The webhook isn't added until the next commit, so perhaps this line should be added there?

spec/models/step_spec.rb
```diff
@@ -124,4 +124,41 @@
...
+        )
+        step = create(:step)
+        step.request_broken_links
+        expect(step.link_report.last.batch_id).to eql 1234
+      end
+    end
+
+    context 'when there are no links in the step' do
+      it 'should have an associated LinkReport with a batch id' do
```
leenagupte: If there are no links in the step, I wouldn't expect a batch report to be requested from link_checker_api. 

Perhaps the expectation of the test could be simplified to:
`expect(LinkReport).not_to receive(:create_batch)`

spec/models/step_spec.rb
```diff
@@ -124,4 +124,41 @@
       expect(step_item.broken_links.length).to eql 1
     end
   end
+
+  describe '.request_broken_links' do
+    context 'when there are links in the step' do
+      it 'should have an associated LinkReport with a batch id' do
```
leenagupte: This feels like a copy of the test in `spec/models/link_report_spec.rb`. As that test is checking that the LinkReport is saved correctly, the expectation for this test could just be that LinkReport is being called correctly.

Something like:

`expect(step.link_report).to receive(:create_batch)`

app/models/link_report.rb
```diff
@@ -9,19 +9,22 @@ def batch_links
...
+      batch_links,
+      webhook_uri: Plek.new.external_url_for("collections-publisher") + "/link_report"
+      )
     self.batch_id = batch.id
     self.save
   end
 
 private
-  
+
```
leenagupte: Can you fix-up lint errors to the place they were introduced?

app/models/link_report.rb
```diff
@@ -9,19 +9,22 @@ def batch_links
...
   def list_contents(list_contents)
     list_contents.map { |content_item| content_item.fetch(:href) } unless list_contents.nil?
   end
 
   def prefix_govuk(path_to_prefix)
-    "https://www.gov.uk" + path_to_prefix    
+    "https://www.gov.uk" + path_to_prefix
```
leenagupte: Can you fix-up lint errors to the place they were introduced?

spec/models/step_spec.rb
```diff
@@ -147,13 +147,12 @@
     context 'when there are no links in the step' do
       it 'should have an associated LinkReport with a batch id' do
         link_checker_api_create_batch(
-          uris: [
-          ],
+          uris: [],
```
leenagupte: Can you fix-up lint errors to the place they were introduced?

spec/models/step_spec.rb
```diff
@@ -147,13 +147,12 @@
...
           id: 1234,
           webhook_uri: "https://collections-publisher.test.gov.uk/link_report"
         )
         step = create(
-          :step, 
+          :step,
```
leenagupte: Can you fix-up lint errors to the place they were introduced?

spec/models/link_report_spec.rb
```diff
@@ -29,7 +29,8 @@
           "https://www.gov.uk/also/good/stuff",
           "https://www.gov.uk/not/as/great",
           "http://example.com/good"
-        ]
+        ],
+        webhook_uri: "https://collections-publisher.test.gov.uk/link_report"
```
leenagupte: You could move this into its own test, and assert that the correct webhook is passed to link-checker-api.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -122,4 +128,9 @@ def set_step_by_step_page
   def step_by_step_page_params
     params.require(:step_by_step_page).permit(:title, :slug, :introduction, :description)
   end
+
+  def steps_in_step_by_step_page
+    @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
```
leenagupte: `@step_by_step_page` is already being set in `set_step_by_step_page`, so you don't need to do it again here.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -122,4 +128,9 @@ def set_step_by_step_page
   def step_by_step_page_params
     params.require(:step_by_step_page).permit(:title, :slug, :introduction, :description)
   end
+
+  def steps_in_step_by_step_page
```
leenagupte: I don't think you need this function. In `batch` you could just use: 
`@step_by_step_page.steps.each(&:request_broken_links)`

config/routes.rb
```diff
@@ -12,6 +12,7 @@
     post :reorder
     get :unpublish
     post :unpublish
+    post :batch
```
leenagupte: This should be renamed to be more descriptive. What's the batch of?

leenagupte:
I don't know why GitHub has hidden some of the comments, but they are visible if you look at each commit individually.

app/lib/services.rb
```diff
@@ -19,6 +19,7 @@ def self.content_store
   end
 
   def self.link_checker_api
-    @link_checker_api ||= GdsApi::LinkCheckerApi.new(Plek.new.find("link-checker-api"))
+    @link_checker_api ||= GdsApi::LinkCheckerApi.new(
```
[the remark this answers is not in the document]
leenagupte: Ok, then split it. Changes to the formatting of this line appear in a couple of places and can squashed together.

app/lib/services.rb
```diff
@@ -20,6 +20,7 @@ def self.content_store
 
   def self.link_checker_api
     @link_checker_api ||= GdsApi::LinkCheckerApi.new(
-      Plek.new.find("link-checker-api"))
+      Plek.new.find("link-checker-api")
+      )
```
leenagupte: As you're reformatting this chunk, the closing `)` needs to be out-dented by two characters.

PR #457, opened 2018-08-20
Batch links for api

jonodrew:
[Ticket!](https://trello.com/c/nX8IgZ6z/766-request-a-broken-link-report-m)

This PR adds a number of methods to allow users to parse Step content for links, batch them, and then send them to link-checker-api. It includes a rogue commit that hotfixes an issue with the webhook. The issue only came to light when I tried it out for real, and [it's in this commit](https://github.com/alphagov/collections-publisher/commit/78cc154e95976785539c03144aafbd2d9b6f7137).

### DIFF
--- a/app/models/link_report.rb
+++ b/app/models/link_report.rb
@@ -1,3 +1,26 @@
 class LinkReport < ApplicationRecord
   belongs_to :step
+
+  def create_record
+    batch_response = request_batch_of_links
+    if request_batch_of_links.present?
+      self.batch_id = batch_response.id
+      self.save
+    end
+  end
+
+private
+
+  def batch_of_links
+    StepContentParser.new.all_paths(self.step.contents)
+  end
+
+  def request_batch_of_links
+    if batch_of_links.any?
+      Services.link_checker_api.create_batch(
+        batch_of_links,
+        webhook_uri: Plek.new.external_url_for("collections-publisher") + "/link_report"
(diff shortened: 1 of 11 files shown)


app/services/step_content_parser.rb
```diff
@@ -45,12 +45,24 @@ def base_paths(step_text)
...
+  end
+
 private
 
   def relative_paths(content)
     content.scan(LINK_REGEX).select { |href| href[0] =~ /^\/[a-z0-9]+.*/i if href.any? }.flatten
   end
 
+  def absolute_paths(content)
+    content.scan(LINK_REGEX).flatten
```
leenagupte: This method is misleading.  It doesn't just return absolute paths, but a mix of absolute and relative_paths.

app/models/link_report.rb
```diff
@@ -1,3 +1,32 @@
 class LinkReport < ApplicationRecord
   belongs_to :step
+
+  def create_record
+    batch_response = request_batch_of_links
+    if batch_response.nil?
+      self.destroy
```
leenagupte: I don't think you need to do an explicit destroy.
You could check if `batch_response.present?` and only `save` the record if it is.

jonodrew: I've done this explicitly because I think explicit is better than implicit, even if it's not technically necessary.

PR #461, opened 2018-08-22
[WIP] Only show error message if there is a broken link

leenagupte:
Trello: https://trello.com/c/VKiDCA1o

## What's a changed

Only show an error message for broken links on the step overview page if the step has broken links.

### Before
screen shot 2018-08-22 at 12 20 05 copy

### After
screenshot-collections-publisher dev gov uk-2018 08 28-10-00-26

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -4,6 +4,10 @@ class Step < ApplicationRecord
   validates :title, :logic, presence: true
   has_many :link_reports
 
+  def broken_links?
+    broken_links.present? && broken_links.any?
+  end
+
   def broken_links
     collect_broken_links unless most_recent_batch.nil?
   end
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -8,6 +8,10 @@ class Step < ApplicationRecord
     broken_links.present? && broken_links.any?
   end
 
+  def link_report?
+    most_recent_batch.present?
+  end
+
   def broken_links
     collect_broken_links unless most_recent_batch.nil?
   end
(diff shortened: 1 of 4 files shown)


spec/models/step_by_step_page_spec.rb
```diff
@@ -134,6 +137,68 @@
     }.to raise_error(ActiveRecord::RecordNotUnique)
   end
 
+  describe 'broken link checking' do
+    let(:step_by_step_with_step) { create(:step_by_step_page) }
+
+    it 'checks if a link check was requested and there are no broken links' do
```
DilwoarH: @leenagupte - this test title doesn't seem to represent the actual test, I think the title for this and the one below need to be swapped

spec/models/step_by_step_page_spec.rb
```diff
@@ -134,6 +137,68 @@
...
+        ]
+      )
+
+      create(:link_report, batch_id: 1, step_id: step.id)
+
+      expect(step_by_step_with_step.broken_links?).to be true
+      expect(step_by_step_with_step.link_report_requested?).to be true
+    end
+
+    it "checks if a link check was requested and there is at least one broken link" do
```
DilwoarH: same as above

spec/models/step_by_step_page_spec.rb
leenagupte: Good spot! Thanks! Will change and fix-up.

spec/models/step_by_step_page_spec.rb
leenagupte: Good spot! Thanks! Will change and fix-up.

PR #466, opened 2018-08-29
Adds InternalChangeNote model

jonodrew:
This commit adds an InternalChangeNote model with a foreign key on the StepByStepPage model. This enables us to build new functionality to record the changes that content designers have made to a step-by-step.

[Ticket](https://trello.com/c/OvHzWmh0/814-change-database-schema-to-add-changenote-model) and [more context](https://trello.com/c/G0fchnmS/733-ability-to-record-change-notes-l)

## Why
Research with content designers revealed that they like being able to see the changes they've made to a step-by-step. Ultimately this will probably be similar to git versioning, but as an MVP we're just implementing change notes. They'll have an author, a timestamp, and some content, all of which are validated. This will allow content designers to track changes they're making and also externalise knowledge they have about a specific step by step.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,9 +10,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_08_17_100339) do
+ActiveRecord::Schema.define(version: 2018_08_29_153755) do
 
-  create_table "list_items", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+  create_table "internal_change_notes", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.string "author"
+    t.text "description"
+    t.bigint "step_by_step_page_id"
+    t.datetime "created_at"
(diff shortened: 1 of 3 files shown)


cbaines:
The Publishing API has support for change notes, they're normally displayed on the content pages, but they don't have to be. Have you considered using those?

jonodrew:
I've spoken to @cbaines and I think the publishing API is more for external-facing changenotes, while this feature request is for recording changes internal to the team. Because of that I think we need to press ahead with a local database for recording changes.

db/schema.rb
```diff
@@ -10,9 +10,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_08_17_100339) do
+ActiveRecord::Schema.define(version: 2018_08_29_153755) do
 
-  create_table "list_items", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+  create_table "internal_change_notes", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.string "author"
```
cbaines: You might want to consider using the `users` table, which is automatically populated when users login (through the gds-sso gem), rather than just storing the author as a string (assuming this is the author of the note).

PR #473, opened 2018-09-03
Internal changenotes controller

jonodrew:
This PR adds a controller and a form for publishers to add changenotes to their Step-By-Steps. It's a very rough MVP and does not contain a method to retrieve those notes. That method will be added in another PR.

## Context
[Ticket for this PR](https://trello.com/c/OqkwUwN3/818-create-changenote-controller-with-create-action) and [wider context](https://trello.com/c/G0fchnmS/733-ability-to-record-change-notes-l)

## Why
Publishers want a way to record changes they've made to a SBS, or leave notes for people on 2i to read. This is a small PR, because the team is currently working on pushing to master at least once a day as an experiment to find out if it gets value to users faster.

### DIFF
--- a/app/controllers/internal_change_notes_controller.rb
+++ b/app/controllers/internal_change_notes_controller.rb
@@ -1,0 +1,16 @@
+class InternalChangeNotesController < ApplicationController
+  def create
+    InternalChangeNote.create(internal_change_note_params.merge(step_by_step_page_id: step_by_step_page.id, created_at: Time.now, author: current_user.name))
+    redirect_to step_by_step_page
+  end
+
+private
+
+  def step_by_step_page
+    StepByStepPage.find(params[:step_by_step_page_id])
+  end
+
+  def internal_change_note_params
+    params.require(:internal_change_note).permit(:description)
+  end
+end
--- a/spec/controllers/internal_change_notes_controller_spec.rb
+++ b/spec/controllers/internal_change_notes_controller_spec.rb
@@ -1,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe InternalChangeNotesController, type: :controller do
+  before do
+    allow(Services.publishing_api).to receive(:lookup_content_id)
+  end
+
+  describe "POST #create" do
+    it "creates an InternalChangeNote" do
+      stub_user.name = "Test author"
+      create(:step_by_step_page, id: 0)
+      expect(InternalChangeNote).to receive(:create)
+      post :create, params: { step_by_step_page_id: 0, internal_change_note: { description: "Test description" } }
+    end
+    it "saves it to the database" do
+      stub_user.name = "Test author"
+      create(:step_by_step_page, id: 0)
+      post :create, params: { step_by_step_page_id: 0, internal_change_note: { description: "Test description" } }
+      expect(InternalChangeNote.first.author).to eql "Test author"
+    end
+  end
+end
--- a/spec/features/managing_steps_spec.rb
+++ b/spec/features/managing_steps_spec.rb
@@ -63,6 +63,15 @@ RSpec.feature "Managing step by step pages" do
         and_I_see_the_steps_updated_on_the_step_by_step_details_page
       end
     end
+
+    context "and I would like to leave a change note on a step by step page" do
+      scenario "User leaves a change note" do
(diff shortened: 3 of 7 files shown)


app/views/step_by_step_pages/change_notes.html.erb
```diff
@@ -0,0 +1,48 @@
...
+%>
+
+<%= render 'nav', step_by_step_page: @step_by_step_page, active: "change-notes" %>
+
+<%
+  left_col = "col-md-7"
+  right_col = "col-md-5"
+%>
+
+<%= form_for(InternalChangeNote.new, url: internal_change_notes_create_path) do |form| %>
```
leenagupte: Ideally the view should have the same name as the controller. So this file should be called `internal_change_note.html.erb`.

You should be able to create a new instance of the class (`InternalChangeNote.new`) in the controller method, and just refer to it here.

app/controllers/internal_change_notes_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+private
+
+  def step_by_step_page
+    StepByStepPage.find(params[:id])
+  end
+
+  def internal_change_note_params
+    params.require(:internal_change_note).permit(:author, :description)
```
leenagupte: I don't think you need `author` in here.

app/controllers/internal_change_notes_controller.rb
```diff
@@ -0,0 +1,16 @@
+class InternalChangeNotesController < ApplicationController
+  def create
+    InternalChangeNote.create(internal_change_note_params.merge(step_by_step_page_id: step_by_step_page.id, created_at: Time.now, author: current_user.name))
```
leenagupte: Following on from the comment in the view, you are doing a `new` and a `create` for `InternalChangeNote`. Can you create one object and just save that when the form is submitted.

spec/controllers/internal_change_notes_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe InternalChangeNotesController, type: :controller do
```
leenagupte: The controller test is good 👍, can you also add a feature test?

app/views/step_by_step_pages/change_notes.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
+
+  <div class="form-group row">
+    <div class="<%= left_col %>">
+      <%= form.text_area :description, required: true, class: "form-control", rows: 8 %>
+    </div>
+  </div>
+  <%= hidden_field_tag :id, @step_by_step_page.id%>
```
leenagupte: If you add the POST route as a resource on step-by-step-pages, you don't need this hidden field as the `step_by_step_id` will be passed as a param to the page path.

spec/features/managing_steps_spec.rb
```diff
@@ -181,4 +194,21 @@ def and_I_see_the_steps_updated_on_the_step_by_step_details_page
...
+  def and_I_write_a_change_note
+    fill_in "Description", with: "I've changed this step by step!"
+  end
+
+  def and_I_complete_a_change_note
+    and_I_write_a_change_note
+    click_on "Save"
+  end
+
+  def then_I_can_see_the_change_note_page
```
leenagupte: You're not using this step.

spec/features/managing_steps_spec.rb
```diff
@@ -181,4 +194,17 @@ def and_I_see_the_steps_updated_on_the_step_by_step_details_page
     expect(page).to have_css("table > tbody > tr:nth-child(1) > th:nth-child(2)", text: "Dress like the Fonz")
     expect(page).to have_css("table > tbody > tr:nth-child(2) > th:nth-child(2)", text: "Check how awesome you are")
   end
+
+  def and_I_write_a_change_note
+    fill_in "Description", with: "I've changed this step by step!"
+  end
+
+  def and_I_complete_a_change_note
+    and_I_write_a_change_note
```
leenagupte: Not a deal breaker, but I would either move this step into the scenario or consolidate it into `and_I_complete_a_change_note`

PR #488, opened 2018-09-14
Rescue error raised from missing batch report

leenagupte:
Trello: https://trello.com/c/e1vAwREO

Link-checker-api throw a 5xx error if `get_batch` is called for a batch report that doesn’t exist.

This isn’t usually a problem in Production, but step by step pages, and their associated link reports are copied
to integration and staging each night. This means that if when you try to view the step by step in those environments, it’s
querying link-checker-api with a batch id that was created in (and probably only exists in) production. This causes link-checker-api to
fail with a 5xx.

We should update link-checker-api to fail more gracefully if it can’t find a requested batch, but until then, allow collections-publisher
to load step by steps even if the link report details cannot be found.

Error in [sentry](https://sentry.io/govuk/app-collections-publisher/issues/686572073/?query=is:unresolved)

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -27,13 +27,19 @@ class Step < ApplicationRecord
 private
 
   def collect_broken_links
+    return [] if batch_link_report.blank?
+
     batch_link_report.links.keep_if do |link|
       link.fetch('status') == 'broken'
     end
   end
 
   def batch_link_report
-    @batch_link_report ||= Services.link_checker_api.get_batch(batch_link_report_id)
+    begin
+      @batch_link_report ||= Services.link_checker_api.get_batch(batch_link_report_id)
+    rescue GdsApi::HTTPServerError
+      nil
+    end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/step.rb
```diff
@@ -27,13 +27,19 @@ def request_broken_links
...
       link.fetch('status') == 'broken'
     end
   end
 
   def batch_link_report
-    @batch_link_report ||= Services.link_checker_api.get_batch(batch_link_report_id)
+    begin
+      @batch_link_report ||= Services.link_checker_api.get_batch(batch_link_report_id)
+    rescue GdsApi::HTTPServerError
+      []
```
sihugh: It feels a little weird to use an empty array here as it sort of implies that a fully populated `@batch_link_report` would also be an array, which I don't think it is.  Maybe use `nil` instead?

leenagupte:
@sihugh I've fixed up the changes. Cheers!

PR #493, opened 2018-09-18
Adds method to discard change notes without an edition

jonodrew:
# What
Drop all change notes associated with a draft but without an associated edition when that draft is discarded. 

# Why
When a SBS guide is published, it gets an edition number. At that point, we tag all the change notes **without** an edition number to that edition. However, if the draft is discarded, we don't want the change notes to hang around. If we leave them in the table they'll be tagged to the next published edition of that SBS guide. This would result in change notes that refer to a discarded draft being associated with a published guide, and that would cause confusion. 

# How
This has been implemented with a method on the StepByStepPage that is called from the StepByStepPage controller. It drops all the changenotes associated to that StepByStepPage that don't have an edition.

[Ticket](https://trello.com/c/x0k2Fsc6/841-drop-change-notes-for-the-discarded-draft-s)

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -72,6 +72,10 @@ class StepByStepPage < ApplicationRecord
     steps.map(&:link_report?).any?
   end
 
+  def discard_notes
+    internal_change_notes.where(edition_number: nil).delete_all
+  end
+
 private
 
   def generate_content_id
--- a/spec/models/step_by_step_page_spec.rb
+++ b/spec/models/step_by_step_page_spec.rb
@@ -245,4 +245,22 @@ RSpec.describe StepByStepPage do
       end
     end
   end
+
+  describe '.discard_notes' do
+    before(:each) do
+      @step_by_step_page = create(:step_by_step_page)
+      create(:internal_change_note, edition_number: 1, step_by_step_page_id: @step_by_step_page.id)
+      create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
+    end
+    context 'when there are existing change notes with a version and new change notes without a version' do
+      before(:each) do
+        @step_by_step_page.discard_notes
+        @step_by_step_page.reload
+      end
+      it 'should only delete the change notes without an edition' do
+        expect(@step_by_step_page.internal_change_notes.count).to eq 1
+        expect(@step_by_step_page.internal_change_notes.first[:edition_number]).to eql 1
+      end
+    end
+  end
 end


app/controllers/step_by_step_pages_controller.rb
```diff
@@ -97,6 +97,7 @@ def internal_change_notes
 
   def discard_draft
     StepNavPublisher.discard_draft(@step_by_step_page)
+    @step_by_step_page.discard_notes
```
leenagupte: We're missing feature tests for deleting step by step pages. Can you add some?

app/models/step_by_step_page.rb
```diff
@@ -72,6 +72,10 @@ def links_checked?
     steps.map(&:link_report?).any?
   end
 
+  def discard_notes
+    InternalChangeNote.where(step_by_step_page_id: self.id, edition_number: nil).delete_all
```
leenagupte: Does using `internal_change_notes.where(edition_number: nil).delete_all` work here? We shouldn't need to pass in the step_by_step_id as it should already know who it's children are.

spec/models/step_by_step_page_spec.rb
```diff
@@ -245,4 +245,34 @@
...
+    context 'when there are no change notes for this step by step' do
+      it 'the number of change notes associated to this step by step guide without an edition should be 0' do
+        step_by_step_page = create(:step_by_step_page, id: 1)
+        step_by_step_page.discard_notes
+        expect(InternalChangeNote.where(step_by_step_page_id: step_by_step_page.id, edition_number: nil)).to be_empty
+      end
+    end
+
+    context 'when there are untagged change notes for a step by step guide' do
+      it 'there should be no change notes without a version' do
```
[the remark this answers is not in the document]
jonodrew: I don't agree with this comment. I think context makes it more explicit and I'm going to keep it in.

app/models/step_by_step_page.rb
jonodrew: Yes, that works perfectly. Thank you

app/controllers/step_by_step_pages_controller.rb
jonodrew: Should I make a new PR for that, or combine it with this one?

leenagupte: Can you add it as a preparatory commit to this one? Cheers.

jonodrew: As we discussed it's turned out to be a bit thornier than expected. I'll move it to its own PR and ticket.

leenagupte: I think you need to remove this function call from here. As discussed at stand_up, we can instead call `discard_notes` from a `before_destroy` block in the StepByStepPages model, which we're going to tackle as part of the bug card.

spec/models/step_by_step_page_spec.rb
```diff
@@ -245,4 +245,25 @@
...
+      create(:internal_change_note, edition_number: 1, step_by_step_page_id: @step_by_step_page.id)
+      create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
+    end
+    context 'when there are existing change notes with a version and new change notes without a version' do
+      before(:each) do
+        @step_by_step_page.discard_notes
+        @step_by_step_page.reload
+      end
+      # this is necessarily a longer and more complex query because it's searching for a negative
+      it 'should delete the change notes without an edition' do
```
leenagupte: I think you can combine these two tests into one, "it should only delete the change notes without an edition"
and then check that:
1. @step_by_step_page.internal_change_notes.count is 1
2. @step_by_step_page.internal_change_notes.first[:edition_number] is also 1

jonodrew: I don't understand the second test. Is it testing to make sure that the right edition number is allocated to the only change note?

PR #503, opened 2018-09-21
Fix latest edition number bug

leenagupte:
The fix applied in PR #502 didn’t completely fix the problem.

We need to parse the content returned by GDS API before we can apply Hash methods (with_indifferent_access) to it.

Error in [sentry](https://sentry.io/govuk/app-collections-publisher/issues/696647763/?query=is:unresolved)


leenagupte:
@jonodrew They do now!

PR #506, opened 2018-09-24
Adds method to assign current user to current draft

jonodrew:
# What
This commit adds a small method that assigns whoever the current user is to the current draft when it is updated. This happens if a new guide is created or an existing guide is edited in any way. The meat of this commit is updating areas where this method is called and updating the tests as well.

# Why
It will allow users to see who is currently responsible for a draft.

I'm going to extend this in a future PR so that a change note is generated. This will create a "Chain of Custody" that describes who's been assigned the draft.

### DIFF
--- a/spec/support/step_nav_steps.rb
+++ b/spec/support/step_nav_steps.rb
@@ -54,7 +54,7 @@ module StepNavSteps
   end
 
   def expect_update_worker
-    allow(StepByStepDraftUpdateWorker).to receive(:perform_async).with(@step_by_step_page.id)
+    allow(StepByStepDraftUpdateWorker).to receive(:perform_async).with(@step_by_step_page.id, stub_user.name)
   end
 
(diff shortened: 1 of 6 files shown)


leenagupte:
@jonodrew I thought we were only going to do this the first time a draft is opened, not every time a change is saved?

spec/support/step_nav_steps.rb
```diff
@@ -54,7 +54,7 @@ def then_the_page_is_unpublished
   end
 
   def expect_update_worker
-    allow(StepByStepDraftUpdateWorker).to receive(:perform_async).with(@step_by_step_page.id)
+    allow(StepByStepDraftUpdateWorker).to receive(:perform_async).with(@step_by_step_page.id, stub_user.name)
```
leenagupte: You could get to `stub_user.name`! 🥇

PR #508, opened 2018-09-24
Generate changenote when assigned

jonodrew:
# What
This PR is for a small function that runs when a step by step is going into draft. It generates an InternalChangeNote with the text `"Draft created by #{current_user.name}"`. 

# Why
We think that this will help users to keep track of who started the current draft and make communication and traceability easier.

[Ticket](https://trello.com/c/lsMQWc6u/875-add-a-change-note-when-the-draft-is-assigned-to-a-new-user)

### DIFF
--- a/app/workers/step_by_step_draft_update_worker.rb
+++ b/app/workers/step_by_step_draft_update_worker.rb
@@ -24,8 +24,17 @@ class StepByStepDraftUpdateWorker
 
   def update_assigned_to
     unless step_by_step_page.has_draft?
+      generate_internal_change_note
       step_by_step_page.assigned_to = @current_user
       step_by_step_page.save
     end
+  end
+
+  def generate_internal_change_note
+    change_note = step_by_step_page.internal_change_notes.new(
+      author: @current_user,
+      description: "Draft created by #{@current_user}",
+    )
+    change_note.save
   end
 end
--- a/spec/workers/step_by_step_draft_update_worker_spec.rb
+++ b/spec/workers/step_by_step_draft_update_worker_spec.rb
@@ -38,4 +38,23 @@ RSpec.describe StepByStepDraftUpdateWorker do
       end
     end
   end
+
+  describe '#generate_internal_change_note' do
+    let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
+    context 'when the guide is a new draft' do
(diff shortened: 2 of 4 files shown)


spec/workers/step_by_step_draft_update_worker_spec.rb
```diff
@@ -22,12 +22,14 @@
 
   describe '#update_assigned_to' do
     let(:step_by_step_page) { create(:step_by_step_page_with_steps, assigned_to: "Old author") }
+
```
leenagupte: Can you get rid of this line?

spec/workers/step_by_step_draft_update_worker_spec.rb
```diff
@@ -22,12 +22,14 @@
...
     context 'when a guide is a new draft' do
       it 'should assign the draft to the current user' do
         described_class.new.perform(step_by_step_page.id, @current_user.name)
         expect(StepByStepPage.find(step_by_step_page.id).assigned_to).to eql "New author"
       end
     end
+
```
leenagupte: And this one?

app/workers/step_by_step_draft_update_worker.rb
```diff
@@ -24,8 +24,18 @@ def update_draft
...
       step_by_step_page.assigned_to = @current_user
       step_by_step_page.save
     end
   end
+
+  def generate_internal_change_note
+    change_note = step_by_step_page.internal_change_notes.new(
+      author: @current_user,
+      description: "Draft created by #{@current_user}",
+      created_at: Time.now
```
leenagupte: You shouldn't need to pass in the `created_at` time, as this field should be auto-populated by ActiveRecord with the current date and time.

The reason that it is required is, we've added a validation to the `InternalChangeNote` model that forces the `created_at` date to be passed in. If you remove that validation, your tests will pass, and the created_at date will still (always) be updated.

See line: https://github.com/alphagov/collections-publisher/blob/master/app/models/internal_change_note.rb#L2

jonodrew:
@leenagupte thank you for the feedback. I think I've made all the suggested changes. Could I ask for a second look?

jonodrew:
> @jonodrew If you reverse the order of the commits, I'll approve this.

Deal!

PR #509, opened 2018-09-24
ADR about discarding drafts

leenagupte:
Trello: https://trello.com/c/mL5qzXSC


leenagupte:
@VitalieMogoreanu  The change notes are dropped when the draft is discarded. See commit: https://github.com/alphagov/collections-publisher/commit/7992c0e4348062048ed9cffdf4aa977dd7552c5a

We don't populate the edition ID in the change notes until the draft is published, so it's easy to tell which change notes should be dropped.

Publishing-api only stores one draft for an edition. It doesn't matter how many times you edit and save it, the number of drafts remains 1.
I tried to explain that idea in this line: https://github.com/alphagov/collections-publisher/blame/master/docs/arch/adr-002-discarding-drafts.md#L28

PR #512, opened 2018-09-26
Assign user and log change

jonodrew:
# What
These changes:
* display the name of the user who most recently saved a guide
* generate a change note when a different user saves a guide

# Why
After discussing with our content designers we've made these changes to make it easier to track the history of a step by step guide. The automatically generated chain of custody, alongside user-written change notes, should give a rich history of the guide in question. 

[Ticket](https://trello.com/c/KC9JoEZM/845-display-the-name-of-the-user-who-initiated-the-current-draft-m-l)

### DIFF
--- a/app/views/shared/steps/_heading.html.erb
+++ b/app/views/shared/steps/_heading.html.erb
@@ -2,3 +2,7 @@
   <small><%= action %><span class="sr-only"> - </span></small><br />
   <%= step_nav %>
 </h1>
+
+<% if @step_by_step_page.assigned_to.present? %>
+  <%= render 'shared/steps/last-saved-by', assigned_to: @step_by_step_page.assigned_to %>
+<% end %>
--- a/app/workers/step_by_step_draft_update_worker.rb
+++ b/app/workers/step_by_step_draft_update_worker.rb
@@ -23,18 +23,24 @@ class StepByStepDraftUpdateWorker
   end
 
   def update_assigned_to
-    unless step_by_step_page.has_draft?
-      generate_internal_change_note
+    unless assigned_to_current_user?
       step_by_step_page.assigned_to = @current_user
       step_by_step_page.save
+      generate_internal_change_note
     end
   end
 
   def generate_internal_change_note
     change_note = step_by_step_page.internal_change_notes.new(
       author: @current_user,
-      description: "Draft created by #{@current_user}",
+      description: "Draft saved by #{@current_user}",
     )
     change_note.save
+  end
+
+private
+
+  def assigned_to_current_user?
+    step_by_step_page.assigned_to == @current_user
   end
 end
--- a/spec/workers/step_by_step_draft_update_worker_spec.rb
+++ b/spec/workers/step_by_step_draft_update_worker_spec.rb
@@ -32,9 +32,16 @@ RSpec.describe StepByStepDraftUpdateWorker do
       before do
         step_by_step_page.mark_draft_updated
       end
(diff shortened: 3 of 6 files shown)


spec/workers/step_by_step_draft_update_worker_spec.rb
```diff
@@ -25,16 +25,16 @@
     context 'when a guide is a new draft' do
       it 'should assign the draft to the current user' do
         described_class.new.perform(step_by_step_page.id, @current_user.name)
-        expect(StepByStepPage.find(step_by_step_page.id).assigned_to).to eql "New author"
+        expect(StepByStepPage.find(step_by_step_page.id).assigned_to).to include "New author"
```
leenagupte: Shouldn't this still check equality?

spec/workers/step_by_step_draft_update_worker_spec.rb
```diff
@@ -25,16 +25,16 @@
...
       end
     end
     context 'when a guide is a draft that has already been updated' do
       before do
         step_by_step_page.mark_draft_updated
       end
-      it 'should remain assigned to the old author' do
+      it 'should be assigned to the new author' do
```
leenagupte: I think what you are trying to test here is that `assigned_to` field hasn't been updated.

Rails has a nifty way to check for changed attributes. Basically, `assigned_to_changed?` should be false.

There's a better write up here: https://jasoncharnes.com/changed-attributes-rails

spec/workers/step_by_step_draft_update_worker_spec.rb
```diff
@@ -44,16 +44,26 @@
...
-        expect(step_by_step_page.internal_change_notes.count).to eql 0
+        expect(step_by_step_page.internal_change_notes.count).to eql 1
+      end
+    end
+    context "when it is updated by the same user" do
+      before do
+        step_by_step_page.mark_draft_updated
+      end
+      it "should not generate another change note" do
+        described_class.new.perform(step_by_step_page.id, @current_user.name)
```
leenagupte: I think you use the factory to set the initial user: `create(:step_by_step_page, assigned_to: @current_user.name)`

app/workers/step_by_step_draft_update_worker.rb
```diff
@@ -23,17 +23,17 @@ def update_draft
   end
 
   def update_assigned_to
-    unless step_by_step_page.has_draft?
-      generate_internal_change_note
+    unless step_by_step_page.assigned_to == @current_user
```
leenagupte: Nitpick: You could move this check to a function e.g. `assigned_to_current_user?`

app/views/shared/steps/_heading.html.erb
```diff
@@ -2,6 +2,7 @@
   <small><%= action %><span class="sr-only"> - </span></small><br />
   <%= step_nav %>
 </h1>
-<br>
-<p>Last saved by <%= @step_by_step_page.assigned_to %></p>
-<br>
+
+<% if @step_by_step_page.assigned_to %>
```
leenagupte: I think you need to add a `present?` to this check.

leenagupte: The conditional behaves strangely when you test for nil values:

```
irb(main):006:0> b = ""
=> ""
irb(main):007:0> b.present?
=> false
irb(main):008:0> puts "hello" if b
hello
=> nil
irb(main):009:0> puts "hello" if b.present?
=> nil
```

PR #529, opened 2018-11-12
Add sort options to government documents finder

andrewgarner:
It is safe to remove the `default_order` attribute as this is also the default in finder-frontend.

Part of https://trello.com/c/B8oMuUhi


andrewgarner:
Closing this in favour of adding to the news-and-communications finder.

PR #536, opened 2018-12-07
Edit campaign route and description

vanitabarrett:
We have updated the URL and description following user research.

### DIFF
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -23,9 +23,9 @@ class SpecialRoutePublisher
       prefix: [
         {
           content_id: "ecb55f9d-0823-43bd-a116-dbfab2b76ef9",
-          base_path: "/prepare-uk-leaving-eu",
-          title: "Prepare for the UK leaving the EU",
-          description: "How to prepare for Brexit in March 2019 if you're a British citizen or have indefinite leave to remain in the UK.",
+          base_path: "/prepare-eu-exit-live-uk",
+          title: "Prepare for EU Exit if you live in the UK",
+          description: "",
         },
       ]
     }
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -2,6 +2,7 @@ require 'gds_api/publishing_api/special_route_publisher'
 
 class SpecialRoutePublisher
   def initialize(publisher_options)
+    @publishing_api = publisher_options[:publishing_api]
     @publisher = GdsApi::PublishingApi::SpecialRoutePublisher.new(publisher_options)
   end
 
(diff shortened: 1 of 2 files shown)


lib/special_route_publisher.rb
```diff
@@ -23,9 +23,9 @@ def self.routes
       prefix: [
         {
           content_id: "ecb55f9d-0823-43bd-a116-dbfab2b76ef9",
-          base_path: "/prepare-uk-leaving-eu",
-          title: "Prepare for the UK leaving the EU",
-          description: "How to prepare for Brexit in March 2019 if you're a British citizen or have indefinite leave to remain in the UK.",
+          base_path: "/prepare-eu-exit-live-UK",
```
sihugh: should possibly be a lower-case UK?

lib/special_route_publisher.rb
```diff
@@ -18,14 +18,27 @@ def publish(route_type, route)
     )
   end
 
+  def publish(route_type, route)
```
andrewgarner: Should this be `unpublish` @vanitabarrett ?

vanitabarrett: @andrewgarner Yep, it should be. Also this is WIP, I haven't had a chance to test this out yet, not sure if it works...

andrewgarner: Ahh okay, no worries.

`GdsApi::PublishingApi::SpecialRoutePublisher` doesn't have `unpublish` as part of it's API so I think you might need to use https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api_v2.rb#L131-L162

vanitabarrett:
@andrewgarner I've got the unpublish task working (tested on integration). Are you able to re-review?

lib/special_route_publisher.rb
```diff
@@ -2,6 +2,10 @@
 
 class SpecialRoutePublisher
   def initialize(publisher_options)
+    @publishing_api = GdsApi::PublishingApiV2.new(
```
andrewgarner: How about `@publishing_api = publisher_options[:publishing_api]`

lib/tasks/publishing_api.rake
```diff
@@ -48,4 +48,29 @@ namespace :publishing_api do
       end
     end
   end
+
+  desc "Unpublish special routes"
+  task unpublish_special_routes: :environment do
+    publishing_api = GdsApi::PublishingApiV2.new(
```
andrewgarner: How about `publishing_api = Services.publishing_api`?

https://github.com/alphagov/collections-publisher/blob/0ac7c03d072c068de543e684702cfc67dd7265f4/app/lib/services.rb#L6-L12

PR #540, opened 2018-12-12
Handle missing batch link reports

rubenarakelyan:
This commit adds an exception handler for 404 errors from link-checker-api. This can happen if the batch link report has been removed, which happens autmoatically on a schedule to prevent build-up of old reports. The app should handle this case.


thomasleese:
I wonder if it's worth making the app run another batch check in this case, but this seems like a fine solution for now.

PR #581, opened 2019-03-15
Get prepare-eu-exit page into site search

sihugh:
We want to make the `/prepare-eu-exit` page appear in GOV.UK site search under the services supergroup.  I'm doing this by tagging it as an `answer` document type in the short term - we can change this _soon_.  

The page is not appearing in search at all at the moment which is a problem.

I've set the description to one concocted by the Bravigation teams' content specialists.

Related to: https://github.com/alphagov/collections/pull/1056

https://trello.com/c/5GLtZUxs/152-get-prepare-eu-exit-in-search

### DIFF
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -9,7 +9,6 @@ class SpecialRoutePublisher
   def publish(route_type, route)
     @publisher.publish(
       route.merge(
-        format: "special_route",
         publishing_app: "collections-publisher",
         rendering_app: "collections",
         type: route_type,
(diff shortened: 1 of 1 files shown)


lib/special_route_publisher.rb
```diff
@@ -9,7 +9,6 @@ def initialize(publisher_options)
   def publish(route_type, route)
     @publisher.publish(
       route.merge(
-        format: "special_route",
```
bilbof: Could you add some context on why `special_route` is no longer needed here?

sihugh: Ah yeah - format isn't really a thing any more. It was sort of a mixture of doc type and schema back in the day.  

The special route publisher defaults the `document_type` and `schema` now unless you override them --> https://github.com/alphagov/gds-api-adapters/blob/674c2faab55e0ad6e6958c747ab1a66b564a4b59/lib/gds_api/publishing_api/special_route_publisher.rb#L18-L19.  

We're OK with the doc type being answer, because that'll get rummager on our side, but the schema of our content item should be special_route because that's more appropriate.

bilbof: Makes sense - thanks for explaining!

PR #595, opened 2019-04-02
Update citizen readiness title

sihugh:
SEO needs us to use Brexit.


sihugh:
"It depends".  Users tend to know about Brexit, I think, but there are instances where content people have moved towards the other variant - perhaps more in specialist content.

PR #614, opened 2019-05-15
Add primary publsihing organisation for Mainstream Browse Pages

MahmudH:
The primary publishing organisation is Government Digital Service for all Mainstream Browse Pages.
This is because GDS owns these pages.

### DIFF
--- a/app/presenters/mainstream_browse_page_presenter.rb
+++ b/app/presenters/mainstream_browse_page_presenter.rb
@@ -1,6 +1,8 @@
 class MainstreamBrowsePagePresenter < TagPresenter
 private # rubocop:disable Layout/IndentationWidth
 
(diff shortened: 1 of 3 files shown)


app/presenters/mainstream_browse_page_presenter.rb
```diff
@@ -11,6 +11,7 @@ def links
       "active_top_level_browse_page" => active_top_level_browse_page_id,
       "top_level_browse_pages" => @tag.class.sorted_parents.map(&:content_id),
       "second_level_browse_pages" => second_level_browse_pages,
+      "primary_publishing_organisation" => ["af07d5a5-df63-4ddc-9383-6a666845ebe9"],
```
cbaines: It would be good to pull the content id value here out in to a constant, at least so it has a human readable name.

PR #618, opened 2019-05-20
Assign secondary content to step by step pages

leenagupte:
Trello: https://trello.com/c/iE60uYXs

Related to:
* https://github.com/alphagov/govuk_publishing_components/pull/900
* https://github.com/alphagov/publishing-api/pull/1501
* https://github.com/alphagov/govuk-content-schemas/pull/896

## Motivation

Secondary content is content that is about the service, but not needed by enough users to be included in the step by step. An example of this is internal links in content that is part of the step by step.

At the moment, if users click on internal links, they lose sight of the step by step journey they are on.

This PR allows content designers to assign content on GOV.UK as secondary content on a step by step.

## Expected changes

A new tab in collections-publisher to assign secondary content to a step by step page in its links.

Secondary content behaves in the same way as all other attributes of a step by step. When secondary content is added, it is either added to the existing draft, or a new draft is created. Secondary content does not appear on live until the draft is published. If the draft is discarded, the secondary content links will be reverted to the previous version's.

Screen Shot 2019-06-06 at 12 48 44
Screen Shot 2019-06-06 at 12 49 03

There is a [PR in review](https://github.com/alphagov/govuk_publishing_components/pull/900) to display the step by step breadcrumb and side bar on secondary content.

### DIFF
--- a/app/controllers/secondary_content_links_controller.rb
+++ b/app/controllers/secondary_content_links_controller.rb
@@ -34,10 +34,11 @@ private
   end
 
   def content_item
-    content_id = Services.publishing_api.lookup_content_id(base_path: params[:base_path], with_drafts: true)
+    base_path = URI.parse(params[:base_path]).path
+    content_id = Services.publishing_api.lookup_content_id(base_path: base_path, with_drafts: true)
 
     if content_id.nil?
-      @error = "#{params[:base_path]} doesn't exist on GOV.UK."
+      @error = "#{base_path} doesn't exist on GOV.UK."
       return {}
     end
 
--- a/app/controllers/secondary_content_links_controller.rb
+++ b/app/controllers/secondary_content_links_controller.rb
@@ -1,6 +1,18 @@
 class SecondaryContentLinksController < ApplicationController
   before_action :require_gds_editor_permissions!
 
+  def create
+    @secondary_content_link = step_by_step_page.secondary_content_links.new(content_item)
+
+    if @secondary_content_link.save
+      update_downstream
+
+      redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id), notice: 'Secondary content was successfully linked.'
+    else
+      redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id)
+    end
+  end
+
   def index
     @secondary_content_link = step_by_step_page.secondary_content_links.new
     @secondary_content_links = step_by_step_page.secondary_content_links.all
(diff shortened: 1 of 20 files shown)


app/controllers/secondary_content_links_controller.rb
```diff
@@ -33,21 +33,43 @@ def destroy
...
+
   def secondary_content_link
     @secondary_content_link ||= step_by_step_page.secondary_content_links.find(params[:id])
   end
 
   def update_downstream
     StepByStepDraftUpdateWorker.perform_async(step_by_step_page.id, current_user.name)
   end
+
+  def step_params
```
leenagupte: @DilwoarH Where is this used?

app/controllers/secondary_content_links_controller.rb
```diff
@@ -33,21 +33,43 @@ def destroy
...
   def step_by_step_page
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
   end
 
+  def secondary_content_links
+    @secondary_content_links = StepByStepPage.find(params[:step_by_step_page_id]).secondary_content_links
+  end
+
+  def new_secondary_content_link
+    step_by_step_page.secondary_content_links.new
```
leenagupte: @DilwoarH do we need this? It's doing the same thing as the `new` action.

leenagupte: I think `new` could be changed to:

```
  def new
    @secondary_content_link = step_by_step_page.secondary_content_links.new
  end
```

And the form created around `@secondary_content_link`?

app/controllers/secondary_content_links_controller.rb
leenagupte: Once we have the `base_path` we can get the `title` and `content_id` from publishing-api, so we might want to move all of this to the model.

app/controllers/secondary_content_links_controller.rb
DilwoarH: The issue is (correct me if I'm wrong) that the new would assume that there would be another view called `new.html.erb`. In our current approach, everything happens in one view, therefore, this wouldn't be needed.

app/controllers/secondary_content_links_controller.rb
DilwoarH: Yeah - this was just done as a placeholder, your function expected these params to be passed in, so we had to pass it in to make everything work.

DilwoarH: @leenagupte - this is used here https://github.com/alphagov/collections-publisher/pull/618/files#diff-49365286d564b9a2a3a2a422330233deR9

app/controllers/secondary_content_links_controller.rb
leenagupte: Not sure tbh. One to try out.

spec/models/secondary_content_link_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+          :secondary_content_link,
+          step_by_step_page: step_by_step_page,
+          base_path: "/same-step-by-step/shiny-new-path"
+        )
+        expect(secondary_content_link).to be_valid
+        expect(secondary_content_link.save).to eql true
+        expect(secondary_content_link).to be_persisted
+      end
+
+      it 'returns and error when the base_path and step_by_step_page are not unique' do
```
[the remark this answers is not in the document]
leenagupte: @brucebolt Fixed and fixed up.

PR #622, opened 2019-05-21
Add primary publishing organisation to root browse page

MahmudH:
Root browse page is a special case as it is separate from the rest of mainstream browse pages, therefore needs a separate way of adding an organisation

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -100,6 +100,13 @@ namespace :publishing_api do
 
       puts "Patching links for #{page.content_id}..."
     end
+
+    root_page = RootBrowsePagePresenter.new('state' => 'published')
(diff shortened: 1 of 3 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -100,6 +100,13 @@ namespace :publishing_api do
 
       puts "Patching links for #{page.content_id}..."
     end
+
+    root_page = RootBrowsePagePresenter.new('state' => 'published')
+    Services.publishing_api.patch_links(
+      root_page.content_id,
+      root_page.render_links_for_publishing_api
+    )
+    puts "Patching links for root page..."
```
cbaines: The patching has already happened by this point in the code, so maybe put this in the past tense.

PR #629, opened 2019-05-30
Cope with trailing whitespace in step markdown

sihugh:
At present, steps with multiple trailing lines will create an empty paragraph once parsed.

This breaks Google's [rich snippet parser](https://search.google.com/test/rich-results?id=sm1WcP0wSd_gSZ-3ZSSVUQ) for HowTo schemas, so we should fix it.

Screenshot 2019-05-30 15 52 56

Note that it doesn't have any visible effects on the page as we seem to defend against that in the step by step component.

https://collections-publisher.publishing.service.gov.uk/step-by-step-pages/66/internal-change-notes

### DIFF
--- a/spec/services/step_content_parser_spec.rb
+++ b/spec/services/step_content_parser_spec.rb
@@ -47,6 +47,7 @@ RSpec.describe StepContentParser do
 
           Conducted by Mr Andrew Preview
 
+
         HEREDOC
 
         expect(subject.parse(step_text)).to eq([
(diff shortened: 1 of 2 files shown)


spec/services/step_content_parser_spec.rb
```diff
@@ -47,6 +47,7 @@
 
           Conducted by Mr Andrew Preview
 
+
```
leenagupte: Nitpick police: Why was this line added?

sihugh: That's because the single blank line didn't reveal the issue :-(

PR #640, opened 2019-07-10
Fix ability to preview step by step on first attempt

ChrisBAshton:
Sets `draft_updated_at` before sending payload to publishing-api.

This fixes a bug whereby live step-by-steps can't be previewed
if they have only been edited once since publish. If someone
added/removed/edited a step in a live step-by-step and then
clicked 'Preview', they would see the "Sorry, we're experiencing
technical difficulties page" normally only seen by users missing
an access token in the request URL. However, if they then made an
additional edit, they could preview the step-by-step as expected.

The bug arises from the fact that StepNavPresenter only sets the
`access_limited` field in the publishing-api payload if the
StepByStepPage `has_draft?` method returns `true`:

https://github.com/alphagov/collections-publisher/blob/909a35f7a688f12447579186d8f0eee48cd60afe/app/presenters/step_nav_presenter.rb#L37

But if the
StepByStepPage is already live, then `has_draft?` returns `false`
at this point, as we haven't made the edit yet!
On subsequent edits, the `has_draft?` method returns `true` because
the StepByStepPage is in draft/'Unpublished changes' state from
the previous edit, and therefore the `access_limited` field is now
set and the user can now preview their step-by-step.

By calling `mark_draft_updated` _before_ initialising the StepNavPresenter
and getting the payload to pass to publishing-api, we ensure that
the step-by-step is in draft mode by the time we call `has_draft?`
and we can therefore preview our change right away.

See https://trello.com/c/g1M754qO/

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -21,7 +21,7 @@ FactoryBot.define do
 
   factory :published_step_by_step_page, parent: :step_by_step_page_with_steps do
     draft_updated_at { 3.hours.ago }
-    published_at { Time.zone.now }
+    published_at { 3.hours.ago }
   end
 
   factory :step_by_step_page_with_navigation_rules, parent: :step_by_step_page_with_steps do
--- a/app/services/step_nav_publisher.rb
+++ b/app/services/step_nav_publisher.rb
@@ -1,9 +1,9 @@
 class StepNavPublisher
   def self.update(step_by_step_page, publish_intent = PublishIntent.minor_update)
+    step_by_step_page.mark_draft_updated
     presenter = StepNavPresenter.new(step_by_step_page)
     payload = presenter.render_for_publishing_api(publish_intent)
     Services.publishing_api.put_content(step_by_step_page.content_id, payload)
-    step_by_step_page.mark_draft_updated
   end
 
   def self.discard_draft(step_by_step_page)
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -112,7 +112,7 @@ RSpec.feature "Managing step by step pages" do
   end
 
   scenario "User reverts a step by step page" do
-    given_there_is_a_published_step_by_step_page
+    given_there_is_a_published_step_by_step_page_with_unpublished_changes
     when_I_view_the_step_by_step_page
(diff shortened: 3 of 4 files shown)


ChrisBAshton:
@leenagupte the 'saved a new draft' bit happens in `step_by_step_page.mark_draft_updated`:

```ruby
  def mark_draft_updated
    update_attribute(:draft_updated_at, Time.zone.now)
  end
```

If we construct the payload for publishing API before we've called `step_by_step_page.mark_draft_updated`, then the `fields[:access_limited]` doesn't get set, as `has_draft?` returns false:

```ruby
# StepNavPresenter
  def optional_fields
    fields = {}
    fields[:access_limited] = access_limited_tokens if step_nav.has_draft?
    fields
  end

# StepByStepPage
  def has_draft?
    draft_updated_at.present? && draft_updated_at != published_at
  end
  # `draft_updated_at` is `nil` at this point ^^
```

spec/factories.rb
```diff
@@ -24,6 +24,11 @@
     published_at { Time.zone.now }
   end
 
+  factory :published_step_by_step_page_live, parent: :step_by_step_page_with_steps do
```
leenagupte: Can `published_step_by_step_page` reused instead of creating a new factory?

ChrisBAshton: No, as the draft date is separate to the published date, so `step_nav.status` is actually `unpublished_changes` instead of `live`.

(I think it's badly named and should be renamed to `published_step_by_step_with_changes`, and my own factory should be `published_step_by_step`!)

ChrisBAshton:
Have re-written the commit 👍

spec/factories.rb
leenagupte: Yes, I think `published_step_by_step_page` should be changed to return a step by step that has a status of "live".

If there are any tests that fail because they rely on the status being "unpublished changes", then a new factory could be added. 

If you need to create a new factory I'd recommend doing these changes in a new commit. Otherwise, if you are only changing the `draft_updated_at` time in `published_step_by_step_page` then a line in the commit note explaining why should be enough.

ChrisBAshton: That's now done 👍 one of the tests did fail, so I've encapsulated it in a commit before the main commit of the PR. Putting it as the second commit would've required making a change to something I did in the first commit.

PR #646, opened 2019-07-15
Move duplicated 'update_downstream' method into helper

ChrisBAshton:
The definition of 'update_downstream' was repeated in three places.
Ideally we shouldn't have to update this in more than one place (DRY).

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -1,5 +1,6 @@
 class StepByStepPagesController < ApplicationController
   include PublishingApiHelper
+  include StepByStepHelper
 
   before_action :require_gds_editor_permissions!
   before_action :set_step_by_step_page, only: %i[show edit update destroy]
@@ -115,10 +116,6 @@ private
     StepNavPublisher.discard_draft(@step_by_step_page)
   rescue GdsApi::HTTPNotFound
     Rails.logger.info "Discarding #{@step_by_step_page.content_id} failed"
-  end
-
-  def update_downstream
-    StepByStepDraftUpdateWorker.perform_async(@step_by_step_page.id, current_user.name)
   end
 
   def publish_page(publish_intent)
--- a/app/helpers/step_by_step_helper.rb
+++ b/app/helpers/step_by_step_helper.rb
@@ -1,0 +1,5 @@
+module StepByStepHelper
+  def update_downstream
+    StepByStepDraftUpdateWorker.perform_async(@step_by_step_page.id, current_user.name)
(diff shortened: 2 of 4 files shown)


app/controllers/step_by_step_pages_controller.rb
```diff
@@ -1,8 +1,10 @@
 class StepByStepPagesController < ApplicationController
   include PublishingApiHelper
+  include StepByStepHelper
 
-  before_action :require_gds_editor_permissions!
   before_action :set_step_by_step_page, only: %i[show edit update destroy]
+  before_action :require_gds_editor_permissions!
```
leenagupte: This change can be discarded as it's only a line move.

app/helpers/step_by_step_helper.rb
```diff
@@ -0,0 +1,5 @@
+module StepByStepHelper
```
leenagupte: Can you add a test for this helper?

ChrisBAshton: There's no new functionality being added here, just a refactor - all current tests continue to pass. If I change `@step_by_step_page.id` to 'foo' (or anything else) the tests fail. That gives me confidence this works as expected.

I think adding a test would just end up shifting complexity into `/spec` instead and this wouldn't be much of a refactor at all.

leenagupte: @ChrisBAshton I think that the values for step_by_step_page.id and current_user.name should be passed in. This will allow the helper method to be tested and give it more flexibility.

The benefit of adding a test is that you don't just have confidence that it works today, but that it continues to work in the future.

Calling `@step_by_step_page` directly in the helper method works at the moment, because so far every controller that uses the helper method has `@step_by_step_page` defined, but if someone creates a new controller, and uses another variable name e.g. `@step_nav` this helper method won't work for them.

Example test:

```
    it "adds a job to the draft worker" do
      allow(Services.publishing_api).to receive(:lookup_content_id)
      user = create(:user, name: "New author")
      step_by_step_page = create(:step_by_step_page)

      Sidekiq::Testing.fake! do
        expect {
          helper.update_downstream(step_by_step_page.id, user.name)
        }.to change(StepByStepDraftUpdateWorker.jobs, :size).by(1)
      end
    end
```

Calls to `update_downstream` would have to be amended for this to work: 
```
update_downstream(step_by_step_page.id, current_user.name)
```

ChrisBAshton:
I think I'll quietly close this PR - it was intended to be a quick tidy-up! We have enough collections-publisher work coming up and perhaps this refactor will happen organically as part of that.

PR #651, opened 2019-07-16
Add scheduling permission

leenagupte:
Trello: https://trello.com/c/i5BbBKPg

A new "Scheduling" permission has been added to collections-publisher.
This will allow us to "soft" deploy the features being added to implement scheduling and
hide them behind this special permission.
When scheduling is ready to go "live" this permission will be removed.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -39,4 +39,8 @@ private
       GdsApi::GovukHeaders.set_header(:x_govuk_authenticated_user, current_user.uid)
     end
   end
+
+  def require_scheduling_permissions!
+    authorise_user!("Scheduling")
+  end
 end


leenagupte:
> Makes sense. Two questions:
> 
> 1. How did you do `A new "Scheduling" permission has been added to collections-publisher.`? Is there a PR for that somewhere or was it done through the signon UI...?
> 2. Just to check, we would simply wrap `current_user.has_permission? "Scheduling"` around anything which renders the option to schedule?

1. There is a UI that allows you to add new permissions, but you need to be a Superadmin.
2. We can just add `require_scheduling_permissions!` to the controller and specify which actions it applies to, e.g. https://github.com/alphagov/collections-publisher/blob/master/app/controllers/topics_controller.rb#L3

PR #652, opened 2019-07-18
Schedule step by step publication

ChrisBAshton:
Adds a basic scheduling screen. The input form for that screen will be improved in a future PR, once we have access to govuk_publishing_components.

- Creates new /schedule endpoint (GET and POST)
- Ensures that a publish can only be scheduled for a future date, not one in the past
- only allow scheduling of step by steps that are in draft form.
- user can see the scheduled date/time.

Screen Shot 2019-07-26 at 07 47 13

And once one is scheduled:

Screen Shot 2019-07-26 at 07 46 58

Trello card: https://trello.com/c/4qnCWOOe/30-allow-publishers-to-schedule-a-step-by-step-for-publication


app/models/step_by_step_page.rb
```diff
@@ -56,6 +57,10 @@ def status
     draft_status
   end
 
+  def scheduled_for_publishing?
+    has_draft? && scheduled_at.present? && scheduled_at.future?
+  end
```
ChrisBAshton: This duplicates what Leena added earlier. Will remove.

spec/models/step_by_step_page_spec.rb
```diff
@@ -226,6 +249,18 @@
...
+      # TODO
+    end
+
+    it 'should not allow scheduling a date in the past' do
+      # TODO
+    end
+
+    it 'should only allow scheduling of draft or scheduled step by steps' do
+      # TODO
+    end
```
ChrisBAshton: Possibly duplicates some of Leena's tests.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -127,6 +141,11 @@ def publish_page(publish_intent)
     @step_by_step_page.mark_as_published
   end
 
+  def schedule_to_publish(schedule_intent)
+    StepNavPublisher.update(@step_by_step_page, schedule_intent)
```
ChrisBAshton: Is this line needed?

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -9,6 +9,7 @@
     before do
       given_I_am_a_GDS_editor
       setup_publishing_api
+      stub_request(:put, %r{/publish-intent/}) # stub any PUT calls made by the scheduled publish worker
```
ChrisBAshton: Change comment to say `StepByStepScheduledPublishWorker` so it's copy & pasteable and searchable

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: Yes, it's needed - one of the tests fails otherwise (`status` isn't correctly reported).

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -0,0 +1,82 @@
...
+  action: 'Set a date and time to publish'
+%>
+
+<%= render 'nav', step_by_step_page: @step_by_step_page, active: "schedule" %>
+
+<% if @schedule_intent %>
+  <%= render 'shared/steps/form_errors', resource: @schedule_intent %>
+<% end %>
+
+<% if @step_by_step_page.status[:name] == 'live' %>
```
leenagupte: f the status is live, couldn't the whole page be hidden instead?

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -0,0 +1,82 @@
...
+  <div class="alert alert-info">
+    Scheduled to be published at <%= @step_by_step_page.scheduled_at %>
+  </div>
+<% else %>
+  <div class="row">
+    <div class="col-md-7">
+      <%= form_tag do %>
+        <div class="form-group">
+          <%= label_tag :scheduled_at, 'Schedule datetime (e.g. "2020-04-20 10:26:51 UTC")' %>
+          <%= text_field_tag :scheduled_at, params[:scheduled_at]  %>
```
leenagupte: `govuk_publishing_components` has been added as a dependency, so rather creating a text field with all of the accompanying validation, you could use the components instead.

Something else to be aware of is that this text field is going to be replaced with two components. The values captured by the two components will need to be stitched together in the backend before being saved into `scheduled_at`.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -0,0 +1,82 @@
...
+<%= render 'shared/steps/step_breadcrumb', links: links %>
+
+<%= render 'shared/steps/heading',
+  step_nav: @step_by_step_page.title,
+  action: 'Set a date and time to publish'
+%>
+
+<%= render 'nav', step_by_step_page: @step_by_step_page, active: "schedule" %>
+
+<% if @schedule_intent %>
```
leenagupte: Can you move the addition `@schedule_intent` to the commit where it's created? Generally, it's preferable all of the tests pass in each commit.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -63,13 +63,27 @@ def publish
       @publish_intent = PublishIntent.new(params)
       if @publish_intent.valid?
         publish_page(@publish_intent)
-        generate_internal_change_note
+        generate_internal_change_note(@publish_intent)
```
leenagupte: Why has this changed?

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -165,16 +166,11 @@ def revert_page
   def generate_internal_change_note(intent)
     change_note = @step_by_step_page.internal_change_notes.new(
       author: current_user.name,
-      description: note_description(intent)
+      description: intent.note_description(current_user)
```
leenagupte: You shouldn't need to pass in `current_user`. It's a special variable added in one of the gems that accessible from the controllers.

app/models/schedule_intent.rb
```diff
@@ -0,0 +1,12 @@
+class ScheduleIntent < PublishIntent
```
leenagupte: What is `ScheduleIntent` used for? Why does it inherit from `PublishIntent`?

Can you separate out creating the `ScheduleIntent` class and it's associated commits into a separate commit and explain what it's for?

app/controllers/step_by_step_pages_controller.rb
leenagupte: I don't think you need this line. When the scheduled date is saved, you don't need to re-save the draft you just need to tell publishing-api that a step-by-step is due to be published on a certain date, which is what `StepNavPublisher.schedule_for_publishing(@step_by_step_page)` does.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -0,0 +1,82 @@
...
+  <div class="row">
+    <div class="col-md-7">
+      <%= form_tag do %>
+        <div class="form-group">
+          <%= label_tag :scheduled_at, 'Schedule datetime (e.g. "2020-04-20 10:26:51 UTC")' %>
+          <%= text_field_tag :scheduled_at, params[:scheduled_at]  %>
+        </div>
+
+        <div class="form-group">
+          <%= label_tag :change_note, "Change note" %>
```
leenagupte: I don't think this is needed. For now we can just assume that all scheduled publish requests will be a "minor" change, so we don't need to record a change note.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -0,0 +1,82 @@
...
+          <%= text_field_tag :change_note, params[:change_note]  %>
+        </div>
+
+        <div class="form-group">
+          <%= label_tag :update_type, "Update type" %>
```
leenagupte: Same here. If we assume that the scheduled publication is a minor change, we can remove this check.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -9,6 +9,7 @@
     before do
       given_I_am_a_GDS_editor
       setup_publishing_api
+      stub_request(:put, %r{/publish-intent/}) # stub any PUT calls made by StepByStepScheduledPublishWorker
```
leenagupte: There is a helper method that already does this: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api.rb#L37

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,16 +21,97 @@
       then_I_visit_the_navigation_steps_page_again
       and_I_see_my_selected_preferences
     end
+
```
leenagupte: It's good that you've added feature tests 👍 but can you move them to where the controller actions were added.

It makes things easier to follow if the code change and the associated tests are in the same place.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,16 +21,97 @@
...
+    scenario "User schedules publishing" do
+      given_there_is_a_draft_step_by_step_page
+      and_I_visit_the_scheduling_page
+      and_I_fill_in_the_scheduling_form
+      when_I_submit_the_form
+      then_I_should_see "has been scheduled to publish"
+      and_my_step_by_step_should_have_the_status "Scheduled"
```
leenagupte: Nitpick: Use "the" instead of "my"

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,16 +21,97 @@
...
+    scenario "User tries to schedule publishing for date in the past" do
+      given_there_is_a_draft_step_by_step_page
+      and_I_visit_the_scheduling_page
+      and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
+      when_I_submit_the_form
+      then_I_should_see "Scheduled at can't be in the past"
+      and_my_step_by_step_should_have_the_status "Draft"
+    end
+
+    scenario "User tries to schedule publishing for an already live step by step" do
```
leenagupte: I don't think you need this test as the user shouldn't be able to see the "scheduling" page if the step by step has a status of "published" and that can be confirmed with a controller test.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,16 +21,97 @@
...
   def given_there_is_a_step_by_step_page_with_navigation_rules
     @step_by_step_page = create(:step_by_step_page_with_navigation_rules)
   end
 
+  def given_there_is_a_draft_step_by_step_page
+    @step_by_step_page = create(:step_by_step_page_with_steps)
+    expect(@step_by_step_page.status[:name]).to eq 'draft'
+  end
+
+  def given_there_is_a_live_step_by_step_page
```
leenagupte: This could be replaced with the existing `given_there_is_a_published_step_by_step_page`

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Exactly - I'm keen that work gets done in a separate PR, as it will also require tests to ensure we're stitching together the various input values correctly.

That piece of work can be done in https://trello.com/c/oFvkJGff/44-use-govuk-components-in-the-scheduling-interface, and the design for it is still being decided. This is very much a placeholder until that happens.

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Fair enough 👍 wasn't sure whether to put all the UI stuff in one commit, but your suggestion makes more sense.

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
```diff
@@ -6,25 +6,10 @@
...
-
-    step_by_step_page.reload
-    expect(step_by_step_page.has_been_published?).to be true
-  end
-
-  it "generates a change note" do
-    described_class.new.perform(step_by_step_page.id)
-    expect(step_by_step_page.internal_change_notes.first.description).to eq("Published on schedule")
-  end
+  let(:future_datetime) { Time.now + 1.hour }
```
leenagupte: Can you move this change to a separate commit? Preferably when the "InFuture" validator was added? Or the commit directly after it.

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
```diff
@@ -34,12 +19,22 @@
...
-    step_by_step_page = create(:draft_step_by_step_page, scheduled_at: Time.now + 1.hour)
     described_class.new.perform(step_by_step_page.id)
 
     expect(StepNavPublisher).to_not have_received(:publish)
 
     step_by_step_page.reload
     expect(step_by_step_page.has_been_published?).to be false
   end
+
+  it "publishes the step by step (and generates a change note) when it's time to publish" do
```
leenagupte: For readability, can you leave them as separate tests?

Also, why have they moved location? It would be easier to see what's changed if only the affected lines are updated.

app/controllers/application_controller.rb
```diff
@@ -11,12 +11,16 @@ class ApplicationController < ActionController::Base
 
 private
 
-  helper_method :gds_editor?, :active_navigation_item
+  helper_method :gds_editor?, :user_can_schedule?, :active_navigation_item
```
leenagupte: It feels like this commit is doing two things. Restricting access to the scheduling function, and refactoring how change notes are generated. Can you split them, so that each commit does one thing.

ChrisBAshton: Whoops - they were originally two commits.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -141,4 +162,8 @@ def and_I_see_my_selected_preferences
     expect(page.all(:select)[0].value).to eq("never")
     expect(page.all(:select)[1].value).to eq("always")
   end
+
+  def and_I_should_see_a_permission_exception
```
leenagupte: The scheduling action should be hidden, rather than tell users that they don't have access to something. Also, making "scheduling" live will involve removing the scheduling permission, and moving access to scheduling to the normal "GDS Editor" permission. Think of the scheduling permission as a feature flag that we're going to turn off.

app/views/step_by_step_pages/schedule.html.erb
AgaDufrat: We talked about this in the dev planning session and "Change note is created when a step by step is scheduled" is one of the acceptance criteria on the card but happy to remove it.

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Actually I may just merge this commit and the next one, because you also want the `routes.rb` change to be in the same commit as the controller tests. That would just leave a half-finished `schedule.html.erb file` in this commit, which isn't very useful.

app/views/step_by_step_pages/schedule.html.erb
leenagupte: @AgaDufrat I meant an internal change note, not a public one.

leenagupte: Sorry, I assumed the internal change would be auto-generated in the same way that the "Draft saved by" one is.

app/views/step_by_step_pages/schedule.html.erb
leenagupte: I'd be wary of merging commits.

The new route should be added with the controller, but A rather than adding a full view, you can just add a placeholder one with `<p>To Do</p>`

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
ChrisBAshton: Done 👍

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
ChrisBAshton: Done 👍

app/views/step_by_step_pages/schedule.html.erb
AgaDufrat: I see. I can remove the form and we can just automatically generate a change note (minor update by default) saying "Changes scheduled by user_name for publishing at date_time".

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: We've just tried this but it made it difficult to test. General [advice seems to be](https://stackoverflow.com/questions/22076699/rspec-model-testing-current-user-in-model) to pass `current_user` from the controller into your model; the model shouldn't reference the controller.

leenagupte: Ah, sorry. I think when I looked at this, it looked like `note_description` was still in the controller.

spec/features/step_by_step_navigation_spec.rb
ChrisBAshton: Even so, we need to decide what should happen if an unauthorised user manages to get to the `/schedule` page, and have a test for that. Doesn't matter to me if it's an error message or a 403 response code or a redirect to the main step by step page. Any thoughts?

spec/features/step_by_step_navigation_spec.rb
ChrisBAshton: Done 👍 had to add a refactor commit to consolidate all these methods into the step_nav_steps.rb.

spec/features/step_by_step_navigation_spec.rb
leenagupte: Adding `require_scheduling_permissions!` to the controller will restrict access to users who have the "Scheduling" permissions. Other users who tried to hack the URL to access the schedule page will be kicked out. There is more information on how this works in the [gds-sso gem](https://github.com/alphagov/gds-sso/blob/b09c0510d417dce9f3a3fcad10630e29fd9c7a03/README.md#securing-your-application)

spec/features/step_by_step_navigation_spec.rb
ChrisBAshton: Thanks - done!

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: What do you mean by hidden? (We can hide the 'Schedule' button from the publish-or-delete.html, but what happens if a user with Scheduling permissions gets to this page anyway?)

leenagupte: We hide the publish button if a draft doesn't exist:  https://github.com/alphagov/collections-publisher/blob/master/app/views/step_by_step_pages/publish_or_delete.html.erb#L28

We could nest the schedule button inside the same check.

leenagupte: It's unlikely that a content designer here is going to try and hack the url and cause mischief, so I'm not worried about adding extra checks. It would be different if we were opening step by step publishing to the wider publishing community.

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: Have now removed this line, but needed to call `step_by_step_page.mark_draft_updated` within the `schedule_for_publishing` method.

leenagupte: Why?

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -119,4 +138,10 @@ def stub_publishing_api
     stub_any_publishing_api_put_content
     stub_any_publishing_api_publish
   end
+
+  def create_step_by_step_page
+    build(:step_by_step_page_with_steps).tap do |step_page|
+      step_page.save(validate: false)
+    end
+  end
```
ChrisBAshton: Can we avoid creating a new method and instead do something like this on line 34?

```ruby
let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
```

spec/features/step_by_step_navigation_spec.rb
AgaDufrat: Thank you for pointing that out. We were basically testing gds-sso gem. We are now testing if the ‘Schedule’ button doesn’t exist if you don’t have permission as well as 200 and 403 response in the controller.

spec/controllers/step_by_step_pages_controller_spec.rb
leenagupte: You should be able to re-use the `step_by_step_page` created on line 7.

AgaDufrat: yes, I don't need that and I don't need line 35 either!

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: Ah, turns out `given_there_is_a_draft_step_by_step_page` didn't actually create a step by step which set the `draft_updated_at` field. Have fixed the test, which means no need to call `mark_draft_updated` anymore.

app/models/schedule_intent.rb
AgaDufrat: ScheduleIntent no longer inherits from PublishIntent, since we are not giving the option to specify a change note.

The controller was getting big and it’s an attempt to make it simpler. This class follows the pattern of the PublishIntent model. As Chris said ‘it’s a self-contained model for the schedule-specific changenote and validation logic, and we also need a hook for displaying validation errors in the frontend.’
Having it a separate class prevents from complicating PublishIntent and I’m not sure how we could validate that the change_note is present for immediate publish but not for scheduled there. Plus we don’t have to make changes to StepNavPresenter. 
We will make it a separate commit.

ChrisBAshton: Have now pulled it out into a separate commit 👍

leenagupte: I'm still not sure what it's being used for as there is also validation logic in the `StepByStepPages` model.

It feels like things are getting over-complicated.
I thought that it would just be possible to:
1. Add a new button on publish-and-delete to direct users to the schedule screen
2. Add a new route, controller action and screen to capture the `scheduled_at` date
3. validate that the date the user supplied is in the future
4. Save `scheduled_at`
5. Call `StepNavPublisher.schedule_for_publishing` to schedule the job.

I'm not clear where `ScheduleIntent` fits into that.

ChrisBAshton: What would you suggest doing as an alternative? Aga & I have talked it through and we think this is nicely encapsulated - the alternative would be complicating the controller. If anything I'd suggest we take the duck typing even further and encapsulate the bulk of the `publish`/`schedule` controller methods inside the intents, e.g.

step_by_step_pages_controller.rb:

```ruby
def publish
  @publish_intent.action!
end

def schedule
  @schedule_intent.action!
end

```

publish_intent.rb:

```ruby
def action!
  if request.post?
    @publish_intent = PublishIntent.new(params)
    if @publish_intent.valid?
      publish_page
      generate_internal_change_note
      set_change_note_version
      redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been published."
    end
  end
end
```

schedule_intent.rb:

```ruby
def action!
  if request.post?
    @schedule_intent = ScheduleIntent.new(params)
    if @schedule_intent.valid?
      @step_by_step_page.update_attribute(:scheduled_at, params[:scheduled_at])
      schedule_to_publish
      generate_internal_change_note
      set_change_note_version
      redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled to publish."
    end
  end
end
```

leenagupte: The default update type is "minor", so you don't need to specifically set that anywhere, which is what I think you are trying to do with `ScheduleIntent`? So the code can be kept quite simple.

in `step_by_step_pages_controller.rb`:

```
  def schedule
    if request.post?
        if @step_by_step_page.update(:scheduled_at, params[:scheduled_at])
          StepNavPublisher.schedule_for_publishing(@step_by_step_page)
          generate_internal_change_note
          set_change_note_version
          redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled."
      end
    end
  end
```

`@step_by_step_page.update(:scheduled_at, params[:scheduled_at])` will run the validations in the model, including `in_future`. Any validation errors should be stored in `@step_by_step_page.errors`

So to show the error in the view, you could have:

```
<% if @step_by_step_page.errors.any? %>
  <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
<% end %>
```

leenagupte: For neatness, you could encapsulate `StepNavPublisher.schedule_for_publishing(@step_by_step_page)` in a method.

`PublishIntent` is a model, so it doesn't feel right to add controller actions in there.

ChrisBAshton: Have now removed ScheduleIntent 👍

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
```diff
@@ -6,25 +6,28 @@
...
-    described_class.new.perform(step_by_step_page.id)
-    expect(step_by_step_page.internal_change_notes.first.description).to eq("Published on schedule")
+    Timecop.travel future_datetime do
+      described_class.new.perform(step_by_step_page.id)
+      expect(step_by_step_page.internal_change_notes.first.description).to eq("Published on schedule")
+    end
   end
 
   it "doesn't publish the step by step if it isn't scheduled" do
-    step_by_step_page = create(:draft_step_by_step_page)
```
leenagupte: I think this line needs to be kept in, otherwise the test doesn't make sense as there is a scheduled_at date, but it's in the future, which is what the next test is testing.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,12 +23,71 @@
...
+  def and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
+    fill_in 'scheduled_at', with: '1937-04-20 10:26:51 UTC'
+  end
+
+  def when_I_submit_the_form
+    click_on 'Schedule to publish'
+  end
+
+  def and_there_should_be_no_schedule_button
+    expect(page).not_to have_content 'Schedule to publish'
```
leenagupte: Nitpick: You can check that the button itself doesn't exist. 
e.g. expect(page).not_to have_css("button", text: "Schedule to publish")

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -20,12 +23,71 @@
       then_I_visit_the_navigation_steps_page_again
       and_I_see_my_selected_preferences
     end
+
+    scenario "User schedules publishing" do
```
leenagupte: These tests are really great, but given they are about publishing rather than setting navigation, perhaps they should live in `managing_step_by_step_pages_spec.rb` with the rest of the publishing tests?

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -31,8 +31,10 @@
           <p>Before publishing check for broken links.</p>
           <p>After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
           <%= link_to 'Publish changes', step_by_step_page_publish_path(@step_by_step_page), class: "btn btn-primary publish-button" %>
-          <p>Or</p>	
-          <%= link_to 'Schedule publish', step_by_step_page_schedule_path(@step_by_step_page), class: "btn btn-info" %>	
+          <p>Or</p>
```
leenagupte: This "Or" needs to be wrapped in the guard clause as well, otherwise users without the scheduling permission are shown:

Screen Shot 2019-07-24 at 17 51 26

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -30,6 +30,22 @@
     end
   end
 
+  describe "GET Step schedule page" do
```
leenagupte: Typo: This should be "Step by step schedule page"

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -30,6 +30,22 @@
...
+    it "can only be accessed by users with Scheduling permissions" do
+      stub_user.permissions = ["signin", "GDS Editor", "Scheduling"]
```
leenagupte: You could re-write this as: `stub_user.permissions << "Scheduling"`

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -30,6 +30,22 @@
...
+      get :schedule, params: { step_by_step_page_id: step_by_step_page.id }
+
+      expect(response.status).to eq(200)
+    end
+
+    it "cannot be accessed by users without Scheduling permissions" do
+      stub_user.permissions = ["signin", "GDS Editor"]
```
leenagupte: This line isn't needed as these permissions are being set when `stub_user` is created.

spec/features/step_by_step_navigation_spec.rb
```diff
@@ -26,37 +26,57 @@
...
-    end
+  def given_I_have_scheduling_permissions
+    stub_user.permissions << "Scheduling"
   end
 
   def and_I_visit_the_navigation_rules_page
     visit step_by_step_page_navigation_rules_path(@step_by_step_page)
   end
 
+  def and_I_visit_the_publish_or_delete_page
```
leenagupte: This step already exists in `managing_step_by_step_pages_spec.rb`. If these tests are moved there, the existing `the_publish_or_delete_page` could be reused.

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
ChrisBAshton: Think this tests may need re-writing, as it passes regardless 😕

leenagupte: Yes, it passes because the scheduled_at date is in the future. That's why I originally re-created the `draft_step_by_step` object without a `scheduled_at` date.

ChrisBAshton: Understood.

ChrisBAshton: Had to duplicate the creation of the step_by_step in each test otherwise got slug uniqueness validation errors 👍

ChrisBAshton:
Thanks Leena. Think all the comments have been addressed now, and have tested it out locally 👍

spec/workers/step_by_step_scheduled_publish_worker_spec.rb
leenagupte: You shouldn't need to do that. I'd double check how and where the `step_by_step` object is being created.

ChrisBAshton: I needed to change it from:

```ruby
let(:step_by_step_page) { create(:draft_step_by_step_page, scheduled_at: future_datetime) }
```

...to:

```ruby
let!(:step_by_step_page) { create(:draft_step_by_step_page, scheduled_at: future_datetime) }
```

...otherwise the creation of the step by step didn't happen until we called it in the test, by which time we've used Timecop to travel to another time and we get an error saying "Scheduled at can't be in the past".

But by creating it early, it means if we then make another `create()` call in the test, then a second step_by_step is created in the same test run, and we get the slug uniqueness validation error.

At least the current version of the code is explicit about the test setup, so each test is easy to understand in isolation.

leenagupte: It might be possible to get around this by moving some of the tests into a separate "context" and calling `let!` from there.

leenagupte:
@ChrisBAshton Can you update the screenshots in the PR description? Thanks!

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -310,4 +342,37 @@ def then_I_see_a_page_reverted_success_notice
...
+  def and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
+    fill_in 'scheduled_at', with: '1937-04-20 10:26:51 UTC'
+  end
+
+  def when_I_submit_the_form
+    click_on 'Schedule to publish'
+  end
+
+  def and_there_should_be_no_schedule_button
+    expect(page).not_to have_content 'Schedule to publish'
```
leenagupte: Nitpick: You can check that the button itself doesn't exist.
e.g. expect(page).not_to have_css("button", text: "Schedule to publish")

AgaDufrat: I did that yesterday 🤔. Not sure why you can still see `expect(page).not_to have_content 'Schedule to publish'` Could you double check the Files tab, please? I cannot see it there.

AgaDufrat: Ah... I see now. It's still in a previous commit. _Off to fix it_

AgaDufrat:
I showed it to Sam (Kate is on leave until next Friday) and he is happy with it 🎉

PR #656, opened 2019-07-18
Add scheduled job worker

leenagupte:
Trello: https://trello.com/c/ZDJTzOzG

## What

Creates a job to schedule publication of the step by step, and processes that job on the scheduled date/time.

Scheduling a job:
* Notifies publishing-api of the intent to publish on date/time
* Adds a job to `StepByStepScheduledPublishWorker`

Processing a job:
* Only publishes step by steps with a status of "Scheduled"
* Only publishes step by steps with a scheduled date that's in the past
* Clears the `scheduled_at` date & time when the publish request is sent to publishing-api, i.e. it sets the status to "Published"
* Generates an internal change note

New jobs can be created by calling:

```
StepNavPublisher.schedule_for_publishing(step_by_step_page)
```

Note: the UI for users will be created in a separate PR.

### DIFF
--- a/app/workers/step_by_step_scheduled_publish_worker.rb
+++ b/app/workers/step_by_step_scheduled_publish_worker.rb
@@ -7,13 +7,14 @@ class StepByStepScheduledPublishWorker
   end
 
   def perform(id)
-    step_nav = nil
+    step_nav = StepByStepPage.find_by(id: id)
 
-    StepByStepPage.transaction do
-      step_nav = StepByStepPage.lock.find_by(id: id)
-      StepNavPublisher.publish(step_nav)
-      step_nav.mark_as_published
-      generate_internal_change_note(step_nav)
+    if publish_now?(step_nav)
+      step_nav.with_lock do
+        StepNavPublisher.publish(step_nav)
+        step_nav.mark_as_published
+        generate_internal_change_note(step_nav)
+      end
     end
   end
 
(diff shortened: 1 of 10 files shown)


app/workers/step_by_step_scheduled_publish_worker.rb
```diff
@@ -7,13 +7,14 @@ class StepByStepScheduledPublishWorker
...
   def perform(id)
-    step_nav = nil
+    step_nav = StepByStepPage.find_by(id: id)
 
-    StepByStepPage.transaction do
-      step_nav = StepByStepPage.lock.find_by(id: id)
-      StepNavPublisher.publish(step_nav)
-      step_nav.mark_as_published
-      generate_internal_change_note(step_nav)
+    if step_nav.scheduled_for_publishing?
```
ChrisBAshton: I see a potential problem with this:

1. Step by step is scheduled for publication on Friday
2. Step by step is unscheduled by the content designer (and from your commit message, it sounds like there's no way of clearing the sidekiq job queue at this point)
3. Step by step is scheduled for publication Saturday
4. When Friday comes around, `scheduled_for_publishing?` will be `true`, so we'll publish the step by step. But we'll be a day earlier than the content designer intended.

leenagupte: Hmmm. Good point. I'll give this another thought.

PR #657, opened 2019-07-19
Add GOVUK publishing components to collections publisher

DilwoarH:
Add the publishing component gem to collections publisher.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem 'gds-api-adapters', '~> 59.5'
 gem 'gds-sso', '~> 14.1'
 gem 'govuk_admin_template', '~> 6.7'
 gem 'govuk_app_config', '~> 1.20'
+gem 'govuk_publishing_components', '~> 17.16'
 gem 'govuk_sidekiq', '~> 3.0'
 gem 'plek', '~> 3.0'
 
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'gds-api-adapters', '~> 59.5'
 gem 'gds-sso', '~> 14.1'
 gem 'govuk_admin_template', '~> 6.7'
 gem 'govuk_app_config', '~> 1.20'
+gem "govuk_publishing_components", "~> 17.16"
```
ChrisBAshton: Can this be single quotes like the others?

app/assets/stylesheets/application.scss
```diff
@@ -2,3 +2,4 @@
 // = require_self
 // = require select2
 // = require select2-bootstrap
+@import "govuk_publishing_components/all_components";
```
ChrisBAshton: To an outsider, it looks like we pull in the CSS from all components in govuk_publishing_components, even if we only use one or two of the components. Is that what happens or is there some post-processing step somewhere?

(I'm sure your code is fine though!)

leenagupte: We're going to end up using a lot of the components when we switch to the design system, so we might as well import them all, rather than dragging them in one by one.

PR #658, opened 2019-07-19
Adds markdown editor to collections publisher

DilwoarH:
## What
Adds a markdown editor to the step editor page

## Why
This is the first phase of adding real-time feedback for markdown validation. The next phase will be to add a GovSpeak preview.

### Before
Screen Shot 2019-07-19 at 15 42 42

### After
Screen Shot 2019-07-19 at 15 40 29

### Updated after
collections-publisher dev gov uk_step-by-step-pages_1_steps_new


package-lock.json
```diff
@@ -0,0 +1,11 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "markdown-toolbar-element": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/markdown-toolbar-element/-/markdown-toolbar-element-0.2.0.tgz",
```
leenagupte: I wonder if it's the missing hash at the end of this url that's causing a problem?

https://github.com/alphagov/content-publisher/blob/1c6209ee2f11a3ae4c348bd8d7489632451fa1b6/yarn.lock#L832

leenagupte: That's the only difference I could find between the setup here an in content-publisher.

app/assets/javascripts/components/markdown-editor.js
```diff
@@ -0,0 +1 @@
+//= require markdown-toolbar-element/dist/index.umd.js
```
leenagupte: Should this file be marked as "global"? https://github.com/alphagov/content-publisher/blob/d272ddc0dd00d8469c8a58e787624f874bf3b0aa/app/assets/javascripts/components/markdown-editor.js#L1

app/views/components/_markdown-editor.html.erb
```diff
@@ -0,0 +1,52 @@
+<%
+  label_text ||= "Enter your markdown here"
+  textarea ||= {}
+  error_items ||= []
+   component_classes = %w[app-c-markdown-editor govuk-form-group]
```
leenagupte: Nitpick: I don't think this should be indented.

app/views/components/_markdown-editor.html.erb
```diff
@@ -0,0 +1,52 @@
...
+  component_classes << "govuk-form-group--error" if error_items.any?
+   textarea[:id] = textarea[:id].present? ? textarea[:id] : textarea[:name]
```
leenagupte: You're using this check in couple of places. Can it be extracted somewhere? Is `textarea` only defined in the view?

app/views/components/_markdown-editor.html.erb
```diff
@@ -0,0 +1,52 @@
+<%
+  label_text ||= "Enter your markdown here"
+  textarea ||= {}
+  error_items ||= []
+   component_classes = %w[app-c-markdown-editor govuk-form-group]
+  component_classes << "govuk-form-group--error" if error_items.any?
```
leenagupte: How is `error_items` being populated?

app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -0,0 +1,139 @@
+// adding this in here for now as this is the only place that the components are needed (right now)
+@import "govuk_publishing_components/all_components";
```
DilwoarH: This can be removed

app/views/components/_markdown-editor.html.erb
```diff
@@ -0,0 +1,44 @@
...
+%>
+<%= tag.div class: component_classes do %>
+
+  <%= render "govuk_publishing_components/components/label", {
+    html_for: textarea[:id].present? ? textarea[:id] : textarea[:name],
+    text: label_text
+  } %>
+
+   <div class="app-c-markdown-editor__container js-markdown-editor__container">
+    <div class="app-c-markdown-editor__head js-markdown-editor__head">
```
ChrisBAshton: Should the JS classes only be added with JS, rather than being in the DOM by default? (`js-markdown-editor__head` etc).

alex-ju: We're using the `js-` prefixed classes to query elements from the DOM - sort of a way of separating style from function. (e.g. this class is used for querying elements, not styling them)

ChrisBAshton: Thanks for clarifying - no problem here then!

app/views/components/_markdown-editor.html.erb
alex-ju: Uodated

app/views/components/_markdown-editor.html.erb
alex-ju: We don't need all this logic if we're explicit about the ID. Simplified the code.

app/views/components/_markdown-editor.html.erb
alex-ju: Removed this as we're not throwing errors for this field at the moment.

app/assets/stylesheets/components/_markdown-editor.scss
alex-ju: Removed

.gitignore
```diff
@@ -1,3 +1,7 @@
 log/*.log
 tmp
 db/dumps
+
```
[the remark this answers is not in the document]
alex-ju: Removed

.gitignore
```diff
@@ -1,3 +1,7 @@
...
+node_modules
+/yarn-error.log
+/npm-debug.log
```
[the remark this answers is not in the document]
alex-ju: Removed 👍

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM ruby:2.6.3
-RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs && apt-get clean
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential nodejs yarn && apt-get clean
```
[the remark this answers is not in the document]
alex-ju: Split up the commit into several. Very open to suggestions about how this should look like as you know more about the project than I do.

alex-ju:
Updated this one addressing all the comments and updated the 'after' capture.

Refactored the code a bit and added polyfills to have the markdown toolbar Web Component working down to IE11.

This is good for another round of review.

Dockerfile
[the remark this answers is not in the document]
ChrisBAshton: Could tweak this commit message to be a bit more descriptive, something like:

```
Update Dockerfile to install yarn & node dependencies

- install node and yarn
- install npm dependencies through yarn
```

alex-ju:
> The component looks good! What's missing are some tests, and some docs that describe how to use the component.

Thanks for the review @leenagupte. I've updated to setup the component guide in the app (http://collections-publisher.dev.gov.uk/component-guide/markdown_editor) then added guidance and tests for the component. We'll need to update the other two for consistency.

config/routes.rb
```diff
@@ -59,6 +59,7 @@
   post '/link_report', to: 'link_report#update'
 
   mount GovukAdminTemplate::Engine, at: "/style-guide"
+  mount GovukPublishingComponents::Engine, at: "/component-guide"
```
leenagupte: Thanks for adding this! 👍

spec/components/markdown_editor_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe "Markdown editor", type: :view do
```
leenagupte: Thanks for adding the tests! Can you fix them up to the first commit, where the component is added?

alex-ju: Sure, figured it would be easier to review what's been changed since the last review. Fixed-up now to the first commit and rebased to master.

PR #664, opened 2019-07-24
Scheduling UI revamp - Design System

ChrisBAshton:
This PR splits up the datetime text field into [Date](https://govuk-publishing-components.herokuapp.com/component-guide/date_input) and time (the latter of which is based on the autocomplete component copied from [Content Publisher](https://github.com/alphagov/content-publisher/)). The time field now provides a dropdown list of half-hourly suggested times - and a special case of 12:01am - but can also take any arbitrary time. These fields' placeholder values default to '9am tomorrow'. It stores this schedule time in London timezone format, so that BST is accounted for.

In addition, this PR also swaps out the Bootstrap submit button with [GOV.UK submit button](https://govuk-publishing-components.herokuapp.com/component-guide/button), and changes the redirect value of the cancel button to a more appropriate location.

| Before | After |
|--------|------|
|61949262-20b7f400-afa2-11e9-9103-9197a389c11f|Screen Shot 2019-08-08 at 12 17 25|

This design uses Content Publisher for inspiration. Note that I've not changed the error title "There is a problem", purely because "You need to" may not apply to other form error messages outside of scheduling.

| Error message content publisher | Error message collections publisher |
|----------------------------------|-------------------------------------|
|Screen Shot 2019-07-30 at 10 52 53|Screen Shot 2019-07-30 at 15 24 33|

https://trello.com/c/oFvkJGff/44-use-govuk-components-in-the-scheduling-interface

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -1,0 +1,63 @@
+//= require accessible-autocomplete/dist/accessible-autocomplete.min.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Autocomplete () { }
+
+  Autocomplete.prototype.start = function ($module) {
+    this.$module = $module[0]
+
+    var $input = this.$module.querySelector('input')
+
+    if ($input) {
+      this.initAutoCompleteInput($input)
+    }
+  }
+
+  Autocomplete.prototype.initAutoCompleteInput = function ($input) {
+    var withoutNarrowingResults = this.$module.dataset.autocompleteWithoutNarrowingResults
+
+    var list = document.getElementById($input.getAttribute('list'))
+    var options = []
+
+    if (list) {
+      options = [].map.call(list.querySelectorAll('option'), function (option) {
+        return option.value
+      })
+    }
+
+    if (!options.length) {
+      return
+    }
+
+    new window.accessibleAutocomplete({ // eslint-disable-line no-new, new-cap
+      id: $input.id,
+      name: $input.name,
+      element: this.$module,
+      showAllValues: withoutNarrowingResults,
+      defaultValue: $input.value,
+      autoselect: !withoutNarrowingResults,
+      dropdownArrow: withoutNarrowingResults ? this.dropdownArrow : null,
+      source: function (query, syncResults) {
+        if (withoutNarrowingResults) {
+          syncResults(options)
+        } else {
+          syncResults(query
+            ? options.filter(function (option) {
+              return option.toLowerCase().indexOf(query.toLowerCase()) !== -1
+            }) : []
+          )
+        }
+      }
+    })
+
+    $input.parentNode.removeChild($input)
+  }
+
+  Autocomplete.prototype.dropdownArrow = function (config) {
+    return '<svg class="' + config.className + '" style="top: 8px;" viewBox="0 0 512 512" ><path d="M256,298.3L256,298.3L256,298.3l174.2-167.2c4.3-4.2,11.4-4.1,15.8,0.2l30.6,29.9c4.4,4.3,4.5,11.3,0.2,15.5L264.1,380.9  c-2.2,2.2-5.2,3.2-8.1,3c-3,0.1-5.9-0.9-8.1-3L35.2,176.7c-4.3-4.2-4.2-11.2,0.2-15.5L66,131.3c4.4-4.3,11.5-4.4,15.8-0.2L256,298.3  z"/></svg>'
+  }
+
+  Modules.Autocomplete = Autocomplete
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,1 +1,2 @@
 @import "govuk_publishing_components/all_components";
+@import "./components/autocomplete";
--- a/app/assets/stylesheets/components/_autocomplete.scss
+++ b/app/assets/stylesheets/components/_autocomplete.scss
@@ -1,0 +1,46 @@
+@import url(asset-path("accessible-autocomplete/dist/accessible-autocomplete.min.css"));
+
+// TODO: Replace with $govuk-focus-colour when the new colour scheme is released
+$app-focus-colour: #ffdd00;
+
+.autocomplete__wrapper {
+  background: govuk-colour("white");
+}
+
+.autocomplete__input {
+  @include govuk-font(19);
+  z-index: 1;
+  color: $govuk-text-colour;
+}
+
+.autocomplete__dropdown-arrow-down {
+  z-index: 0;
+}
+
+.autocomplete__option,
+.autocomplete__selected-option {
+  @include govuk-font(19);
+  color: $govuk-text-colour;
+
+  mark {
+    border-bottom: 3px solid $govuk-focus-text-colour;
+    background: $app-focus-colour;
+  }
+}
+
+.autocomplete__option--focused,
+.autocomplete__option:hover {
+  color: govuk-colour("white");
+
+  .autocomplete__option-hint {
+    color: govuk-colour("white");
+  }
+}
+
+.app-c-autocomplete--narrow {
+  max-width: 200px;
+
+  @include govuk-media-query($from: tablet) {
+    max-width: 230px;
+  }
+}
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -77,9 +77,12 @@ class StepByStepPagesController < ApplicationController
   def schedule
     set_current_page_as_step_by_step
     if request.post?
-      if @step_by_step_page.update_attributes(scheduled_at: params[:scheduled_at])
+      date_params = params[:schedule][:date].permit(:year, :month, :day)
+      parser = DatetimeParser.new(date: date_params, time: params[:schedule][:time])
+      scheduled_at = parser.parse
+      if scheduled_at && @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
-        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{params[:scheduled_at]}"
+        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{scheduled_at}"
         generate_internal_change_note(note_description)
         set_change_note_version
         redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled to publish."
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -1,5 +1,6 @@
 class StepByStepPagesController < ApplicationController
   include PublishingApiHelper
+  include TimeOptionsHelper
   layout 'admin_layout'
 
   before_action :require_gds_editor_permissions!
@@ -77,8 +78,10 @@ class StepByStepPagesController < ApplicationController
   def schedule
     set_current_page_as_step_by_step
     if request.post?
-      date_params = params[:schedule][:date].permit(:year, :month, :day)
-      parser = DatetimeParser.new(date: date_params, time: params[:schedule][:time])
+      date_params = params[:schedule][:date].permit(:year, :month, :day).to_h.symbolize_keys
+      time_param = params[:schedule][:time]
+      @schedule_placeholder = default_datetime_placeholder(date_params.merge(time: time_param))
+      parser = DatetimeParser.new(date: date_params, time: time_param)
       scheduled_at = parser.parse
       if scheduled_at && @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
@@ -89,6 +92,8 @@ class StepByStepPagesController < ApplicationController
       else
         render :schedule
       end
+    else
+      @schedule_placeholder = default_datetime_placeholder
     end
   end
 
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -90,7 +90,7 @@ class StepByStepPagesController < ApplicationController
         render :schedule
       elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
-        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{scheduled_at}"
+        note_description = "Minor update scheduled by #{current_user.name} for publishing on #{format_full_date_and_time(scheduled_at)}"
         generate_internal_change_note(note_description)
         set_change_note_version
         redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled to publish."
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -84,6 +84,9 @@ class StepByStepPagesController < ApplicationController
       @parser = DatetimeParser.new(date: date_params, time: time_param)
       scheduled_at = @parser.parse
       if @parser.issues.any?
+        @parser.issues.each do |issue|
+          @step_by_step_page.errors.add :base, issue.values.first
+        end
         render :schedule
       elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -81,9 +81,11 @@ class StepByStepPagesController < ApplicationController
       date_params = params[:schedule][:date].permit(:year, :month, :day).to_h.symbolize_keys
       time_param = params[:schedule][:time]
       @schedule_placeholder = default_datetime_placeholder(date_params.merge(time: time_param))
-      parser = DatetimeParser.new(date: date_params, time: time_param)
-      scheduled_at = parser.parse
-      if scheduled_at && @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
+      @parser = DatetimeParser.new(date: date_params, time: time_param)
+      scheduled_at = @parser.parse
+      if @parser.issues.any?
+        render :schedule
+      elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
         note_description = "Minor update scheduled by #{current_user.name} for publishing at #{scheduled_at}"
         generate_internal_change_note(note_description)
@@ -143,6 +145,13 @@ class StepByStepPagesController < ApplicationController
   def internal_change_notes
     set_current_page_as_step_by_step
     @internal_change_note = InternalChangeNote.new
+  end
+
+  helper_method :issues_for
+  def issues_for(namespace)
+    return if @parser.nil?
+
+    @parser.issues_for(namespace).map { |error| { text: error } }
   end
 
 private
--- a/app/helpers/time_options_helper.rb
+++ b/app/helpers/time_options_helper.rb
@@ -16,6 +16,10 @@ module TimeOptionsHelper
     datetime.strftime("%-l:%M%P")
   end
 
+  def format_full_date_and_time(datetime)
+    datetime.strftime('%A, %d %B %Y at %-l:%M %P')
+  end
+
   def default_datetime_placeholder(
     year: 1.day.from_now.year,
     month: 1.day.from_now.month,
--- a/app/helpers/time_options_helper.rb
+++ b/app/helpers/time_options_helper.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module TimeOptionsHelper
+  def time_options
+    %w[am pm].flat_map do |period|
+      hours = [12] + (1..11).to_a
+      hours.flat_map do |hour|
+        next ["12:01am", "12:30am"] if hour == 12 && period == "am"
+
+        ["#{hour}:00#{period}", "#{hour}:30#{period}"]
+      end
+    end
+  end
+end
--- a/app/helpers/time_options_helper.rb
+++ b/app/helpers/time_options_helper.rb
@@ -11,4 +11,22 @@ module TimeOptionsHelper
       end
     end
   end
+
+  def format_time_12_hour_clock(datetime)
+    datetime.strftime("%-l:%M%P")
+  end
+
+  def default_datetime_placeholder(
+    year: 1.day.from_now.year,
+    month: 1.day.from_now.month,
+    day: 1.day.from_now.day,
+    time: format_time_12_hour_clock(1.day.from_now.change(hour: 9))
+  )
+    {
+      year: year,
+      month: month,
+      day: day,
+      time: time,
+    }
+  end
 end
--- a/app/services/datetime_parser.rb
+++ b/app/services/datetime_parser.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+class DatetimeParser
+  def initialize(date:, time:)
+    @raw_date = date.to_h
+    @raw_time = time.to_s
+  end
+
+  def parse
+    begin
+      check_date_is_valid
+      check_time_is_valid
+
+      Time.find_zone("London").local(
+        raw_date[:year].to_i,
+        raw_date[:month].to_i,
+        raw_date[:day].to_i,
+        time[:hour],
+        time[:minute]
+      )
+    rescue ArgumentError
+      nil
+    end
+  end
+
+private
+
+  attr_reader :raw_date, :raw_time
+
+  def check_date_is_valid
+    day, month, year = raw_date.values_at(:day, :month, :year)
+    Date.strptime("#{day}-#{month}-#{year}", "%d-%m-%Y")
+  end
+
+  def check_time_is_valid
+    raise ArgumentError unless parsed_time_values
+  end
+
+  def parsed_time_values
+    @parsed_time_values ||= raw_time.match(%r{
+      \A
+      (?<hour>(2[0-3]|1[0-9]|0?[0-9]))
+      :
+      (?<minute>[0-5][0-9])
+      (\s?(?<period>(am|pm)))?
+      \Z
+    }ix)
+  end
+
+  def time
+    @time ||= begin
+      hour = parsed_time_values[:hour].to_i
+      period = parsed_time_values[:period]&.downcase
+      minute = parsed_time_values[:minute].to_i
+
+      if period == "am" && hour == 12
+        hour = 0
+      elsif period == "pm" && hour < 12
+        hour += 12
+      end
+
+      { hour: hour, minute: minute }
+    end
+  end
+end
--- a/app/services/datetime_parser.rb
+++ b/app/services/datetime_parser.rb
@@ -1,26 +1,30 @@
 # frozen_string_literal: true
 
 class DatetimeParser
+  attr_reader :issues
+
   def initialize(date:, time:)
     @raw_date = date.to_h
     @raw_time = time.to_s
+    @issues = []
+  end
+
+  def issues_for(param)
+    issues.pluck(param).compact
   end
 
   def parse
-    begin
-      check_date_is_valid
-      check_time_is_valid
+    check_date_is_valid
+    check_time_is_valid
+    return if issues.any?
 
-      Time.find_zone("London").local(
-        raw_date[:year].to_i,
-        raw_date[:month].to_i,
-        raw_date[:day].to_i,
-        time[:hour],
-        time[:minute]
-      )
-    rescue ArgumentError
-      nil
-    end
+    Time.find_zone("London").local(
+      raw_date[:year].to_i,
+      raw_date[:month].to_i,
+      raw_date[:day].to_i,
+      time[:hour],
+      time[:minute]
+    )
   end
 
 private
@@ -30,10 +34,14 @@ private
   def check_date_is_valid
     day, month, year = raw_date.values_at(:day, :month, :year)
     Date.strptime("#{day}-#{month}-#{year}", "%d-%m-%Y")
+  rescue ArgumentError
+    issues << { date: "Enter a valid date" }
   end
 
   def check_time_is_valid
-    raise ArgumentError unless parsed_time_values
+    unless parsed_time_values
+      issues << { time: "Enter a valid time" }
+    end
   end
 
   def parsed_time_values
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -1,0 +1,51 @@
+<%
+  id ||= "autocomplete-#{SecureRandom.hex(4)}"
+  type ||= nil
+  error_id = "error-#{SecureRandom.hex(4)}"
+  error_items ||= []
+  width ||= nil
+  data_attributes ||= {}
+  input ||= {}
+  name ||= nil
+  label = label.symbolize_keys if label
+  aria = error_id if error_items.any?
+
+  root_classes = %w(app-c-autocomplete govuk-form-group)
+  root_classes << "app-c-autocomplete--#{width}" if width
+  root_classes << "govuk-form-group--error" if error_items.any?
+
+  if name && label
+%>
+  <%= tag.div class: root_classes,
+              data: data_attributes.merge(module: "autocomplete", "autocomplete-type": type) do %>
+    <%= tag.fieldset class: "govuk-fieldset", "aria-describedby": aria do %>
+      <%= render "govuk_publishing_components/components/label", {
+        html_for: id
+      }.merge(label) %>
+
+      <% if error_items.any? %>
+        <%= render "govuk_publishing_components/components/error_message", {
+          id: error_id,
+          items: error_items,
+        } %>
+      <% end %>
+
+      <% options = Array(input[:options]) %>
+
+      <%= tag.input name: name,
+                    value: input[:value],
+                    class: "govuk-input",
+                    id: id,
+                    list: options.any? ? "#{id}-list" : nil,
+                    type: "text" %>
+
+      <% if options.any? %>
+        <%= tag.datalist id: "#{id}-list" do %>
+          <% options.each do |option| %>
+            <%= tag.option(value: option) %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -39,6 +39,7 @@
         name: "schedule[date]",
         hint: "For example, 01 08 2027",
         id: "date",
+        error_items: issues_for(:date),
         items: [
           {
             name: "day",
@@ -69,6 +70,7 @@
           options: time_options,
           value: @schedule_placeholder[:time],
         },
+        error_items: issues_for(:time),
         data_attributes: {
           "autocomplete-without-narrowing-results": true,
         },
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -42,15 +42,18 @@
         items: [
           {
             name: "day",
-            width: 2
+            width: 2,
+            value: @schedule_placeholder[:day]
           },
           {
             name: "month",
-            width: 2
+            width: 2,
+            value: @schedule_placeholder[:month]
           },
           {
             name: "year",
-            width: 4
+            width: 4,
+            value: @schedule_placeholder[:year]
           }
         ]
       } %>
@@ -64,6 +67,7 @@
         },
         input: {
           options: time_options,
+          value: @schedule_placeholder[:time],
         },
         data_attributes: {
           "autocomplete-without-narrowing-results": true,
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -84,7 +84,7 @@
           } %>
         </div>
         <div class="govuk-grid-column-full govuk-!-margin-top-3">
-          <%= link_to "Cancel", @step_by_step_page, class: "govuk-link" %>
+          <%= link_to "Cancel", step_by_step_page_publish_or_delete_path(@step_by_step_page), class: "govuk-link" %>
         </div>
       </div>
     <% end %>
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -31,7 +31,7 @@
   <div class="col-md-7">
     <%= form_tag do %>
       <div class="form-group">
-        <%= label_tag :scheduled_at, 'Schedule datetime (e.g. "2020-04-20 10:26:51 UTC")' %>
+        <%= label_tag :scheduled_at, 'Schedule datetime (e.g. "2020-04-20 10:26:51 London")' %>
         <%= text_field_tag :scheduled_at, params[:scheduled_at]  %>
       </div>
 
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -62,6 +62,9 @@
           text: "Time",
           bold: true,
         },
+        input: {
+          options: time_options,
+        },
         data_attributes: {
           "autocomplete-without-narrowing-results": true,
         },
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -77,9 +77,15 @@
         width: "narrow",
       } %>
 
-      <div class="form-group">
-        <%= submit_tag "Schedule to publish", class: "btn btn-primary" %>
-        <%= link_to "Cancel", @step_by_step_page %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full">
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Schedule to publish",
+          } %>
+        </div>
+        <div class="govuk-grid-column-full govuk-!-margin-top-3">
+          <%= link_to "Cancel", @step_by_step_page, class: "govuk-link" %>
+        </div>
       </div>
     <% end %>
   </div>
--- a/app/views/step_by_step_pages/schedule.html.erb
+++ b/app/views/step_by_step_pages/schedule.html.erb
@@ -14,6 +14,10 @@
   ]
 %>
 
+<% legend = capture do %>
+  <span class="govuk-heading-s govuk-!-margin-bottom-0">Date</span>
+<% end %>
+
 <%= render 'shared/steps/step_breadcrumb', links: links %>
 
 <%= render 'shared/steps/heading',
@@ -30,10 +34,39 @@
 <div class="row">
   <div class="col-md-7">
     <%= form_tag do %>
-      <div class="form-group">
-        <%= label_tag :scheduled_at, 'Schedule datetime (e.g. "2020-04-20 10:26:51 London")' %>
-        <%= text_field_tag :scheduled_at, params[:scheduled_at]  %>
-      </div>
+      <%= render "govuk_publishing_components/components/date_input", {
+        legend_text: legend,
+        name: "schedule[date]",
+        hint: "For example, 01 08 2027",
+        id: "date",
+        items: [
+          {
+            name: "day",
+            width: 2
+          },
+          {
+            name: "month",
+            width: 2
+          },
+          {
+            name: "year",
+            width: 4
+          }
+        ]
+      } %>
+
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: "Time",
+          bold: true,
+        },
+        data_attributes: {
+          "autocomplete-without-narrowing-results": true,
+        },
+        width: "narrow",
+      } %>
 
       <div class="form-group">
         <%= submit_tag "Schedule to publish", class: "btn btn-primary" %>
--- a/spec/controllers/step_by_step_pages_controller_spec.rb
+++ b/spec/controllers/step_by_step_pages_controller_spec.rb
@@ -5,6 +5,7 @@ require "gds_api/test_helpers/publishing_api"
 RSpec.describe StepByStepPagesController do
   include GdsApi::TestHelpers::PublishingApiV2
   include GdsApi::TestHelpers::PublishingApi
+  include TimeOptionsHelper
 
   let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
   let(:stub_user) { create(:user, name: "Name Surname", permissions: ["signin", "GDS Editor"]) }
@@ -114,7 +115,7 @@ RSpec.describe StepByStepPagesController do
       schedule_for_future
 
       expect(step_by_step_page.scheduled_at.class.name).to eq 'Time'
-      expect(step_by_step_page.scheduled_at.in_time_zone("London").strftime('%d/%m/%Y %H:%M:%S')).to eq '20/04/2030 10:26:00'
+      expect(format_full_date_and_time(step_by_step_page.scheduled_at.in_time_zone("London"))).to eq 'Saturday, 20 April 2030 at 10:26 am'
     end
 
     it "sets the status to Scheduled" do
--- a/spec/controllers/step_by_step_pages_controller_spec.rb
+++ b/spec/controllers/step_by_step_pages_controller_spec.rb
@@ -90,12 +90,32 @@ RSpec.describe StepByStepPagesController do
     end
   end
 
+  describe "#schedule" do
+    let(:step_by_step_page) { create(:draft_step_by_step_page) }
+
+    before :each do
+      stub_scheduling_permissions
+      stub_publishing_api_for_scheduling
+    end
+
+    it "sets `scheduled_at` to a datetime" do
+      schedule_publishing(step_by_step_page, "2030-04-20 10:26 London")
+
+      expect(step_by_step_page.scheduled_at.class.name).to eq 'Time'
+      expect(step_by_step_page.scheduled_at.in_time_zone("London").strftime('%d/%m/%Y %H:%M:%S')).to eq '20/04/2030 10:26:00'
+    end
+
+    it "sets the status to Scheduled" do
+      schedule_publishing(step_by_step_page)
+
+      expect(step_by_step_page.status[:name]).to eq 'scheduled'
+    end
+  end
+
   describe "#unschedule" do
     before :each do
-      stub_user.permissions << "Scheduling"
-
-      allow(Services.publishing_api).to receive(:get_content)
-        .and_return(content_item(step_by_step_page))
+      stub_scheduling_permissions
+      stub_publishing_api_for_scheduling
     end
 
     it "clears Scheduled status and sets it back to Draft" do
@@ -173,6 +193,21 @@ RSpec.describe StepByStepPagesController do
 
     stub_any_publishing_api_put_content
     stub_any_publishing_api_publish
+  end
+
+  def stub_scheduling_permissions
+    stub_user.permissions << "Scheduling"
+  end
+
+  def stub_publishing_api_for_scheduling
+    allow(Services.publishing_api).to receive(:get_content)
+      .and_return(content_item(step_by_step_page))
+  end
+
+  def schedule_publishing(step_by_step_page, schedule_time = (Time.zone.now + 1.day).to_s)
+    stub_default_publishing_api_put_intent
+    post :schedule, params: { step_by_step_page_id: step_by_step_page.id, scheduled_at: schedule_time }
+    step_by_step_page.reload
   end
 
   def unschedule_publishing(step_by_step_page)
--- a/spec/controllers/step_by_step_pages_controller_spec.rb
+++ b/spec/controllers/step_by_step_pages_controller_spec.rb
@@ -96,17 +96,29 @@ RSpec.describe StepByStepPagesController do
     before :each do
       stub_scheduling_permissions
       stub_publishing_api_for_scheduling
+      stub_default_publishing_api_put_intent
+    end
+
+    def schedule_for_future
+      post :schedule, params: {
+        step_by_step_page_id: step_by_step_page.id,
+        schedule: {
+          date: { year: "2030", month: "04", day: "20" },
+          time: "10:26am",
+        },
+      }
+      step_by_step_page.reload
     end
 
     it "sets `scheduled_at` to a datetime" do
-      schedule_publishing(step_by_step_page, "2030-04-20 10:26 London")
+      schedule_for_future
 
       expect(step_by_step_page.scheduled_at.class.name).to eq 'Time'
       expect(step_by_step_page.scheduled_at.in_time_zone("London").strftime('%d/%m/%Y %H:%M:%S')).to eq '20/04/2030 10:26:00'
     end
 
     it "sets the status to Scheduled" do
-      schedule_publishing(step_by_step_page)
+      schedule_for_future
 
       expect(step_by_step_page.status[:name]).to eq 'scheduled'
     end
@@ -202,12 +214,6 @@ RSpec.describe StepByStepPagesController do
   def stub_publishing_api_for_scheduling
     allow(Services.publishing_api).to receive(:get_content)
       .and_return(content_item(step_by_step_page))
-  end
-
-  def schedule_publishing(step_by_step_page, schedule_time = (Time.zone.now + 1.day).to_s)
-    stub_default_publishing_api_put_intent
-    post :schedule, params: { step_by_step_page_id: step_by_step_page.id, scheduled_at: schedule_time }
-    step_by_step_page.reload
   end
 
   def unschedule_publishing(step_by_step_page)
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -9,8 +9,6 @@ RSpec.feature "Managing step by step pages" do
   include GdsApi::TestHelpers::LinkCheckerApi
   include GdsApi::TestHelpers::PublishingApi
 
-  let(:schedule_time) { "2030-04-20 10:26:51 London" }
-
   before do
     given_I_am_a_GDS_editor
     setup_publishing_api
@@ -153,7 +151,7 @@ RSpec.feature "Managing step by step pages" do
       when_I_submit_the_form
       then_I_should_see "has been scheduled to publish"
       and_the_step_by_step_should_have_the_status "Scheduled"
-      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at #{schedule_time}"
+      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at 2030-04-20 10:26:00 +0100"
       and_the_step_by_step_is_not_editable
     end
 
@@ -393,11 +391,18 @@ RSpec.feature "Managing step by step pages" do
   end
 
   def and_I_fill_in_the_scheduling_form
-    fill_in 'scheduled_at', with: schedule_time
+    fill_in 'schedule[date][year]', with: "2030"
+    fill_in 'schedule[date][month]', with: "04"
+    fill_in 'schedule[date][day]', with: "20"
+    fill_in 'schedule[time]', with: "10:26am"
   end
 
   def and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
-    fill_in 'scheduled_at', with: '1937-04-20 10:26:51 London'
+    yesterday = Time.current - 1.day
+    fill_in 'schedule[date][year]', with: yesterday.year
+    fill_in 'schedule[date][month]', with: yesterday.month
+    fill_in 'schedule[date][day]', with: yesterday.day
+    fill_in 'schedule[time]', with: "10:26am"
   end
 
   def when_I_submit_the_form
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -9,6 +9,8 @@ RSpec.feature "Managing step by step pages" do
   include GdsApi::TestHelpers::LinkCheckerApi
   include GdsApi::TestHelpers::PublishingApi
 
+  let(:yesterday) { Time.current - 1.day }
+
   before do
     given_I_am_a_GDS_editor
     setup_publishing_api
@@ -158,9 +160,11 @@ RSpec.feature "Managing step by step pages" do
     scenario "User tries to schedule publishing for date in the past" do
       given_there_is_a_draft_step_by_step_page
       and_I_visit_the_scheduling_page
+      then_inputs_should_have_tomorrows_date
       and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
       when_I_submit_the_form
       then_I_should_see "Scheduled at can't be in the past"
+      and_I_can_still_see_the_date_and_time_values_I_entered
       and_the_step_by_step_should_have_the_status "Draft"
     end
 
@@ -398,11 +402,23 @@ RSpec.feature "Managing step by step pages" do
   end
 
   def and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
-    yesterday = Time.current - 1.day
     fill_in 'schedule[date][year]', with: yesterday.year
     fill_in 'schedule[date][month]', with: yesterday.month
     fill_in 'schedule[date][day]', with: yesterday.day
     fill_in 'schedule[time]', with: "10:26am"
+  end
+
+  def then_inputs_should_have_tomorrows_date
+    tomorrow = Time.current.tomorrow
+    expect(find_field('schedule[date][year]').value).to eq tomorrow.year.to_s
+    expect(find_field('schedule[date][month]').value).to eq tomorrow.month.to_s
+    expect(find_field('schedule[date][day]').value).to eq tomorrow.day.to_s
+    expect(find_field('schedule[time]').value).to eq "9:00am"
+  end
+
+  def and_I_can_still_see_the_date_and_time_values_I_entered
+    expect(find_field('schedule[date][day]').value).to eq yesterday.day.to_s
+    expect(find_field('schedule[time]').value).to eq "10:26am"
   end
 
   def when_I_submit_the_form
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -184,6 +184,15 @@ RSpec.feature "Managing step by step pages" do
       and_the_step_by_step_should_have_the_status "Draft"
       and_there_should_be_a_change_note "Publishing was unscheduled by Test author."
     end
+
+    scenario "User tries using invalid values for schedule date and time" do
+      given_there_is_a_draft_step_by_step_page
+      and_I_visit_the_scheduling_page
+      and_I_fill_in_the_scheduling_form_with_nonsense_data
+      when_I_submit_the_form
+      then_I_should_see "Enter a valid date", :within_the_date_component
+      and_I_should_see "Enter a valid time", :within_the_time_component
+    end
   end
 
   def and_it_has_change_notes
@@ -408,6 +417,13 @@ RSpec.feature "Managing step by step pages" do
     fill_in 'schedule[time]', with: "10:26am"
   end
 
+  def and_I_fill_in_the_scheduling_form_with_nonsense_data
+    fill_in 'schedule[date][year]', with: Time.current.year
+    fill_in 'schedule[date][month]', with: ""
+    fill_in 'schedule[date][day]', with: Time.current.day
+    fill_in 'schedule[time]', with: "foo"
+  end
+
   def then_inputs_should_have_tomorrows_date
     tomorrow = Time.current.tomorrow
     expect(find_field('schedule[date][year]').value).to eq tomorrow.year.to_s
@@ -433,9 +449,21 @@ RSpec.feature "Managing step by step pages" do
     and_there_should_be_no_schedule_button
   end
 
-  def then_I_should_see(content)
-    expect(page).to have_content content
+  def then_I_should_see(content, scope = nil)
+    scope_selector = case scope
+                     when :within_the_date_component
+                       'form > .govuk-form-group:not(.app-c-autocomplete)'
+                     when :within_the_time_component
+                       'form > .app-c-autocomplete'
+                     else
+                       'body'
+                     end
+    within scope_selector do
+      expect(page).to have_content content
+    end
   end
+
+  alias_method :and_I_should_see, :then_I_should_see
 
   def and_the_step_by_step_should_have_the_status(status)
     visit step_by_step_pages_url
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -153,7 +153,7 @@ RSpec.feature "Managing step by step pages" do
       when_I_submit_the_form
       then_I_should_see "has been scheduled to publish"
       and_the_step_by_step_should_have_the_status "Scheduled"
-      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at 2030-04-20 10:26:00 +0100"
+      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing on Saturday, 20 April 2030 at 10:26 am"
       and_the_step_by_step_is_not_editable
     end
 
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -190,7 +190,9 @@ RSpec.feature "Managing step by step pages" do
       and_I_visit_the_scheduling_page
       and_I_fill_in_the_scheduling_form_with_nonsense_data
       when_I_submit_the_form
-      then_I_should_see "Enter a valid date", :within_the_date_component
+      then_I_should_see "Enter a valid date", :at_the_top_of_the_page
+      and_I_should_see "Enter a valid time", :at_the_top_of_the_page
+      and_I_should_see "Enter a valid date", :within_the_date_component
       and_I_should_see "Enter a valid time", :within_the_time_component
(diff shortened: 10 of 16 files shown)


app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module DateTimeHelper
```
ChrisBAshton: Don't think this is used

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -76,15 +76,25 @@ def publish
...
         generate_internal_change_note(note_description)
         set_change_note_version
         redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled to publish."
       else
         render :schedule
       end
+      # rescue StandardError => e
+      #   @step_by_step_page.errors.add :base, e.message
+      #   render :schedule
+      # end
```
ChrisBAshton: 👀

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -76,9 +76,17 @@ def publish
...
+      scheduled_at = parser.parse
+      if parser.issues.any?
+        parser.issues.each do |err|
+          @step_by_step_page.errors.add err.keys.first, err.values.first
+        end
+        render :schedule
+      elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
-        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{params[:scheduled_at]}"
+        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{scheduled_at}"
```
ChrisBAshton: This method is getting big now. Refactor?

app/services/datetime_parser.rb
```diff
@@ -40,13 +40,13 @@ def check_date_is_valid
...
     field_name = "#{issue_prefix}_date".to_sym
-    issue_items << Hash[field_name, "Invalid date"]
+    issue_items << Hash[field_name, "Enter a valid date"]
   end
 
   def check_time_is_valid
     if !parsed_time_values || time[:hour] > 23 || (parsed_time_values[:hour].to_i > 12 && parsed_time_values[:period])
       field_name = "#{issue_prefix}_time".to_sym
-      issue_items << Hash[field_name, "Invalid time"]
+      issue_items << Hash[field_name, "Enter a valid time"]
```
ChrisBAshton: Fix this up to previous commit.

spec/services/datetime_parser_spec.rb
```diff
@@ -83,10 +83,10 @@
       expect(relevant_issues.count).to eq(1)
 
       date_form_message = relevant_issues.first
-      expect(date_form_message).to eq("Invalid date")
+      expect(date_form_message).to eq("Enter a valid date")
```
ChrisBAshton: Ditto.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -60,10 +60,28 @@
...
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: t("schedule.edit.time"),
+          bold: true,
+        },
+        error_items: time_errors,
+        input: {
+          value: params.dig(:schedule, :time) || publish_time.strftime("%-l:%M%P"),
```
ChrisBAshton: Content Publisher has a helper for formatting the time nicely. I think we can add that in a future PR and at the same time change the datetime formatting on the publish-or-delete screen for scheduled step by steps. Think it's out of scope for this PR.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -60,10 +60,28 @@
...
+        }
+      ] : false %>
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: t("schedule.edit.time"),
+          bold: true,
+        },
+        error_items: time_errors,
```
ChrisBAshton: I'm happy to make a change to the autocomplete component so that it can take an `error_message` like date_input does. But also quite keen to avoid making modifications to anything copied over from Content Publisher, so that we don't end up with two forks that grow further apart.

app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,7 @@
-// = require_tree .
-// = require_self
-// = require select2
-// = require select2-bootstrap
```
ChrisBAshton: Dilwoar seems to think we'll still need the last two lines above ^, e.g. `@import "select2".

Doing so breaks my CSS, and not doing so doesn't seem to break anything 🤔

spec/services/datetime_parser_spec.rb
```diff
@@ -70,22 +72,21 @@
...
 
-      date_form_message = parser.issues.items_for(:schedule_date).first[:text]
-      expect(date_form_message).to eq(
-        I18n.t!("requirements.schedule_date.invalid.form_message"),
-      )
+      relevant_issues = parser.issues_for(:schedule_date)
+      expect(relevant_issues.count).to eq(1)
+
+      date_form_message = relevant_issues.first
+      expect(date_form_message).to eq("Enter a valid date")
```
ChrisBAshton: We don't have any tests that this error message actually appears in the page. (Nor its placement in the page, i.e. inline in the relevant component, or in the 'global' error at the top of the page).

Wondering if it's worth doing or if these unit tests are sufficient.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+
+class DatetimeParser
+  def initialize(date:, time:, issue_prefix:)
+    @raw_date = date.to_h
+    @raw_time = time.to_s
+    @issue_prefix = issue_prefix
+  end
+
+  def issues
+    Requirements::CheckerIssues.new(issue_items || [])
```
leenagupte: I don't think copying and pasting the class from Content Publisher like this works. This class `Requirements::CheckerIssues` doesn't exist in Collections Publisher, neither does `Requirements::Issue`, so the tests must have failed when this code was added.

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    it "returns an issue when the date is invalid" do
+      params = { date: { day: "10", month: "60", year: "11" },
+                 time: "10:00",
+                 issue_prefix: :schedule }
+      parser = DatetimeParser.new(params)
+      parser.parse
+
+      form_message = parser.issues.items_for(:schedule_date).first[:text]
+      expect(form_message).to eq(
+        I18n.t!("requirements.schedule_date.invalid.form_message"),
```
leenagupte: We're not using internationalisation to define messages, so these tests must have failed as well.

app/views/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,68 @@
+<%
+  id ||= "autocomplete-#{SecureRandom.hex(4)}"
```
leenagupte: If this auto complete is now being used in two applications, should it instead be made into a component in govuk_publishing_components?

cc @DilwoarH

app/helpers/time_options_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module TimeOptionsHelper
```
leenagupte: Can you add a test for this helper?

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -60,10 +60,28 @@
...
+      <% time_errors = @step_by_step_page.issues.include?(:schedule_time) ? [
+        {
+          text: @step_by_step_page.issues[:schedule_time],
+        }
+      ] : false %>
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: t("schedule.edit.time"),
```
leenagupte: Where does this text come from?

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -60,10 +60,28 @@
         ]
       } %>
 
-      <div class="form-group">
-        <%= label_tag "schedule[time]", 'Time' %>
-        <%= text_field_tag "schedule[time]" %>
-      </div>
+      <% time_errors = @step_by_step_page.issues.include?(:schedule_time) ? [
```
leenagupte: This code should be moved to the controller for readability.

app/views/step_by_step_pages/schedule.html.erb
leenagupte: Perhaps this code could be moved to a presenter or a helper class so we can test that we're passing the right things to the component.

app/views/step_by_step_pages/schedule.html.erb
leenagupte: What is this check doing? What is the difference between `params` and `publish_time`.

Can this logic be moved out of the view?

ChrisBAshton:
I thought it would be valuable to copy across the files from Content Publisher as-is, [in one commit](https://github.com/alphagov/collections-publisher/pull/664/commits/405aa92a65f23baa3b15e14b5518ad8d3cc3e931), and then make the modifications in a [separate commit](https://github.com/alphagov/collections-publisher/pull/664/commits/af83e31f79dce067a829b6bd9d52ac257af9ca8f) so that it's really clear what we've changed about the files. Most of your comments seem to be to do with that first commit, which I can easily resolve by squashing the first two commits together for the sake of each commit being in a test-passable state, but I think we lose some useful information in doing that. Let me know if I should proceed.

Happy to write more tests, I'll work on that this morning.

app/views/components/_autocomplete.html.erb
DilwoarH: This is already in there (https://govuk-publishing-components.herokuapp.com/component-guide/accessible_autocomplete) but I think this version has additional features which the gem version doesn't. I will have a look to see if we can move it to the gem.

app/assets/stylesheets/forms.scss
```diff
@@ -24,15 +24,15 @@ form label {
...
+input[type='text']:not(.govuk-input, .autocomplete__input) {
   @extend .form-control; // scss-lint:disable PlaceholderInExtend
 }
 
 .errors input {
   border-color: $state-danger-border;
 }
 
-button {
+button:not(.govuk-button) {
```
DilwoarH: I've solved this problem on My PR #665 by not including any of the bootstrap stuff. If my PR gets merged in before this then we won't be needing these changes

app/assets/stylesheets/components/_autocomplete.scss
```diff
@@ -0,0 +1,90 @@
+@import url(asset-path("accessible-autocomplete/dist/accessible-autocomplete.min.css"));
```
DilwoarH: Where is this coming from? Can't see a node_module or anything being added?

app/assets/stylesheets/application.scss
DilwoarH: Select2 is used in a few places - we can probably just do `@import "select2"` and `@import "select2-bootstrap"`

app/assets/javascripts/components/autocomplete.js
```diff
@@ -0,0 +1,221 @@
...
+            }) : []
+          )
+        }
+      }
+    })
+
+    $input.parentNode.removeChild($input)
+  }
+
+  Autocomplete.prototype.initAutoCompleteSearchTopics = function () {
```
DilwoarH: Is this used?

app/views/components/_autocomplete.html.erb
ChrisBAshton: Absolutely! I think it's worth doing the work here for now while we figure out what we need from the component (for example, I said in the commit messages I'd like to add an `error_message` parameter rather than use the more verbose `error_items` parameter).

Once we've added to collections-publisher, work on a joined up plan to shift it out to govuk_publishing_components and remove it from Content Publisher and Collections Publisher.

There's talk about retiring the existing [autocomplete component](https://govuk-publishing-components.herokuapp.com/component-guide/accessible_autocomplete) in favour of the one in Content Publisher (which I think is having a11y improvements made to it at the moment...?). We'd need to liaise with @alex-ju on that.

app/assets/javascripts/components/autocomplete.js
ChrisBAshton: Not by us. But I think it's dangerous to start removing code in case we want to move this out to a separate component in govuk_publishing_apps / design system. That's why I've copied over files as-is.

Convo happening here: https://github.com/alphagov/collections-publisher/pull/664#pullrequestreview-268531804

app/assets/stylesheets/components/_autocomplete.scss
ChrisBAshton: I think it must be coming from https://github.com/alphagov/govuk_publishing_components/blob/master/package.json#L16.

We have the govuk_publishing_components gem, which I believe includes all the bundled node_modules. Not quite sure how it all works - Rails magic! It resolves though:

Screen Shot 2019-07-31 at 09 50 42

...and breaks if I change the path.

app/views/components/_autocomplete.html.erb
alex-ju: Yeah, we should chat about how to share this bit of code efficiently - on which we constantly work it seems.

leenagupte:
@ChrisBAshton I've responded to you on slack, but the summary is, yes, please add the missing tests, and yes it's good and ok to re-use the prior art from Content Publisher, but only what's really necessary, and the point that it's needed.

app/models/step_by_step_page.rb
```diff
@@ -90,6 +90,17 @@ def discard_notes
     internal_change_notes.where(edition_number: nil).delete_all
   end
 
+  def issues
```
leenagupte: Is this the same as the `issues` method in `DateParser`? Why are both needed?

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -12,8 +12,13 @@
       text: 'Schedule'
     }
   ]
+  publish_time ||= Time.current.tomorrow.change(hour: 9)
```
leenagupte: Consider renaming this to `default_publish_time` or `placeholder_time` so it's clear whats being populated in the component.

app/assets/stylesheets/application.scss
ChrisBAshton: Done 👍

app/helpers/time_options_helper.rb
ChrisBAshton: Done 👍

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -30,16 +35,30 @@
...
-
-        <%= label_tag "schedule[date][year]", 'Year' %>
-        <%= text_field_tag "schedule[date][year]" %>
-      </div>
+      <%= render "govuk_publishing_components/components/date_input", {
+        legend_text: legend,
+        name: "schedule[date]",
+        hint: "For example, 01 08 2027",
+        id: "date",
+        error_message: @step_by_step_page.issues[:schedule_date],
```
leenagupte: Does adding `error_message` here mean that 

```
<% if @step_by_step_page.errors.any? %>
  <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
<% end %>
```
is no longer needed?

app/views/components/_autocomplete.html.erb
leenagupte: Are there any tests for this component? Can you add some?

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Good spot, `t("schedule.edit.time"),` shouldn't be here.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    it "does not contain 12:00am special case" do
+      expect(subject).not_to include "12:00am"
+    end
+
+    it "returns 48 items (hourly - except 12am special case - and half-hourly)" do
+      expect(subject.count).to eq(48)
+    end
+
+    it "returns all the AM times before the PM times" do
+      index_of_last_am = subject.each_with_index.map { |time, index| time.end_with?("am") ? index : nil  }.compact.last
```
leenagupte: You could use a [select](https://www.shortcutfoo.com/app/dojos/ruby-arrays/cheatsheet) here to just get all of the "am" values.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      index_of_first_pm = subject.index { |time| time.end_with?("pm") }
+      expect(index_of_last_am).to be < index_of_first_pm
+    end
+
+    it "should only contain time values" do
+      subject.each do |time|
+        expect(time).to match /[0-9]{1,2}:[0-9]{2}(am|pm)/
```
leenagupte: Regular Expressions are difficult to read. This is one time where a comment explaining the expected outcomes would be useful.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.describe TimeOptionsHelper do
```
leenagupte: Test class is great! Can you squash it up to the same commit as `TimeOptionsHelper`.

app/views/step_by_step_pages/schedule.html.erb
leenagupte: For example, why is `publish_time` being formatted with `("%-l:%M%P")`? What does that format look like?

app/assets/javascripts/components/autocomplete.js
leenagupte: I thought auto-complete was already a component in `govuk-publishing_components`? If you remove the unneeded stuff here, surely it will be easier to see what optional features need to be added to support Content Publisher?

spec/helpers/time_options_helper_spec.rb
ChrisBAshton: Already done 👍 can I give you a nudge when this is ready to re-review? There's no way to put a GitHub PR back into Draft and I'm still pushing up changes that aren't ready to view yet.

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Have now moved this to the controller 👍

app/assets/javascripts/components/autocomplete.js
ChrisBAshton: As Dilwoar said [here](https://github.com/alphagov/collections-publisher/pull/664#discussion_r309099647) and I said in a commit message, there's an autocomplete component but it doesn't allow you to choose an input value outside of the list of default values (e.g. `10:36am` wouldn't be allowed).

ChrisBAshton: On the basis that we're planning to consolidate all Autocomplete component code into one shared component (in govuk_publishing_components) that can then be pulled into content-publisher and collections-publisher...

Our users (content designers) won't care that there's a bit more code in the page, and our developers should never have to look at the code once they realise it's a straightforward copy of whatever is in content-publisher. The only reason we're looking at the HTML/CSS/JS in the first place is that it hasn't been abstracted away in a pattern library we can simply import it from. (There are almost certainly unused features in the other components we're pulling in, but we just don't see the extra code).

When the day comes that this component has been copied to something central, we'll be able to see that the files in Collections Publisher have only been added in one commit, that the commit references the commit in Content Publisher that they were forked from, and therefore we can safely delete the files and import from govuk_publishing_components without worrying that we've lost any functionality.

If, however, we've chopped the code up, we'll have to look at all of the changes and cross-reference them with the version in govuk_publishing_components - creating more work for our present and future selves. In the spirit of optimising code for deletion, I really think putting work into removing bits of code here would be counter-productive.

spec/helpers/time_options_helper_spec.rb
ChrisBAshton: Have now changed to multiline regex with comments for each part of the regex 👍

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Content Publisher displays the error messages both inside the relevant part of the form and at the top of the page. Feels a little odd to me too but I assume is backed up by user research!

spec/helpers/time_options_helper_spec.rb
ChrisBAshton: That's not quite what this is doing though - I'm getting the position of the last "am" value in the array so that I can make sure it comes before the first "pm" value in the array, thereby ensuring all "am" values are before all "pm" values.

A select would filter the "am"/"pm" array into just an array of "am" values and lose the index information I need - unless I'm misunderstanding your suggestion :)

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -131,11 +131,12 @@ def internal_change_notes
 
   helper_method :placeholder_values
   def placeholder_values
-    publish_time ||= Time.current.tomorrow
+    publish_time ||= Time.current.tomorrow.change(hour: 9)
     {
       day: params.dig(:schedule, :date, :day) || publish_time.day,
       month: params.dig(:schedule, :date, :month) || publish_time.month,
       year: params.dig(:schedule, :date, :year) || publish_time.year,
+      time: params.dig(:schedule, :time) || publish_time.strftime("%-l:%M%P"),
```
ChrisBAshton: Note to self - make a time helper for this, based on the one in content publisher ^

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -58,10 +58,22 @@
...
-      </div>
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: "Time",
+          bold: true,
+        },
+        input: {
+          value: @placeholder_values[:time],
```
ChrisBAshton: Remove `@`.

app/models/step_by_step_page.rb
ChrisBAshton: Have now removed 👍

spec/helpers/time_options_helper_spec.rb
leenagupte: I assumed you could "select" all of the "am"s and then do a count-1 to get the index of the last one.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -129,6 +129,16 @@ def internal_change_notes
     @internal_change_note = InternalChangeNote.new
   end
 
+  helper_method :placeholder_values
+  def placeholder_values
+    publish_time ||= Time.current.tomorrow
+    {
+      day: params.dig(:schedule, :date, :day) || publish_time.day,
+      month: params.dig(:schedule, :date, :month) || publish_time.month,
+      year: params.dig(:schedule, :date, :year) || publish_time.year,
```
ChrisBAshton: Add a test that this `||` behaviour happens.

app/services/datetime_parser.rb
```diff
@@ -24,18 +32,18 @@ def parse
 
 private
 
-  attr_reader :raw_date, :raw_time, :issue_items
+  attr_reader :raw_date, :raw_time, :issue_items, :issue_prefix
```
ChrisBAshton: Kill the issue prefix

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Have now moved this logic out of the view and moved the formatting to a helper method with tests 👍

spec/helpers/time_options_helper_spec.rb
ChrisBAshton: Unfortunately that wouldn't work, imagine:

```ruby
mixed = ["1am", "2am", "9pm", "3am", "7pm"]

index_of_last_am = mixed.select { |time| time.end_with?("am") }.count - 1
# ^ would become ["1am", "2am", "3am"], therefore index = 3-1 = 2
# when actually at index 2 we have "9pm"
```

app/views/components/_autocomplete.html.erb
ChrisBAshton: What kind of tests? Do you mean Ruby integration tests that ensure things like the presence of `error_items` actually renders errors...? (NB, there were no tests in Content Publisher to copy over).

I think JS unit tests would be ideal, but am conscious that it would be scope creep to write tests for features we're not using (such as multi-option select) - but I wouldn't want to start removing features either, for [reasons outlined in the other thread](https://github.com/alphagov/collections-publisher/pull/664#discussion_r309178694). I'm happy to add some tests (once we've clarified which!) but perhaps that work can be more thoroughly done when it's extracted out to govuk_publishing_components.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def parse
+    @issue_items = []
+
+    check_date_is_valid
+    check_time_is_valid
+    return if issue_items.any?
+
+    Time.current.change(
```
leenagupte: I think you can just use `Time.new` here as no specific part of the date/time is being changed.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class DatetimeParser
+  def initialize(date:, time:)
+    @raw_date = date.to_h
+    @raw_time = time.to_s
```
leenagupte: Should this being calling `to_h` as

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DatetimeParser do
+  describe "#parse" do
+    it "returns the parsed date/time when valid" do
```
leenagupte: There are a lot of things being tested here, at it's not immediately clear what being "valid" means. Can you split this test up into smaller units that describe what's valid about the date/time.  e.g.
"it accepts a time that contains a pm"

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+                                  time: "12:00am")
+      expect(parser.parse).to eq expected_time.change(hour: 0)
+
+      parser = DatetimeParser.new(date: date_params,
+                                  time: "6:00 pm")
+      expect(parser.parse).to eq expected_time.change(hour: 18)
+
+      parser = DatetimeParser.new(date: date_params,
+                                  time: "23:32")
+      expect(parser.parse).to eq expected_time.change(hour: 23, min: 32)
```
leenagupte: Is this correct? Shouldn't the time be set to "11:32pm" to be consistent?

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+RSpec.describe DatetimeParser do
+  describe "#parse" do
+    it "returns the parsed date/time when valid" do
+      date_params = { day: "10", month: "01", year: "2019" }
+      expected_time = Time.zone.local(2019, 1, 10, 0, 0)
+
+      parser = DatetimeParser.new(date: date_params,
+                                  time: "11:00am")
+      expect(parser.parse).to eq expected_time.change(hour: 11)
```
leenagupte: Rather than using `.change` can you show what the actual date/time being set is, e.g. `Time.zone.local(2019, 1, 10, 11, 0)`.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -76,9 +76,12 @@ def publish
   def schedule
     set_current_page_as_step_by_step
     if request.post?
-      if @step_by_step_page.update_attributes(scheduled_at: params[:scheduled_at])
+      date_params = params[:schedule][:date].permit(:year, :month, :day)
```
leenagupte: Can you add a before and after controller test to prove that `scheduled_at` is still be stored correctly.

The "before" should be added to the commit before this one and test that the scheduled_at time was being set correctly and that the state was being updated as expected before any of these changes were added.

The "after" should be added with these changes and test that these changes don't affect the `scheduled_at` being saved.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def parse
+    @issue_items = []
+
+    check_date_is_valid
+    check_time_is_valid
+    return if issue_items.any?
```
leenagupte: Should this raise an exception if there are any issues?

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -76,9 +76,12 @@ def publish
...
+      parser = DatetimeParser.new(date: date_params, time: params[:schedule][:time])
+      scheduled_at = parser.parse
```
leenagupte: `parser.parse` returns `nil` if the date is invalid, but `@step_by_step_page.update_attributes(scheduled_at: scheduled_at)` will still be successful, that means a change note will be added for a `nil` date.

app/services/datetime_parser.rb
leenagupte: At this point in time, `issue_items` isn't really doing anything. Perhaps it would be better to extract all of the error handling code into a separate commit?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -148,9 +149,11 @@
     scenario "User tries to schedule publishing for date in the past" do
       given_there_is_a_draft_step_by_step_page
       and_I_visit_the_scheduling_page
+      then_inputs_should_have_tomorrows_date
       and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
       when_I_submit_the_form
       then_I_should_see "Scheduled at can't be in the past"
+      but_inputs_should_retain_the_values_I_entered
```
leenagupte: Can you rename this to `and_.....` rather than `but_....` so that the test steps follow the [Given
When Then](https://martinfowler.com/bliki/GivenWhenThen.html) pattern.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -141,6 +146,15 @@ def placeholder_values
     }
   end
 
+  helper_method :issues_for
+  def issues_for(namespace)
+    if @parser.nil?
+      false
+    else
+      @parser.issues.map { |issue| issue.keys.first == namespace ? { text: issue.values.first } : nil }.compact
```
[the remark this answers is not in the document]
leenagupte: Also, can you update one of tests to check that the error message is as expected.

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -77,9 +77,15 @@
...
-        <%= submit_tag "Schedule to publish", class: "btn btn-primary" %>
-        <%= link_to "Cancel", @step_by_step_page %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full">
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Schedule to publish",
+          } %>
+        </div>
+        <div class="govuk-grid-column-full govuk-!-margin-top-3">
+          <%= link_to "Cancel", step_by_step_page_publish_or_delete_path(@step_by_step_page) %>
```
leenagupte: Why has the destination of the `Cancel` link changed?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -163,6 +163,18 @@
       then_I_should_see "Scheduled to be published at"
       and_there_should_be_no_schedule_button
     end
+
+    scenario "User tries using invalid values for schedule date and time" do
```
leenagupte: It's great that you've added this test, but it's not obvious which bits of commit 0819de3deb733d6f2dbaa8223c71db72b0a7aedb these tests relate to.

Can you go back an split 0819de3deb733d6f2dbaa8223c71db72b0a7aedb into one commit that adds the global error, along with the appropriate lines from this test, and another that adds in the inline error, also with the appropriate lines from this test.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: ['But' is part of the Given When Then pattern](https://cucumber.io/docs/gherkin/reference/#keywords), and makes it clearer that the date & time values should be retained _despite_ being invalid. Happy to rename `values` though 👍

spec/services/datetime_parser_spec.rb
ChrisBAshton: Agreed 👍 was copied directly from content publisher but happy to improve it!

spec/features/managing_step_by_step_pages_spec.rb
leenagupte: Can you also rename "inputs"? 
Personally, I think "and" is an easier read. E.g. "and_I_can_still_see_the_date_and_time_values_I_entered".

spec/helpers/time_options_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it "does not contain 12:00am special case" do
+      expect(subject).not_to include "12:00am"
+    end
+
+    it "returns 48 items (hourly - except 12am special case - and half-hourly)" do
+      expect(subject.count).to eq(48)
+    end
+
+    it "returns all the AM times before the PM times" do
+      index_of_last_am = subject.each_with_index.map { |time, index| time.end_with?("am") ? index : nil }.compact.last
```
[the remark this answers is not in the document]
ChrisBAshton: Smashing! That's much better 👍

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: Good spot - have changed the conditional.

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: I think the URL was originally `@step_by_step_page` when we had 'Scheduled' as its own tab item. We then moved it to a button on publish-or-delete.html but didn't change the URL.

It now feels a bit weird, in that if you click to 'Publish or delete', then 'Schedule to publish', then 'Cancel', you're not taken back to the 'Publish or delete' page.

Good spot - I'll move it into a separate commit.

app/services/datetime_parser.rb
ChrisBAshton: Hmmm... with `Time.current.change`, all tests work as before. But if I change to `Time.new`, then the datetime is saved slightly differently and the `to_s` is formatted differently ("2030-04-20 10:26:00 +0000" instead of "2030-04-20 10:26:00 UTC"), causing an expectation to fail in managing_step_by_step_pages_spec.rb.

Presumably this is because `Time.new` doesn't contain any timezone information, whereas `Time.current` does. I'm inclined to keep it as `Time.current.change` as per Content Publisher, so that it's clear going from one input to separate date/time inputs doesn't change how the value is stored. What do you think?

ChrisBAshton: Can do `Time.new(...).utc` 👍

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: Done 👍

app/controllers/step_by_step_pages_controller.rb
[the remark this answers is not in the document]
ChrisBAshton: Think this is done now 👍

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -143,7 +143,7 @@
       when_I_submit_the_form
       then_I_should_see "has been scheduled to publish"
       and_the_step_by_step_should_have_the_status "Scheduled"
-      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at #{schedule_time}"
```
ChrisBAshton: Why was this removed? 🤔

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -39,6 +39,7 @@
         name: "schedule[date]",
         hint: "For example, 01 08 2027",
         id: "date",
+        error_items: issues_for(:schedule_date),
```
ChrisBAshton: Move this to separate commit

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -55,10 +56,22 @@
...
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: "Time",
+          bold: true,
+        },
+        input: {
+          options: time_options,
+        },
+        error_items: issues_for(:schedule_time),
```
ChrisBAshton: Move this to separate commit

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -55,10 +56,22 @@
...
-      </div>
+      <%= render "components/autocomplete", {
+        id: "time",
+        name: "schedule[time]",
+        label: {
+          text: "Time",
+          bold: true,
+        },
+        input: {
+          options: time_options,
```
ChrisBAshton: Requires `time_options` from other commit

spec/services/datetime_parser_spec.rb
```diff
@@ -93,4 +93,49 @@
       end
     end
   end
+
+  describe "#issues" do
```
ChrisBAshton: Split this into a separate commit

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,7 @@
...
```
ChrisBAshton: This needs moving

spec/helpers/date_time_helper_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
```
ChrisBAshton: This needs moving

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Have now split into two commits 👍

app/views/components/_autocomplete.html.erb
ChrisBAshton: Have added tests later on in this PR, and have stripped down the component so it only has the features we need, as requested 👍 

Will arrange a chat with Alex & Ollie & Dilwoar in the next week or two to see if we can decide where this component should be moved to, as it doesn't make sense for it to be duplicated across repositories.

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -175,6 +198,22 @@ def stub_publishing_api
...
 
+  def stub_scheduling_permissions
+    stub_user.permissions << "Scheduling"
+  end
+
+  def stub_publishing_api_for_scheduling
+    allow(Services.publishing_api).to receive(:get_content)
+      .and_return(content_item(step_by_step_page))
+
+    stub_default_publishing_api_put_intent
```
leenagupte: Stubbing `put_intent` only required for the the scheduling tests, not the unscheduling ones. 
You could follow the pattern of the unscheduling tests and move this line to the `schedule_publishing` method.

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -90,12 +90,35 @@
...
     before :each do
-      stub_user.permissions << "Scheduling"
+      stub_scheduling_permissions
+      stub_publishing_api_for_scheduling
+    end
 
-      allow(Services.publishing_api).to receive(:get_content)
-        .and_return(content_item(step_by_step_page))
+    it "sets `scheduled_at` to a datetime" do
+      expect(step_by_step_page.scheduled_at.class.name).to eq 'NilClass'
```
leenagupte: This line isn't needed, because it's testing the set-up of the factory, and not the action of the controller.

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -90,12 +90,35 @@
...
+
+      schedule_publishing(step_by_step_page)
+
+      expect(step_by_step_page.scheduled_at.class.name).to eq 'Time'
```
leenagupte: Rather than using a class name, it would be better to check the actual value being stored, so that when the inputs for `scheduled_at` are separated into date and time elements later, we can check that we aren't losing any data, e.g. dropping the number of minutes of storing the "am" time rather than the "pm" etc.

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -90,12 +90,35 @@
...
+    end
+
+    it "sets the status to Scheduled" do
+      expect(step_by_step_page.status[:name]).to eq 'draft'
```
leenagupte: As above, you don't need this line.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def check_time_is_valid
+    raise ArgumentError unless parsed_time_values
+  end
+
+  def parsed_time_values
+    @parsed_time_values ||= raw_time.match(%r{
+      \A
+      (?<hour>(2[0-3]|1[0-9]|0?[0-9]))
```
leenagupte: This is nice! 🥇

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def time
+    @time ||= begin
+      hour = parsed_time_values[:hour].to_i
+      period = parsed_time_values[:period]&.downcase
+      minute = parsed_time_values[:minute].to_i
+
+      hour24 = case period
+               when "am" then hour == 12 ? 0 : hour
+               when "pm" then hour == 12 ? hour : hour + 12
```
leenagupte: I'm not sure what these "then" statements are saying. Are you trying to check that the hour value is valid for it's period?

I think it might be easier to read if it's written as a standard if/else instead.

app/services/datetime_parser.rb
ChrisBAshton: I can't take credit for this - I copied it from Content Publisher - but agree it's nice! 😁

app/services/datetime_parser.rb
ChrisBAshton: I can't take credit for this - I copied it from Content Publisher - but agree it can be better written! I'll have a go at rewriting that bit.

spec/services/datetime_parser_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        expect(parser.parse).to be_nil
+      end
+
+      it "returns nil when the hours are invalid" do
+        parser = DatetimeParser.new(date: valid_date, time: "30:00am")
+        expect(parser.parse).to be_nil
+      end
+    end
+
+    context "a valid time is provided" do
```
leenagupte: Nitpick: I think this should be "an invalid date is provided"

spec/controllers/step_by_step_pages_controller_spec.rb
ChrisBAshton: Done 👍

app/services/datetime_parser.rb
ChrisBAshton: Done 👍

spec/services/datetime_parser_spec.rb
ChrisBAshton: Hmmm I see where you're coming from.

Really this is just a way of grouping related tests. Is it worth renaming:

- `context "a valid date is provided"` (line 10) to `describe "valid times"`, and
- `context "a valid time is provided"` (line 83) to `describe "valid dates"`

And extracting the _invalid_ times & dates into their own `describe "invalid times"` and `describe "invalid dates"` blocks?

leenagupte: I might just be easier to remove the contexts and just update the description of the tests.

i.e. 
`it "returns nil when the date is valid but the time is blank"`
`it "returns nil when the date is invalid but the time is valid"`

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -90,12 +90,32 @@
...
     before :each do
-      stub_user.permissions << "Scheduling"
+      stub_scheduling_permissions
+      stub_publishing_api_for_scheduling
+    end
 
-      allow(Services.publishing_api).to receive(:get_content)
-        .and_return(content_item(step_by_step_page))
+    it "sets `scheduled_at` to a datetime" do
+      schedule_publishing(step_by_step_page, "2030-04-20 10:26:51 London")
```
leenagupte: Nitpick: When the date and time inputs are split, users won't be able to add "seconds", so it might be easier not to assign them here, rather than having to remove them from the test later.

app/services/datetime_parser.rb
```diff
@@ -0,0 +1,65 @@
...
+    }ix)
+  end
+
+  def time
+    @time ||= begin
+      hour = parsed_time_values[:hour].to_i
+      period = parsed_time_values[:period]&.downcase
+      minute = parsed_time_values[:minute].to_i
+
+      if period == "am" && hour == 12
```
leenagupte: Nice! Thanks for simplifying this.

spec/views/components/autocomplete_spec.rb
```diff
@@ -0,0 +1,90 @@
+require 'rails_helper'
+
+RSpec.describe "Autocomplete component" do
```
leenagupte: 👍 For adding this test. 
Can you fix it up to the previous commit, where the component was added.
Also, there are already some component tests in `spec/components` can these be moved to the same place for consistency?

spec/views/components/autocomplete_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    assert_empty render_component(label: { text: "foo" })
+  end
+
+  it "should render text input with randomised id when name & label provided" do
+    render_component(name: "foo", label: { text: "bar" })
+
+    assert_select ".app-c-autocomplete", count: 1
+    randomised_for = nil
+    assert_select ".app-c-autocomplete label", count: 1, text: "bar" do |label|
+      randomised_for = label[0]['for']
```
leenagupte: What is this?

spec/views/components/autocomplete_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+    assert_select ".app-c-autocomplete input[name='foo'][id='#{randomised_for}']", count: 1
+  end
+
+  it "should render text input when id, name & label provided" do
+    render_component(name: "foo", label: { text: "bar" }, id: "baz")
```
leenagupte: I think this should be written over multiple lines for readability.

spec/views/components/autocomplete_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  it "renders a default value if provided" do
+    render_component(
+      id: "foo",
+      name: "foo",
+      label: { text: "Countries" },
+      input: {
+        options: ["United Kingdom", "United States"],
+        value: "Spain"
```
leenagupte: The default, which I assume is what `value` is. isn't one of the options, so doesn't that mean the default value isn't being rendered?

ChrisBAshton: Yes. In theory, the placeholder doesn't need to be one of the values in the options array (although it often is in practice).

spec/views/components/autocomplete_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      error_message = error[0].text
+      expect(error_message).to include "Something wrong with A"
+      expect(error_message).to include "Also something wrong with B"
+    end
+  end
+
+  def render_component(arguments)
+    render partial: "components/autocomplete", locals: arguments
+  end
+end
```
leenagupte: It looks like there are quite a few different ways of calling the component to invoke features (e.g. default values). The convention is to add documentation about how the component can be used to a YAML file to `app/views/components/docs`. Can you add some docs for the auto-complete component.

Content Publisher has an example that cane be used for inspiration: https://github.com/alphagov/content-publisher/blob/master/app/views/components/docs/autocomplete.yml

app/services/datetime_parser.rb
```diff
@@ -11,13 +11,13 @@ def parse
       check_date_is_valid
       check_time_is_valid
 
-      Time.new(
+      Time.find_zone("London").local(
```
leenagupte: Can you squash this up to commit 3d3905f60cd40b1b9511009b23aa65651267f439

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -204,9 +204,19 @@ def stub_publishing_api_for_scheduling
       .and_return(content_item(step_by_step_page))
   end
 
-  def schedule_publishing(step_by_step_page, schedule_time = (Time.zone.now + 1.day).to_s)
+  def schedule_publishing(step_by_step_page, schedule = nil)
     stub_default_publishing_api_put_intent
-    post :schedule, params: { step_by_step_page_id: step_by_step_page.id, scheduled_at: schedule_time }
+    if !schedule
```
leenagupte: Rather than adding a conditional here, the convention is to add default parameters to the method definition. 

e.g. `def schedule_publishing(step_by_step_page, year: 1.year.from_now.year, month: 1.year.from_now.month, day: 1.year.from_now.day, time: '9:00am')`

By using named parameters, the calling method can substitute any or all of the default parameters.

An interesting fact is that you don't need to use "tomorrow" to get a year and a day from now, as `from_now` seems to do that anyway.
See:  https://stackoverflow.com/questions/39586380/1-year-from-now-is-also-one-day-ahead?rq=1

leenagupte: As this method is only be used in two places, I wonder if it's worth keeping? I think it makes the tests more complicated. 

`stub_default_publishing_api_put_intent` could just be moved to the `before` block.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -143,7 +141,7 @@
...
+      and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at 2030-04-20 10:26:00 +0100"
```
leenagupte: Could you fix this up to the first commit?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -382,11 +380,18 @@ def when_I_visit_the_scheduling_page
...
-    fill_in 'scheduled_at', with: schedule_time
+    fill_in 'schedule[date][year]', with: "2030"
+    fill_in 'schedule[date][month]', with: "04"
+    fill_in 'schedule[date][day]', with: "20"
+    fill_in 'schedule[time]', with: "10:26am"
   end
 
   def and_I_fill_in_the_scheduling_form_with_a_date_in_the_past
-    fill_in 'scheduled_at', with: '1937-04-20 10:26:51 London'
+    yesterday = Time.current - 1.day
```
leenagupte: Why has the test date changed?

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  def format_time(datetime)
```
leenagupte: Perhaps this could be renamed to `format_time_in_twelve_hour_clock`

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    datetime.strftime("%-l:%M%P")
+  end
+
+  def scheduled_date_placeholder
```
leenagupte: I think if this had a more generic name, e.g. `date_time_placeholder` this helper and the `TimeOptionsHelper` could be merged.

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    publish_time ||= Time.current.tomorrow.change(hour: 9)
```
leenagupte: Helpers are usually generic, so this should probably be renamed to something like `placeholder_time`.

app/helpers/date_time_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    {
+      day: params.dig(:schedule, :date, :day) || publish_time.day,
```
leenagupte: Helpers can be called from any view, so to make this class reusable it would be better to pass in date and time, they can have default values of `nil` so that the conditionals still work.

e.g. `def default_date_time(date=nil, time=nil)`

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -102,7 +102,7 @@
       schedule_publishing(step_by_step_page, year: "2030", month: "04", day: "20", time: "10:26am")
 
       expect(step_by_step_page.scheduled_at.class.name).to eq 'Time'
-      expect(step_by_step_page.scheduled_at.strftime('%d/%m/%Y %H:%M:%S')).to eq '20/04/2030 10:26:00'
+      expect(step_by_step_page.scheduled_at.in_time_zone("London").strftime('%d/%m/%Y %H:%M:%S')).to eq '20/04/2030 10:26:00'
```
leenagupte: Why has this changed here? Can you fix it up?

app/services/datetime_parser.rb
```diff
@@ -6,34 +6,45 @@ def initialize(date:, time:)
     @raw_time = time.to_s
   end
 
+  def issues
```
leenagupte: This method isn't needed as

leenagupte: Is this method being used?

app/services/datetime_parser.rb
```diff
@@ -6,34 +6,45 @@ def initialize(date:, time:)
...
+
+  def issues_for(param)
+    @issue_items.filter { |issue| issue.keys.first == param }.map { |issue| issue.values.first }
+  end
+
   def parse
-    begin
-      check_date_is_valid
-      check_time_is_valid
+    @issue_items = []
```
leenagupte: This should be moved to the initialize method

app/services/datetime_parser.rb
```diff
@@ -6,34 +6,45 @@ def initialize(date:, time:)
     @raw_time = time.to_s
   end
 
+  def issues
+    @issue_items
+  end
+
+  def issues_for(param)
+    @issue_items.filter { |issue| issue.keys.first == param }.map { |issue| issue.values.first }
```
leenagupte: The `@` isn't needed here.

app/services/datetime_parser.rb
```diff
@@ -6,34 +6,45 @@ def initialize(date:, time:)
     @raw_time = time.to_s
   end
 
+  def issues
+    @issue_items
```
leenagupte: I'd consider renaming this variable to `@issues` and then making the attribute accessor public. That way, you won't need to create a "wrapper" method for it as `attr_reader` does that for you.

app/services/datetime_parser.rb
leenagupte: Are you trying to get the value for the first issue of a specific type? Or are you trying to get them all?

If you're want them all, you should be able to do `issue.items.pluck(param)` to get them all and `issue.items.pluck(param).first` to get the first one.

spec/services/datetime_parser_spec.rb
```diff
@@ -92,4 +92,50 @@
       end
     end
   end
+
+  describe "#issues_for" do
+    let(:date_issues) { parser.issues_for(:schedule_date) }
```
leenagupte: `issues_for` is what's being tested, so it feels odd that it's in the "arrange" part of the test, rather than the "act"

spec/services/datetime_parser_spec.rb
```diff
@@ -92,4 +92,50 @@
...
+
+    context "an invalid time is provided" do
+      let(:parser) { DatetimeParser.new(date: valid_date, time: nil) }
+
+      it "finds issues for schedule_time" do
+        parser.parse
+
+        expect(date_issues.count).to eq(0)
+        expect(time_issues.count).to eq(1)
+        expect(time_issues.first).to eq("Enter a valid time")
```
leenagupte: If `issues_for` is returning the first one issue for a type, then you shouldn't need to do `.first` here.

spec/services/datetime_parser_spec.rb
```diff
@@ -92,4 +92,50 @@
...
+    let(:date_issues) { parser.issues_for(:schedule_date) }
+    let(:time_issues) { parser.issues_for(:schedule_time) }
+
+    context "an invalid time is provided" do
+      let(:parser) { DatetimeParser.new(date: valid_date, time: nil) }
+
+      it "finds issues for schedule_time" do
+        parser.parse
+
+        expect(date_issues.count).to eq(0)
```
leenagupte: This line isn't needed as that's not what's being tested. I think this test is supposed to check that the `:schedule_time` issues are being returned, regardless of if there are any `: schedule_date` issues.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -81,6 +81,9 @@ def schedule
       @parser = DatetimeParser.new(date: date_params, time: params[:schedule][:time])
       scheduled_at = @parser.parse
       if @parser.issues.any?
+        @parser.issues.each do |issue|
+          @step_by_step_page.errors.add :base, "Enter a valid #{issue.keys.first.to_s.humanize.downcase}"
```
leenagupte: Could `issue.values` be used here instead?

app/views/step_by_step_pages/schedule.html.erb
```diff
@@ -77,9 +77,15 @@
...
-        <%= submit_tag "Schedule to publish", class: "btn btn-primary" %>
-        <%= link_to "Cancel", @step_by_step_page %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-full">
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Schedule to publish",
+          } %>
+        </div>
+        <div class="govuk-grid-column-full govuk-!-margin-top-3">
+          <%= link_to "Cancel", step_by_step_page(@step_by_step_page), class: "govuk-link" %>
```
leenagupte: What is `step_by_step_page(@step_by_step_page)`?

spec/services/datetime_parser_spec.rb
ChrisBAshton: It returns an array of issues (as that's what's expected by the frontend components by convention). In practice we typically return a 1-element array.

app/views/step_by_step_pages/schedule.html.erb
ChrisBAshton: Whoops! 👍

spec/views/components/autocomplete_spec.rb
ChrisBAshton: The label has a randomised `for` attribute (e.g. `<label for="autocomplete-abc123">`) if no `id` is provided.

It's important that the `for` attribute of the label matches with the `id` attribute of the input so that it is accessible. I could probably write it a bit nicer though 🤔 will have a think...

spec/views/components/autocomplete_spec.rb
ChrisBAshton: Should that be in the same commit? It's going to be a big commit once it's fixed up to the previous one that adds the view/JS/CSS!

app/services/datetime_parser.rb
ChrisBAshton: Whoops 👍

spec/views/components/autocomplete_spec.rb
leenagupte: Yes please.

leenagupte: If the commit gets too big the JS and CSS can be split out.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: I could do, but the `schedule_time` variable is used in two places at that point, so it's nice and DRY.

In this commit, we're no longer using `schedule_time` in `and_I_fill_in_the_scheduling_form`, as the inputs have been split up, so it no longer makes sense to have a variable, hence the hard-coding here.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: No reason, felt this read a bit more clearly (the repetition of `yesterday` hitting home that it's a time in the past. Otherwise `fill_in 'schedule[date][month], with: '04'` could be perfectly applicable for dates in the past _or_ the future.

app/helpers/date_time_helper.rb
ChrisBAshton: That would make it easier to test, too!

spec/features/managing_step_by_step_pages_spec.rb
leenagupte: The "1937" covers that quite well.

app/services/datetime_parser.rb
ChrisBAshton: Don't think that works. `issue_items` at this point looks something like this:

```ruby
issue_items = [
  { schedule_date: "Invalid date" },
  { schedule_date: "Date missing!" },
  { schedule_time: "Invalid time!" }
]
```

app/services/datetime_parser.rb
ChrisBAshton: I thought so too, but if we call `parse` twice, we'd end up doubling the number of issues recorded. Issues should be cleared every time we run `parse`.

spec/services/datetime_parser_spec.rb
ChrisBAshton: Have moved it out 👍

spec/services/datetime_parser_spec.rb
ChrisBAshton: Have removed from this test and created a dedicated test to ensure issues aren't falsely reported.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: It also comes in handy in a later commit, where the `yesterday` variable is used in testing that the user's inputs are retained when they hit submit and there is an error.

app/services/datetime_parser.rb
leenagupte: So what are you trying to do here?

leenagupte: `issue_items.pluck(:schedule_date)` will return `["Invalid date", "Date missing!"]`

spec/controllers/step_by_step_pages_controller_spec.rb
ChrisBAshton: Hmmm I think having this helper method is nicer than duplicating the `stub_default_publishing_api_put_intent`, `post :schedule, params ...` and `step_by_step_page.reload` in two places. Let me know what you think once I've pushed up the latest rebase 👍

app/helpers/date_time_helper.rb
ChrisBAshton: I've opted for `format_time_12_hour_clock`.

app/helpers/date_time_helper.rb
ChrisBAshton: Hmm that becomes a bit of a pain to use in schedule.html:

```erb
        items: [
          {
            name: "day",
            width: 2,
            value: scheduled_date_placeholder(nil, nil, params.dig(:schedule, :date, :day), nil)[:day]
          },
```

I could do this instead:

```erb
        items: [
          {
            name: "day",
            width: 2,
            value: params.dig(:schedule, :date, :day) || scheduled_date_placeholder[:day]
          },
```

(...and remove the `||` logic from `scheduled_date_placeholder`). It's quite a front-end-y bit of logic anyway.

spec/controllers/step_by_step_pages_controller_spec.rb
leenagupte: To be fair, the `post` method is the thing being tested, so it would be nice to be able to clearly see the differences in how it's being called.

`stub_default_publishing_api_put_intent` can be added to the `before` block so it's only called once. 

That only leaves reloading the step by step, and I think it's fine to call that in two places.

app/helpers/date_time_helper.rb
leenagupte: Or you can pass in the whole date.

app/services/datetime_parser.rb
leenagupte: Isn't a new instance of DateTimeParser called on each submit?

app/services/datetime_parser.rb
ChrisBAshton: Hmm I think the issue was `pluck` would return `[ nil ]` if nothing found, as opposed to `[]`. Have now changed to `issue_items.pluck(param).compact` 👍

app/controllers/step_by_step_pages_controller.rb
ChrisBAshton: 🤦‍♂ 

Yes! Good spot. Much better.

spec/views/components/autocomplete_spec.rb
ChrisBAshton: This was done on Friday 👍

app/assets/stylesheets/forms.scss
ChrisBAshton: Have now removed this commit.

spec/views/components/autocomplete_spec.rb
ChrisBAshton: Have rewritten the test name and the variable name, hopefully clearer.

app/services/datetime_parser.rb
ChrisBAshton: It is, though in theory it doesn't have to be.

Have now moved this to the initialize method, tests continue to pass 👍

ChrisBAshton:
Have pushed up a couple of the final tweaks from your snag list, suggest we merge this now unless there are any blockers.

app/views/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,51 @@
...
+  input ||= {}
+  name ||= nil
+  label = label.symbolize_keys if label
+  aria = error_id if error_items.any?
+
+  root_classes = %w(app-c-autocomplete govuk-form-group)
+  root_classes << "app-c-autocomplete--#{width}" if width
+  root_classes << "govuk-form-group--error" if error_items.any?
+
+  if name && label
```
leenagupte: What is this line doing?

ChrisBAshton: Only rendering an input if both `name` and `label` have been provided. (Input has zero value if it doesn't have a name so we can hook into it, and has very poor accessibility if there is no corresponding label).

leenagupte: Sorry, I missed that this is wrapping the whole thing.

leenagupte: This is not a blocker, but I'd consider moving this to it's own block for readability.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it "should only contain time values in 12-hour clock format" do
+      subject.each do |time|
+        expect(time).to match(/
+          1[0-2]     # first portion of time can be two digits, e.g. 10am
+          |          # or
+          [1-9]      # could just be one digit, e.g. 1am
+          :          # colon
+          [0-5][0-9] # second number must be exactly 2 chars, and no more than 59
+          (am|pm)$   # can only end in am or pm
+        /x)
```
leenagupte: I notice that you've removed the `i` option for ignore case. It's made me wonder what happens is the user enters and uppercase  "AM" or "PM"?

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -77,8 +78,10 @@ def publish
   def schedule
     set_current_page_as_step_by_step
     if request.post?
-      date_params = params[:schedule][:date].permit(:year, :month, :day)
-      parser = DatetimeParser.new(date: date_params, time: params[:schedule][:time])
+      date_params = params[:schedule][:date].permit(:year, :month, :day).to_h.symbolize_keys
+      time_param = params[:schedule][:time]
+      @schedule_placeholder = default_datetime_placeholder(date_params.merge(time: time_param))
```
leenagupte: Can `params[:schedule]` be used here? e.g. `default_datetime_placeholder(params[:schedule])`

spec/helpers/time_options_helper_spec.rb
ChrisBAshton: This test tests the output of `time_options` - so the regex isn't running against user input here, it's just running against the array that we've generated (which should be lowercase).

The user input validation happens [here](https://github.com/alphagov/collections-publisher/pull/664/files#diff-23c966e61d21142f65c7c2b9bbaaaca2R58) and does account for down/uppercase.

app/services/datetime_parser.rb
```diff
@@ -30,10 +34,14 @@ def parse
   def check_date_is_valid
     day, month, year = raw_date.values_at(:day, :month, :year)
     Date.strptime("#{day}-#{month}-#{year}", "%d-%m-%Y")
+  rescue ArgumentError
+    issues << { schedule_date: "Enter a valid date" }
```
leenagupte: This is nice!

app/services/datetime_parser.rb
```diff
@@ -30,10 +34,14 @@ def parse
...
   end
 
   def check_time_is_valid
     raise ArgumentError unless parsed_time_values
+  rescue ArgumentError
```
leenagupte: You could combine these lines into one:

```
unless parsed_time_values
  issues << { schedule_time: "Enter a valid time" }
end
```

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -433,8 +450,26 @@ def then_there_should_be_no_schedule_button
...
+                       'form > .app-c-autocomplete'
+                     else
+                       'body'
+                     end
+    within scope_selector do
+      expect(page).to have_content content
+    end
+  end
+
+  def and_I_should_see(content, scope = nil)
```
leenagupte: You could use an `alias_method` here.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -184,6 +184,16 @@
...
     end
+
+    scenario "User tries using invalid values for schedule date and time" do
+      given_there_is_a_draft_step_by_step_page
+      and_I_visit_the_scheduling_page
+      and_I_fill_in_the_scheduling_form_with_nonsense_data
+      when_I_submit_the_form
+      then_I_should_see "Enter a valid date", :within_the_date_component
+      and_I_should_see "Enter a valid time", :within_the_time_component
+      and_the_input_fields_should_be_in_an_error_state
```
leenagupte: I'm not sure if this test is needed as it's covered by the two lines above.

app/services/datetime_parser.rb
leenagupte: I wonder if `schedule_date` should just be `date` and `schedule_time` should just be `time` to make to make this class reusable? 
Or we could borrow from Content Publisher and [pass in an `issue_prefix`](https://github.com/alphagov/content-publisher/blob/2b7c5ddfa72d518064ae50d24c9bb3a1c7a1b3a6/app/services/datetime_parser.rb#L4)

alex-ju:
Frontend looks good for this 👍. Quite a lot of work to do the scheduling overhaul at once, but I get the constrains.

It terms of 'where the autocomplete code comes from', I'd like to flag that `govuk_publishing_components` removed the autocomplete component in 18.0.0, hence the assets you import in this PR won't be available anymore. Since #658 sets up Yarn dependencies - and hoping that would get merged soon - One option is rebasing this and [adding the autocomplete via Yarn](https://github.com/alphagov/content-publisher/blob/master/package.json#L24). Alternatively if this gets merged first we'll need to raise an additional PR to switch from using the autocomplete assets from `govuk_publishing_components` to using it via Yarn. I'm fine either way, but it's good to be aware of this.

Another thing is that, following the guidance, the errors listed in the error_summary component should link to the associated input to ease correction. As this not yet implemented across the app I think it should be addressed separately.

AgaDufrat:
I wanted to suggest having an uniform datetime format. I know you started changing the format before #682, which uses `strftime('%A, %d %B %Y at %l:%M %P')`, was merged.
This could be addressed by changing:
- line 117 in _step_by_step_pages_controller_spec_  
- line 93 in _step_by_step_pages_controller_
- lines 12, 165 in _managing_step_by_step_pages_spec_

But equally happy to do it in a separate PR.

app/views/components/_autocomplete.html.erb
ChrisBAshton: By that, you mean doing...?

```erb
%>
<% if name && label %>
```

Agreed it would made the code more readable, but it also changes behaviour slightly (if name or label not provided, it should render nothing, but would now render "\n"). I could `.strip` in the tests to get it to pass, but it feels more correct to not render anything at all.

app/services/datetime_parser.rb
ChrisBAshton: I removed the `issue_prefix` in the interests of only copying what's needed.
But I agree renaming to `date` and `time` would work and make it more reusable 👍

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -90,7 +90,7 @@ def schedule
         render :schedule
       elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish
-        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{scheduled_at}"
+        note_description = "Minor update scheduled by #{current_user.name} for publishing at #{format_full_date_and_time(scheduled_at)}"
```
leenagupte: Nitpick: I think the note description should say "on" rather than "at", e.g. "on Saturday, 20 April 2030 at 10:26 am"

ChrisBAshton: Agreed. Good spot.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -42,6 +42,12 @@
     end
   end
 
+  describe "#format_full_date_and_time" do
+    it "formats the date and time correctly" do
+      expect(helper.format_full_date_and_time(Time.current.tomorrow.change(hour: 9))).to eq("Wednesday, 14 August 2019 at 9:00 am")
```
leenagupte: I think this test will fail tomorrow when "Time.current.tomorrow" is Thursday.

leenagupte: You could use Timecop to set "today" to be a specific date, and then time travel to it.

ChrisBAshton: Whoops. I wasn't thinking! This worked fine when testing `format_time` but of course only tests the time of day 🤦‍♂

PR #668, opened 2019-07-30
Unschedule publishing of Step by step

AgaDufrat:
**Why**
We have recently added the functionality to schedule publishing #652 and we should enable publishers to unschedule the publication of a step by step after it has been scheduled.

In the event that a scheduled publication date needs to change we should ensure that we have the functionality to unschedule a piece of content that has been scheduled for publication.

**What**
- users can see an option to unschedule a step by step, which is currently scheduled
- user can unschedule a publication of step by step
- `scheduled_at` datetime is cleared
- the status is updated on index page to 'Draft' or to 'Unpublished changes'
- internal change note is created when a step by step is unscheduled
- access to the feature is limited to users with 'Scheduling' permissions

_Note_
We have decided not to implement confirmation dialogue as it's easy to reschedule.

**Example**
unscheduling-gif

[Trello card](https://trello.com/c/QvgBFTkJ/32-allow-publishers-to-unschedule-a-step-by-step-set-for-publication-m)

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -101,7 +101,12 @@ class StepByStepPagesController < ApplicationController
     end
   end
 
-  def unschedule; end
+  def unschedule
+    set_current_page_as_step_by_step
+    @step_by_step_page.update(scheduled_at: nil)
+    unschedule_publishing
+    redirect_to @step_by_step_page, notice: "Publishing of '#{@step_by_step_page.title}' has been unscheduled."
+  end
 
   def revert
     set_current_page_as_step_by_step
@@ -158,6 +163,10 @@ private
       Rails.logger.info "Unpublishing #{@step_by_step_page.content_id} failed"
     end
     @step_by_step_page.mark_as_unpublished
+  end
+
+  def unschedule_publishing
+    StepNavPublisher.cancel_scheduling(@step_by_step_page)
   end
 
   def revert_page
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -2,7 +2,7 @@ class StepByStepPagesController < ApplicationController
   include PublishingApiHelper
 
   before_action :require_gds_editor_permissions!
-  before_action :require_scheduling_permissions!, only: %i[schedule]
+  before_action :require_scheduling_permissions!, only: %i[schedule unschedule]
   before_action :set_step_by_step_page, only: %i[show edit update destroy]
 
   def index
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -101,6 +101,8 @@ class StepByStepPagesController < ApplicationController
     end
   end
 
+  def unschedule; end
+
   def revert
     set_current_page_as_step_by_step
     if request.post?
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -105,6 +105,9 @@ class StepByStepPagesController < ApplicationController
     set_current_page_as_step_by_step
     @step_by_step_page.update(scheduled_at: nil)
     unschedule_publishing
+    note_description = "Publishing was unscheduled by #{current_user.name}."
+    generate_internal_change_note(note_description)
+    set_change_note_version
     redirect_to @step_by_step_page, notice: "Publishing of '#{@step_by_step_page.title}' has been unscheduled."
   end
 
--- a/app/views/step_by_step_pages/publish_or_delete.html.erb
+++ b/app/views/step_by_step_pages/publish_or_delete.html.erb
@@ -32,13 +32,16 @@
           <p>After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
           <%= link_to 'Publish changes', step_by_step_page_publish_path(@step_by_step_page), class: "btn btn-primary publish-button" %>
           <% if user_can_schedule? %>
-            <p>Or</p>	
+            <p>Or</p>
             <% if @step_by_step_page.scheduled_for_publishing? %>
               <div class="alert alert-info">
                 Scheduled to be published at <%= @step_by_step_page.scheduled_at %>
               </div>
+              <%= form_tag step_by_step_page_unschedule_path(@step_by_step_page), method: :unschedule do %>
+                <button class="btn btn-danger">Unschedule publishing</button>
+              <% end %>
             <% else %>
-              <%= link_to 'Schedule publish', step_by_step_page_schedule_path(@step_by_step_page), class: "btn btn-info" %>	
+              <%= link_to 'Schedule publish', step_by_step_page_schedule_path(@step_by_step_page), class: "btn btn-info" %>
             <% end %>
           <% end %>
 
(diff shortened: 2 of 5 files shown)


app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -32,13 +32,16 @@
           <p>After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
           <%= link_to 'Publish changes', step_by_step_page_publish_path(@step_by_step_page), class: "btn btn-primary publish-button" %>
           <% if user_can_schedule? %>
-            <p>Or</p>
```
ChrisBAshton: 😱 how did a trailing whitespace get in there?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -160,6 +160,12 @@
       then_I_should_see "Scheduled to be published at"
       and_there_should_be_no_schedule_button
     end
+
+    scenario "User unschedules publishing" do
+      given_there_is_a_scheduled_step_by_step_page
+      and_I_visit_the_publish_or_delete_page
+      then_I_see_an_unschedule_button
```
ChrisBAshton: Suggest renaming to `then_I_should_see_an_unschedule_button` or `then_there_should_be_an_unschedule_button`.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -403,4 +409,11 @@ def and_there_should_be_a_change_note(change_note)
     visit step_by_step_page_internal_change_notes_path(@step_by_step_page)
     expect(page).to have_content(change_note)
   end
+
+  def then_I_see_an_unschedule_button
+    within(".publish-or-delete") do
```
ChrisBAshton: Any particular reason you're using `within`? I think the test would work without.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -165,6 +166,9 @@
       given_there_is_a_scheduled_step_by_step_page
       and_I_visit_the_publish_or_delete_page
       then_I_see_an_unschedule_button
+      and_I_click_the_unschedule_button
```
ChrisBAshton: Suggest renaming to `when_I_click_the_unschedule_button`

spec/models/step_by_step_page_spec.rb
```diff
@@ -12,7 +12,6 @@
     let!(:step_by_step_page_2) { create(:step_by_step_page, slug: "a", title: "Agonise over the next title you can think of") }
 
     it 'returns step by step pages in alphabetical order by title' do
-      # binding.pry
```
ChrisBAshton: Thanks, this has been bothering me!

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -416,4 +420,12 @@ def then_I_see_an_unschedule_button
       expect(page).to have_css("button", text: "Unschedule publishing")
     end
   end
+
+  def and_I_click_the_unschedule_button
+    click_on 'Unschedule publishing'
+  end
+
+  def base_path
```
leenagupte: This isn't a blocker, but I'm not sure of the benefit of splitting this out.
`stub_publishing_api_destroy_intent(base_path)` is not much shorter than `stub_publishing_api_destroy_intent('/how-to-be-the-amazing-1')`

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -165,6 +166,9 @@
       given_there_is_a_scheduled_step_by_step_page
       and_I_visit_the_publish_or_delete_page
       then_I_see_an_unschedule_button
+      and_I_click_the_unschedule_button
+      then_I_should_see "has been unscheduled"
+      and_the_step_by_step_should_have_the_status "Draft"
```
ChrisBAshton: Is it worth having another test (possibly in `spec/models/step_by_step_page_spec.rb`) that makes sure a step by step has status "Unpublished changes" -> "Scheduled" -> "Unpublished changes" (in case we've accidentally made it change back to "Draft").

spec/models/step_by_step_page_spec.rb
leenagupte: Hmmm, this should have been picked up by the linter. Ideally this would be removed in a separate commit.

spec/models/step_by_step_page_spec.rb
```diff
@@ -280,6 +279,14 @@
       step_by_step_page.mark_draft_updated
       expect(step_by_step_page.scheduled_for_publishing?).to be false
     end
+
+    it 'clears the scheduled status after unscheduling' do
```
leenagupte: This test isn't needed as it's testing the same things as `it 'is not scheduled for publishing if a draft has not been saved'` and `it 'is not scheduled for publishing if scheduled_at is not present'`.

I think the most appropriate place for a test like this would be in a controller test.

spec/features/managing_step_by_step_pages_spec.rb
leenagupte: I think it's a good idea to test that the status is being updated appropriately, but that should be triggered from a controller test, rather than a model test.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -101,6 +101,16 @@ def unpublish
     end
   end
 
+  def unschedule
```
leenagupte: For safety and consistency, this action should be restricted to users who have the "Scheduling" permission.

spec/features/managing_step_by_step_pages_spec.rb
AgaDufrat: I went for this naming convention because there are 8 other methods in this test file with names starting with `then_I_see`.

spec/features/managing_step_by_step_pages_spec.rb
AgaDufrat: Similarly, this is how is was previously done https://github.com/alphagov/collections-publisher/blob/3107e52cbcfce5aa46c90df9793591fd24c8b3aa/spec/features/managing_step_by_step_pages_spec.rb#L301-L315
Also I think it in some cases it could be good future proofing.

spec/features/managing_step_by_step_pages_spec.rb
AgaDufrat: how about `when_I_unschedule_publishing`? This seems to be the way of naming methods with `click_on` action

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: Fair enough. I might go in later and raise a renaming PR to have the English flow better :)

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: 👍

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: I think it was probably added so that Capybara didn't get confused between clicking on the button, or clicking on the tab of the same name. Not sure if that's still the case!

Again, I may refactor this out later across the whole file - looks good here 👍

spec/features/managing_step_by_step_pages_spec.rb
AgaDufrat: Added 2 test to the controller: 
- Draft -> Scheduled -> Draft
- Unpublished changes -> Scheduled -> Unpublished changes

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -101,6 +101,8 @@ def unpublish
     end
   end
 
+  def unschedule; end
```
leenagupte: Thank you for doing this! 🎉 It's great that you added a placeholder so the tests pass, but that all of the functionality isn't added in one giant commit, which makes it easier to review.

PR #670, opened 2019-07-31
Explicitly pass the CSRF token in AJAX requests

cbaines:
This avoids an issue where jQuery modified by jQuery UJS gets
overridden by the unmodified jQuery required by the
govuk_publishing_components. This stops the X-CSRF-Token being sent,
breaking re-ordering list items.

Explictly add the header in, so that this functionality isn't
dependant on jquery-ujs.


DilwoarH:
Should we add tests to make sure this does not happen in the future?

PR #674, opened 2019-08-02
Convert Step by Step admin to Design systems

DilwoarH:
What
=====
Change the step by step admin in collections publisher to use design system and remove the bootstrap css framework.

Why
====
Design system adds additional accessibility needs which is needed to help support users with accessibility needs. 

Background
=========
The bootstrap v3 that is being used will shortly be deprecated and no longer supported.

User Need
========
*As a admin user, I want an accessible user interface so that I can complete my task without assistance.*

Tickets included in this PR
===================
- https://trello.com/c/Qr1UyzNS/57-change-step-edit-page-to-use-the-govuk-design-system
- https://trello.com/c/1LtOVSfK/56-change-step-by-step-overview-page-to-design-systems
- https://trello.com/c/evdkh2xW/61-change-title-and-metadata-page-to-use-the-govuk-design-system
- https://trello.com/c/nNjQU8p8/67-change-reorder-step-by-step-page-to-use-the-govuk-design-system
- https://trello.com/c/g9AliraQ/68-change-secondary-links-page-to-use-the-govuk-design-system
- https://trello.com/c/QkSf7ZI5/59-change-publish-step-by-step-page-to-use-the-govuk-design-system
- https://trello.com/c/SuD4p0Ys/60-change-change-notes-page-to-use-the-govuk-design-system
- https://trello.com/c/HiN1oxZd/55-change-list-step-by-step-page-to-design-systems-to-design-systems

Screenshots
=========
Screen Shot 2019-08-07 at 14 32 29

Screen Shot 2019-08-05 at 10 41 55
Screen Shot 2019-08-05 at 10 42 13
Screen Shot 2019-08-05 at 10 42 41
Screen Shot 2019-08-05 at 10 42 54
Screen Shot 2019-08-05 at 10 43 06
Screen Shot 2019-08-05 at 10 43 20
Screen Shot 2019-08-05 at 10 43 32
Screen Shot 2019-08-05 at 10 44 12
Screen Shot 2019-08-07 at 11 23 40

PR #677, opened 2019-08-08
Show error message when secondary link slug doesn't exist

ChrisBAshton:
Fixes a regression whereby no error message is displayed if a
slug for a step by step's secondary link doesn't exist.
The error wasn't showing as it's outputting to flash[:danger],
which was no longer getting displayed in admin_layout.html.

The regression wasn't caught in tests as the test didn't assert
that the page contained content (it was missing an 'expect()').

Addresses https://trello.com/c/uZhLL67i/99-show-an-error-message-on-secondary-content-page-if-link-does-not-exist.

Screen Shot 2019-08-08 at 15 30 02

It's worth noting that the form error was probably intended to be displayed further down the page:

https://github.com/alphagov/collections-publisher/blob/87cc196cacb9a66ca1377f1b818a80ae2b25124b/app/views/secondary_content_links/index.html.erb#L47

^ That line currently doesn't appear to be doing anything. Worth revisiting in a future card (and if it's following the lead of Content Publisher, probably want to show the error twice - one at the top of the page, one inline in the erroring component). I think this is out of scope for this PR.

### DIFF
--- a/app/controllers/secondary_content_links_controller.rb
+++ b/app/controllers/secondary_content_links_controller.rb
@@ -10,7 +10,7 @@ class SecondaryContentLinksController < ApplicationController
 
       redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id), notice: 'Secondary content was successfully linked.'
     else
-      flash[:danger] = @error if @error
+      flash[:alert] = @error if @error
       redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id)
     end
   end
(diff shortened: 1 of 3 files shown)


app/views/layouts/admin_layout.html.erb
```diff
@@ -55,6 +55,12 @@
         } %>
       <% end %>
 
+      <% if flash["danger"].present? %>
```
leenagupte: Rather than adding this here, the controller could be updated to set `flash[:alert]` instead. That would follow the changes made to use the design system: https://github.com/alphagov/collections-publisher/pull/674/commits/7a072d84ad4f7cda567a06aa425f8de7583fa85d#diff-5f9771f9b5dbe7b3a9da755f7b5dc9bfL100

ChrisBAshton: That was my first approach, then I realised there are lots of instances of `flash[:danger]` elsewhere in the repository, so thought that approach might require changes in lots of places.

But come to think of it, other controllers don't set `layout 'admin_layout'` so it probably isn't an issue that, say, ListItemsController uses `flash[:danger]`. So I'll go back and do as you've suggested 👍

leenagupte: Yeah, the other pages that use `flash[:danger]` are still using the old admin layout.

PR #678, opened 2019-08-08
Lock 'Delete or publish' & 'Secondary links' pages when publishing is scheduled

AgaDufrat:
## What

**'Delete or publish' page**
User _cannot_ publish unpublish, delete a step by step page or discard draft.
User _can_ unschedule.
<kbd>delete_publish_locked</kbd>

**'Secondary links' page**
User _cannot_ add new secondary content links or delete existing secondary content links.
User _can_ view existing secondary content links.
<kbd>secondary_content_locked</kbd>

## Why

[Trello card: Delete or publish](https://trello.com/c/sLY5dXxq/84-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-publish-and-delete-page-s)
[Trello card: Secondary links](https://trello.com/c/pSlr7lT7/87-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-secondary-links-page-s)

### DIFF
--- a/app/views/secondary_content_links/index.html.erb
+++ b/app/views/secondary_content_links/index.html.erb
@@ -40,48 +40,50 @@
 </div>
 
 <div class="govuk-grid-row govuk-!-margin-top-6 govuk-!-margin-bottom-6">
-  <div class="govuk-grid-column-full">
-    <%= form_for(@secondary_content_link, url: step_by_step_page_secondary_content_links_path) do |form| %>
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full">
-          <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
+  <% if !@step_by_step_page.scheduled_for_publishing? %>
+    <div class="govuk-grid-column-full">
+      <%= form_for(@secondary_content_link, url: step_by_step_page_secondary_content_links_path) do |form| %>
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-full">
+            <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
 
-          <%= render "govuk_publishing_components/components/heading", {
-            text: "Add new secondary link",
-            margin_bottom: 5
-          } %>
+            <%= render "govuk_publishing_components/components/heading", {
+              text: "Add new secondary link",
+              margin_bottom: 5
+            } %>
 
-          <%= render "govuk_publishing_components/components/input", {
-            label: {
-              text: "Base Path"
-            },
-            hint: capture do %>
-              For example: 
-              <a href="https://www.gov.uk/pay-vat" target="_blank" class="govuk-link">https://www.gov.uk/pay-vat</a> 
-              or <a href="https://www.gov.uk/bank-holidays" class="govuk-link" target="_blank">/bank-holidays</a>
-            <% end,
-            name: "base_path"
-          } %>
+            <%= render "govuk_publishing_components/components/input", {
+              label: {
+                text: "Base Path"
+              },
+              hint: capture do %>
+                For example:
+                <a href="https://www.gov.uk/pay-vat" target="_blank" class="govuk-link">https://www.gov.uk/pay-vat</a>
+                or <a href="https://www.gov.uk/bank-holidays" class="govuk-link" target="_blank">/bank-holidays</a>
+              <% end,
+              name: "base_path"
+            } %>
+          </div>
         </div>
-      </div>
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full">
-          <%= render "govuk_publishing_components/components/button", {
-            text: "Add secondary link"
-          } %>
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-full">
+            <%= render "govuk_publishing_components/components/button", {
+              text: "Add secondary link"
+            } %>
+          </div>
         </div>
-      </div>
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full govuk-!-margin-top-3">
-          <%= link_to 'Cancel', step_by_step_page_secondary_content_links_path, class: "govuk-link add-secondary-link-cancel--link" %>
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-full govuk-!-margin-top-3">
+            <%= link_to 'Cancel', step_by_step_page_secondary_content_links_path, class: "govuk-link add-secondary-link-cancel--link" %>
+          </div>
         </div>
-      </div>
-    <% end %>
-  </div>
+      <% end %>
+    </div>
+  <% end %>
 </div>
-
-<hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
-
+<% if !@step_by_step_page.scheduled_for_publishing? %>
+  <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
+<% end %>
 
 <div class="govuk-grid-row secondary-links">
   <div class="govuk-grid-column-full">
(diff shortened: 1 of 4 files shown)


app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -28,14 +28,20 @@
...
+            <p class="govuk-body">Before publishing check for broken links.</p>
+            <p class="govuk-body">After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
+            <%= render "govuk_publishing_components/components/button", {
+              text: "Publish changes",
+              href: step_by_step_page_publish_path(@step_by_step_page)
+            } %>
+          <% end %>
           <% if user_can_schedule? %>
-            <p class="govuk-body govuk-!-margin-top-5">Or</p>
+            <!–– when the feature flag is removed, the conditional statement below should be romoved
```
ChrisBAshton: Don't think this comment is necessary.

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -28,14 +28,20 @@
...
+            and the 'Or' should be included in the conditional statement above ––>
+            <% if !@step_by_step_page.scheduled_for_publishing? %>
```
ChrisBAshton: I'd move this to line 39, i.e. `<% if user_can_schedule? && !@step_by_step_page.scheduled_for_publishing? %>`

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -425,4 +426,29 @@ def then_I_see_an_unschedule_button
   def when_I_unschedule_publishing
     click_on 'Unschedule publishing'
   end
+
+  def and_the_step_by_step_is_not_editable
+    when_I_visit_the_publish_or_delete_page
+    there_should_be_no_publish_button
+    there_should_be_no_disgard_changes_button
```
ChrisBAshton: `discard`

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -425,4 +426,29 @@ def then_I_see_an_unschedule_button
   def when_I_unschedule_publishing
     click_on 'Unschedule publishing'
   end
+
+  def and_the_step_by_step_is_not_editable
```
ChrisBAshton: Nice way of consolidating the steps 👍

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -428,12 +428,36 @@ def when_I_unschedule_publishing
...
+    expect(find('tbody')).to have_content(@step_by_step_page.secondary_content_links.first.title)
+  end
+
+  def and_I_cannot_add_secondary_content_link
+    expect(page).to_not have_css("h2", text: "Add new secondary link")
+    expect(page).to_not have_css("button", text: "Add secondary link")
+  end
+
+  def and_I_cannot_delete_secondary_content_links
+    within("tbody") do
```
ChrisBAshton: I'd be tempted to broaden the search by removing the `within` line above ^

This is because if the interface changes in the future (i.e. for the button to be outside of `tbody`), and then we break the code that locks down the page, our tests would continue to pass and we wouldn't know about it.

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -28,14 +28,20 @@
     <% if @step_by_step_page.has_draft? %>
       <div class="govuk-panel panel-default">
         <div class="panel-body">
-          <p class="govuk-body">Before publishing check for broken links.</p>
-          <p class="govuk-body">After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
-          <%= render "govuk_publishing_components/components/button", {
-            text: "Publish changes",
-            href: step_by_step_page_publish_path(@step_by_step_page)
-          } %>
+          <% if !@step_by_step_page.scheduled_for_publishing? %>
```
leenagupte: Looking at this now, I wonder if it might be clearer to instead have methods like: 

* `@step_by_step_page.can_publish?` or 
* `@step_by_step_page.can_unpublish?`
* `@step_by_step_page.can_delete?`

etc, and wrap each action in one of each?

N.B. method names are only an example.

leenagupte: If you do try this, I'd recommend creating these methods first (with unit tests!) and checking that the existing tests still pass, before adding the `scheduled_for_publishing?` checks.

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -28,14 +28,20 @@
...
+            <p class="govuk-body">After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
+            <%= render "govuk_publishing_components/components/button", {
+              text: "Publish changes",
+              href: step_by_step_page_publish_path(@step_by_step_page)
+            } %>
+          <% end %>
           <% if user_can_schedule? %>
-            <p class="govuk-body govuk-!-margin-top-5">Or</p>
+            <!–– when the feature flag is removed, the conditional statement below should be romoved
+            and the 'Or' should be included in the conditional statement above ––>
```
leenagupte: Not sure if you're aware, but HTML comments are still rendered and can be seen when you inspect the page.

app/views/step_by_step_pages/publish_or_delete.html.erb
AgaDufrat: This wouldn't work because there is an if-else statement below, which determines whether Schedule or Unschedule button is shown, so the 'Unschedule publishing' button would never appear. 
I couldn't think of a better way of doing this without messing indentation.

spec/features/managing_step_by_step_pages_spec.rb
leenagupte: typo: should be "discard"

app/views/step_by_step_pages/publish_or_delete.html.erb
AgaDufrat: Sure, I will remove it. I was worried that whoever removes `<% if user_can_schedule? %>` may just do that, so I wanted to draw their attention that there is more to be done. But you are right, unnecessary.

spec/features/managing_step_by_step_pages_spec.rb
leenagupte: It should be possible to traverse each row in the table and check that it doesn't have a "delete" button.

app/views/secondary_content_links/index.html.erb
```diff
@@ -40,48 +40,50 @@
...
+      <% end %>
+    </div>
+  <% end %>
 </div>
-
-<hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
-
+<% if !@step_by_step_page.scheduled_for_publishing? %>
+  <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
+<% end %>
```
leenagupte: Can the conditional above this be expanded to include the section break?

AgaDufrat: This wouldn't work because there is an if-else statement below, which determines whether Schedule or Unschedule button is shown, so the 'Unschedule publishing' button would never appear.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: `within("tbody")` will already catch any "delete" buttons in table rows, as the table rows are inside "tbody". 👍 

Have thought about this a bit more now and this is fine - [`within` will raise an error](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FSession:within) if the scope cannot be found (i.e. if we move away from using tables as markup), so this test shouldn't falsely pass.

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -45,7 +51,7 @@
                   text: "Unschedule publishing",
                   destructive: true
                 } %>
-              <% end %>
+            <% end %>
```
ChrisBAshton: Quite confusing that this `<% end %>` is at the same level of indentation as the `<% else %>` below it! Looks a bit like the `else` on line 55 is tied with the `if` on line 45, which I don't think is the case 🤔

AgaDufrat: oh yes, this is a mistake. This end belongs to the `form_tag`
I will re-do the whole view with Leena's approach, which should make the view more readable.

PR #679, opened 2019-08-08
Unassign the user when step by step is published

leenagupte:
Trello: https://trello.com/c/KnzX5H6C

Change notes are only generated when a draft is saved by a "new user".
[This][1] was added to prevent a change note from being added every time the
user clicks the "save" button.

Unfortunately, this meant that a change note wasn't being added if the same
user who had previously published the step by step made new changes. To solve this
problem the user is being unassigned when the step by step is published.

[1]: https://github.com/alphagov/collections-publisher/blob/master/app/workers/step_by_step_draft_update_worker.rb#L27

### DIFF
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -124,6 +124,16 @@ RSpec.feature "Managing step by step pages" do
     then_I_see_a_page_reverted_success_notice
   end
 
+  scenario "User publishes and then makes more changes to a step by step page" do
+    given_there_is_a_step_by_step_page_assigned_to_me
+    and_I_visit_the_publish_page
+    and_I_publish_the_page
+    then_there_should_be_a_change_note "Minor update published by #{stub_user.name}"
+    when_I_view_the_step_by_step_page
+    and_I_delete_the_first_step
+    then_there_should_be_a_change_note "Draft saved by #{stub_user.name}"
+  end
+
   scenario "User cannot see Schedule button without Scheduling permissions" do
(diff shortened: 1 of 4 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -124,6 +124,16 @@
     then_I_see_a_page_reverted_success_notice
   end
 
+  scenario "User publishes and then makes more changes to a step by step page" do
+    given_there_is_a_step_by_step_page_assigned_to_me
+    and_I_visit_the_publish_page
+    and_I_publish_the_page
+    and_there_should_be_a_change_note "Minor update published by #{stub_user.name}"
```
ChrisBAshton: `then_there_should_be_a_change_note` (line 134 too)

leenagupte: I'll add an alias method and fix up.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -415,6 +425,8 @@ def and_there_should_be_a_change_note(change_note)
     expect(page).to have_content(change_note)
   end
 
+  alias_method :then_there_should_be_a_change_note, :and_there_should_be_a_change_note
```
ChrisBAshton: Didn't know this exists 😱 nice.

leenagupte: @ChrisBAshton The rule of thumb with Ruby on Rails is, if you ever think "wouldn't it be nice if I was able to do x", look it up. You'll probably find that someone had the same thought and has already written it.

PR #685, opened 2019-08-12
Remove cancel links

alex-ju:
Cancel links from navigation rules and secondary contents links do not have effect  at the moment and they should be removed.

These links were used to hide a container and reset the input value of the field. It turns out they weren't useful the way they were designed and after the UI update they don't have any effects anymore so it's been decided to remove them.

[Trello card](https://trello.com/c/Il5JA4d2)

### DIFF
--- a/app/assets/javascripts/steps.js
+++ b/app/assets/javascripts/steps.js
@@ -14,7 +14,6 @@
       this.setOrder(); // this is called so the order of the list is initalised
       this.bindStatusClicks();
       this.bindCancelAddChangeNoteLink();
-      this.bindCancelAddSecondaryLink();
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/steps.js
```diff
@@ -147,14 +147,6 @@
         $('.add-change-note-description--textarea').val('');
         $('.add-change-note--details').removeAttr('open');
       });
-    },
-
-    bindCancelAddSecondaryLink: function() {
```
leenagupte: This is being called from line 17. Can you remove it from there too?

alex-ju: Good spot, updated.

PR #686, opened 2019-08-13
Ensure change notes can be added to scheduled step by steps

ChrisBAshton:
We lock down a lot of behaviour when a step by step is scheduled,
but it should still be possible to add internal change notes.

Trello: https://trello.com/c/i5J47svc/89-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-change-notes-page

### DIFF
--- a/spec/features/managing_steps_spec.rb
+++ b/spec/features/managing_steps_spec.rb
@@ -67,9 +67,16 @@ RSpec.feature "Managing step by step pages" do
     context "and I would like to leave a change note on a step by step page" do
       scenario "User leaves a change note" do
         given_there_is_a_step_by_step_page
-        and_I_visit_the_change_notes_tab
+        when_I_visit_the_change_notes_tab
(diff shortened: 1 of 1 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -498,6 +499,14 @@ def and_the_step_by_step_is_not_editable
     there_should_be_no_unpublish_button
   end
 
+  def but_the_user_can_still_add_change_notes
```
leenagupte: This is an entire "add a change note" test, not a step. The change note test is in `managing_steps_spec` which is probably not the best place for it, but it probably doesn't need to be tested again here. 

I think this step should be split into two:

`when_I_visit_the_change_notes_page`
`then_I_can_add_a_change_note` - This really only needs to check that the add change note buttons are visible.

PR #687, opened 2019-08-13
Improve edit step page

alex-ju:
This PR improves the overall layout of the edit step page and includes the following changes:
- Increases the font size for the markdown example code (we avoid using 14 pixels as it's too small)
- Update the markdown guidance link to use govuk styles
- Make the 'Step title' input label bold, for consistency with the other fields
- Remove arbitrary margin overwrites
- Remove redundant row grid wrapper on primary action/return link
- Correct the back link/breadcrumbs positioning as per [design system template](https://design-system.service.gov.uk/styles/page-template/) and remove main wrapper overwrite.

<table>
<tr>
<td>

edit-step-before

</td>
<td>

edit-step-after

</td>
</tr>
</table>

[Trello card](https://trello.com/c/ousy4kjO)

### DIFF
--- a/app/views/layouts/admin_layout.html.erb
+++ b/app/views/layouts/admin_layout.html.erb
@@ -38,11 +38,11 @@
       app_name: "Collections Publisher",
       phase: "beta"
     } %>
-    <% if yield(:back_link).present? %>
-      <%= render "govuk_publishing_components/components/back_link", href: yield(:back_link) %>
-    <% end %>
 
-    <main class="govuk-main-wrapper" id="main-content" role="main">
+    <%= yield(:back_link) %>
+    <%= yield(:breadcrumbs) %>
+
+    <main class="govuk-main-wrapper<%= " govuk-main-wrapper--l" if yield(:back_link).blank?%>" id="main-content" role="main">
       <% if flash["notice"].present? %>
         <%= render "govuk_publishing_components/components/success_alert", {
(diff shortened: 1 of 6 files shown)


app/views/layouts/admin_layout.html.erb
```diff
@@ -38,11 +38,11 @@
       app_name: "Collections Publisher",
       phase: "beta"
     } %>
-    <% if yield(:back_link).present? %>
-      <%= render "govuk_publishing_components/components/back_link", href: yield(:back_link) %>
```
leenagupte: So you don't need to render the `back_link` component? Is that included somewhere else?

alex-ju: We're already rendering it [in each template file](https://github.com/alphagov/collections-publisher/blob/master/app/views/steps/edit.html.erb#L18). What we are doing before this PR is that we were testing if anything gets passed from the template file to the layout using `content_for` but instead of using that block we were re-creating it in the layout and passed that block as `href`. The reason why this didn't failed is because we're [not using the back_link block, but the breadcrumbs one](https://github.com/alphagov/collections-publisher/blob/master/app/views/steps/edit.html.erb#L17). Let me know if this doesn't make sense, we can discuss it offline.

leenagupte: Thanks for explaining.

PR #688, opened 2019-08-14
Remove unneeded buttons from the markdown toolbar

alex-ju:
According to the guidance steps content should only need injectors for links and bullets.

### Before
Screen Shot 2019-08-14 at 10 18 38

### After
Screen Shot 2019-08-14 at 10 18 51

[Trello card](https://trello.com/c/PiQCXSt0)

### DIFF
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -12,21 +12,9 @@
     <div class="app-c-markdown-editor__head">
       <div class="app-c-markdown-editor__toolbar">
         <markdown-toolbar class="app-c-markdown-editor__toolbar-group" for="<%= textarea[:id] %>">
-          <md-header-2 class="app-c-markdown-editor__toolbar-button" title="Heading level 2" aria-label="Heading level 2">
-            <%= render "components/markdown_editor/heading_two.svg" %>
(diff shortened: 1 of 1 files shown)


app/views/components/_markdown_editor.html.erb
```diff
@@ -12,21 +12,9 @@
...
-          <md-header-3 class="app-c-markdown-editor__toolbar-button" title="Heading level 3" aria-label="Heading level 3">
-            <%= render "components/markdown_editor/heading_three.svg" %>
-          </md-header-3>
           <md-link class="app-c-markdown-editor__toolbar-button" title="Link" aria-label="Link">
             <%= render "components/markdown_editor/link.svg" %>
           </md-link>
-          <md-quote class="app-c-markdown-editor__toolbar-button" title="Blockquote" aria-label="Blockquote">
-            <%= render "components/markdown_editor/blockquote.svg" %>
-          </md-quote>
-          <md-ordered-list class="app-c-markdown-editor__toolbar-button" title="Numbered list" aria-label="Numbered list">
```
leenagupte: Numbered lists aren't needed either?

alex-ju: It seems like not

PR #689, opened 2019-08-14
Hide Reorder steps tab when the step by step is scheduled

AgaDufrat:
## What
- User _cannot_ see the Reorder steps tab and therefore change the order
- User _can_ still see the order in the steps Overview page

**Not scheduled**
Screen Shot 2019-08-14 at 10 53 32

**Scheduled for publishing**
Screen Shot 2019-08-14 at 10 53 12

_Note:_ No extra check implemented in the controller, as we trust content designers not to try to hack the url. 

## Why

[Trello card](https://trello.com/c/E4uqOWRy/82-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-reorder-page-s)

### DIFF
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -487,6 +487,8 @@ RSpec.feature "Managing step by step pages" do
   end
 
   def and_the_step_by_step_is_not_editable
+    then_there_should_be_no_reorder_steps_tab
+
     when_I_visit_the_secondary_content_page
     then_I_can_see_the_existing_secondary_links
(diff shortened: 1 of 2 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -487,6 +487,8 @@ def when_I_unschedule_publishing
   end
 
   def and_the_step_by_step_is_not_editable
+    there_should_be_no_reorder_steps_tab
```
leenagupte: I think this should be renamed to `then...` as "there" isn't one of the step keywords for cucumber tests: https://cucumber.io/docs/gherkin/reference/#keywords

PR #690, opened 2019-08-15
Show information instead of inputs if step by step can not be edited

alex-ju:
### What

When a user has scheduled a step by step for publishing, they should not be able to make any further changes to that step by step but they should still be allowed to check for broken links and view step content.

### Why
This prevent any unexpected edits being published too early.

[Trello card](https://trello.com/c/YMFbkQ13)

### DIFF
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -489,6 +489,10 @@ RSpec.feature "Managing step by step pages" do
   def and_the_step_by_step_is_not_editable
     then_there_should_be_no_reorder_steps_tab
 
+    when_I_edit_the_step_by_step_page
+    then_I_can_see_the_step_by_step_details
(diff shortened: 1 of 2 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -489,6 +489,10 @@ def when_I_unschedule_publishing
   def and_the_step_by_step_is_not_editable
     then_there_should_be_no_reorder_steps_tab
 
+    when_I_edit_the_step_by_step_page
+    then_I_can_see_the_step_by_step_details
+    and_I_cannot_the_step_by_step_details
```
leenagupte: Typo: I think this test name is missing a word. Perhaps it should be `and_I_cannot_edit_the_step_by_step_details`?

alex-ju: 🤦‍♂ It's missing the verb. Updated now

PR #691, opened 2019-08-15
Remove edit options from overview page for scheduled step by steps

leenagupte:
Trello: https://trello.com/c/ltbNPIGz

## Motivation

When a user has scheduled a step by step for publishing, they should not be able to make any further changes to that step by step

BUT they should still be allowed to add internal changenotes, check for broken links and preview changes.

### Before
Screen Shot 2019-08-15 at 13 52 19

### After
Screen Shot 2019-08-15 at 13 52 44

### DIFF
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -489,6 +489,9 @@ RSpec.feature "Managing step by step pages" do
   def and_the_step_by_step_is_not_editable
     then_there_should_be_no_reorder_steps_tab
 
+    when_I_view_the_step_by_step_page
+    then_I_can_see_the_steps
+
     when_I_edit_the_step_by_step_page
     then_I_can_see_the_step_by_step_details
     and_I_cannot_edit_the_step_by_step_details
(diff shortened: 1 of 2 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -562,4 +564,8 @@ def there_should_be_no_unpublish_button
       expect(page).to_not have_css("button", text: "Unpublish step by step")
     end
   end
+
+  def then_I_can_preview_the_step_by_step
+    expect(page).to_not have_button("Preview")
```
AgaDufrat: I'm a bit confused why we are expecting the page `to_not` have the "Preview" button if we want to be able to preview?

leenagupte: Good spot!

AgaDufrat: Why did the test pass though?

leenagupte: I'm wondering that too.

leenagupte: I think it's because "Preview" is classed as a "link" not a "button"

When I changed the test to `expect(page).to have_button("Preview")`, it failed, but `expect(page).to have_link("Preview")` works.

leenagupte: Will update and fix up.

PR #693, opened 2019-08-16
Lock 'Choose navigation' page when the step by step is scheduled 

AgaDufrat:
## What
- User _cannot_ change the navigation setting
- User _can_ view the navigation setting

**Not scheduled**
not_scheduled

**Scheduled for publishing**
locked

## Why

[Trello card](https://trello.com/c/0yrygTBl/88-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-choose-navigation-pages)

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -54,6 +54,14 @@ FactoryBot.define do
     end
   end
 
+  factory :step_by_step_page_with_secondary_content_and_navigation_rules, parent: :step_by_step_page_with_steps do
+    after(:create) do |step_by_step_page|
+      create(:secondary_content_link, step_by_step_page: step_by_step_page)
+      create(:navigation_rule, step_by_step_page: step_by_step_page)
+      create(:navigation_rule, step_by_step_page: step_by_step_page, title: "Also good stuff", base_path: "/also/good/stuff")
+    end
+  end
+
   factory :step do
     title { "Check how awesome you are" }
     logic { "number" }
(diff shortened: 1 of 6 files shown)


spec/models/navigation_rule_spec.rb
```diff
@@ -139,4 +139,91 @@
...
+          {
+            text: "Never show navigation",
+            value: "never",
+            selected: false
+          }
+        ]
+        expect(resource.options_with_selected).to eq(expected_options)
+      end
+    end
+    context "include_in_links is set to conditionally" do
```
leenagupte: Minor: Can you add a line break here?

spec/factories.rb
```diff
@@ -54,6 +54,14 @@
     end
   end
 
+  factory :step_by_step_page_with_secondary_content_and_navigation_rules, parent: :step_by_step_page_with_steps do
```
leenagupte: I'm not why this factory has been added? Could the existing `step_by_step_page_with_navigation_rules` be used instead?

spec/models/navigation_rule_spec.rb
```diff
@@ -239,7 +239,7 @@
       )
       expect(resource.selected_option_text).to eq("Always show navigation")
     end
-
+
```
leenagupte: Can you fix this up?

spec/factories.rb
AgaDufrat: `step_by_step_page_with_navigation_rules` doesn't have secondary content links, which we also need in this test

PR #694, opened 2019-08-16
Update change note history

alex-ju:
This PR:

- Updates the change notes layout to use a  2/3 column for the internal notes timeline
- Make internal notes permanently visible
- Replaces the accordion with a [timeline entry pattern used consistently with content-publisher](https://github.com/alphagov/content-publisher/blob/60fa3c2fd2b351b2ca3ad5d1bd39f3ea1527918a/app/views/documents/show/_history_tab.html.erb)
- Use the accordion at the edition level (rather then internal note entry level).

<table>
<tr><td valign="top">

internal-notes-before

</td><td valign="top">

collections-publisher dev gov uk_step-by-step-pages_1_internal-change-notes copy

</td></tr>
</table>

**Note**: I appreciate the history of this PR doesn't make much sense as there was a bit of a back and forth on it, so I'd suggesting file-based review and I can squash them before merging.

[Trello card](https://trello.com/c/pmUvssRI)

### DIFF
--- a/app/views/step_by_step_pages/internal_change_notes.html.erb
+++ b/app/views/step_by_step_pages/internal_change_notes.html.erb
@@ -25,56 +25,45 @@
 
 <%= render 'nav', step_by_step_page: @step_by_step_page, active: "internal-change-notes" %>
 
-<%= form_for(@internal_change_note, url: step_by_step_page_internal_change_notes_path) do |form| %>
-  <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
-
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
-      <details class="add-change-note--details">
-        <summary class="add-change-note--summary">
-          <span class="gem-c-button govuk-button">Add a change note</span>
-        </summary>
-        <div class="govuk-grid-column-full">
-          <%= render "govuk_publishing_components/components/textarea", {
-            label: {
-              text: "Description"
-            },
-            name: "internal_change_note[description]",
-            rows: 8
-          } %>
-        </div>
-        <div class="govuk-grid-column-full">
-          <%= render "govuk_publishing_components/components/button", {
-            text: "Save change note"
-          } %>
-        </div>
-        <div class="govuk-grid-column-full govuk-!-margin-top-3">
-          <%= link_to 'Cancel', step_by_step_page_internal_change_notes_path, class: "govuk-link add-change-note-cancel--link" %>
-        </div>
-      </details>
-    </div>
-  </div>
-<% end %>
-
+<% displayed_edition_number = nil %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/accordion", {
-      items: @step_by_step_page.internal_change_notes.map.with_index do |note, index|
-        {
-          heading: {
-            text: "#{note.readable_created_date}"
-          },
-          summary: {
-            text: sanitize("#{note.edition_number ? "Edition #{note.edition_number}" : "Current version"} - Change made by <strong>#{note.author}</strong>")
-          },
-          content: {
-            html: sanitize("
-              <pre class=\"govuk-body change-note--description\">#{note.description}</pre>
-            ")
-          },
-          expanded: index == 0
-        }
-      end
-    } %>
+    <%= form_for(@internal_change_note, url: step_by_step_page_internal_change_notes_path) do |form| %>
+      <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
+
+      <%= render "govuk_publishing_components/components/textarea", {
+        label: {
+          text: "Internal note",
+          bold: true
+        },
+        name: "internal_change_note[description]"
+      } %>
+
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Add internal note",
+        margin_bottom: true
+      } %>
+    <% end %>
+
+    <% @step_by_step_page.internal_change_notes.each do |entry| %>
+      <div class="app-c-timeline-entry">
+        <% if displayed_edition_number != entry.edition_number %>
+          <h3 class="app-c-timeline-entry__title"><%= "#{entry.edition_number.to_i.ordinalize} edition" %></h3>
+          <% displayed_edition_number = entry.edition_number %>
+        <% end %>
+
+        <h4 class="app-c-timeline-entry__type">
+          <%= "Internal note" %>
+        </h4>
+
+        <p class="app-c-timeline-entry__dateline">
+          <%= entry.readable_created_date %> by <%= entry.author %>
+        </p>
+
+        <div class="app-c-timeline-entry__description">
+          <%= entry.description %>
+        </div>
+      </div>
+    <% end %>
   </div>
 </div>
--- a/app/views/step_by_step_pages/internal_change_notes.html.erb
+++ b/app/views/step_by_step_pages/internal_change_notes.html.erb
@@ -25,7 +25,6 @@
 
 <%= render 'nav', step_by_step_page: @step_by_step_page, active: "internal-change-notes" %>
 
-<% displayed_edition_number = nil %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_for(@internal_change_note, url: step_by_step_page_internal_change_notes_path) do |form| %>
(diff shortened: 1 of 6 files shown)


app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -25,56 +25,46 @@
...
-            text: sanitize("#{note.edition_number ? "Edition #{note.edition_number}" : "Current version"} - Change made by <strong>#{note.author}</strong>")
-          },
-          content: {
-            html: sanitize("
-              <pre class=\"govuk-body change-note--description\">#{note.description}</pre>
-            ")
-          },
-          expanded: index == 0
+      items: @step_by_step_page.internal_change_notes
+        .group_by { |edition| edition.edition_number }
```
leenagupte: I think it's fine to use `group_by` here 👍

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -25,56 +25,46 @@
...
+        .map { |edition_number, records|
```
leenagupte: I don't think the `edition_number` being defined here is being used anywhere.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -25,56 +25,46 @@
...
+              text: sanitize("#{records.first.edition_number ? "#{records.first.edition_number.to_i.ordinalize} edition" : "Current version"}")
+            },
+            content: {
+              html: sanitize(records.map { |note|
+                  "<div class='app-c-timeline-entry'>
+                    <h4 class='app-c-timeline-entry__type'>Internal note</h4>
+                    <p class='app-c-timeline-entry__dateline'>#{note.readable_created_date} by #{note.author}</p>
+                    <pre class='app-c-timeline-entry__description'>#{note.description}</pre>
+                  </div>"
+                }.join()
```
leenagupte: What is the `join` doing?

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -25,56 +25,46 @@
...
+              )
+            },
+            expanded: !records.first.edition_number
```
leenagupte: This reads a bit strangely, but it seems to work.

spec/features/managing_steps_spec.rb
```diff
@@ -203,12 +203,12 @@ def and_I_see_the_steps_updated_on_the_step_by_step_details_page
   end
 
   def and_I_write_a_change_note
-    fill_in "Description", with: "I've changed this step by step!"
+    fill_in "Internal note", with: "I've changed this step by step!"
```
leenagupte: It feels like we're missing tests about how change notes are displayed.

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: `group_by` returns a hash where the key is the `edition_number` which is why we need to reflect both arguments on the `map` method.

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: we're using `join` to concatenate/merge an array of strings into a single string to be passed as content for the accordion item. `flat_map` would flatten the array but we'll still need to merge the elements.

spec/features/managing_steps_spec.rb
alex-ju: Added tests to capture the new internal notes' structure on `then_the_change_note_should_be_saved`. After we check for the success message we're now also checking that the history timeline was updated and contains the newly inserted note.

alex-ju:
Thanks for the review @leenagupte. Tried to address the comments.

> When you sort out the commits, could you add a reason for the changes? e.g. `bindCancelAddChangeNoteLink` isn't needed because the new change note text area is always being displayed now etc.

Added more details on 'why' for each commit message.

> At the moment we don't record state changes to step by steps in same way that content-publisher does, but I wonder if in a future enhancement we could change InternalChangeNote to accept a "change_type" and use that to populate the heading, for example, "Draft saved", "Published" and "Internal note" could be the default heading for hand craft notes.

Yes, having a `change_type` sounds like a great idea. This is the reason I kept the type in the title - even though at the moment there's only one.

app/views/step_by_step_pages/internal_change_notes.html.erb
leenagupte: You can use `_` to ignore a return value if it's not used.

alex-ju: Nice! Updated.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -24,26 +24,30 @@
...
-              <pre class=\"govuk-body change-note--description\">#{note.description}</pre>
-            ")
-          },
-          expanded: index == 0
-        }
-      end
-    } %>
+    <% @step_by_step_page.internal_change_notes.each do |entry| %>
+      <div class="app-c-timeline-entry">
+        <% if displayed_edition_number != entry.edition_number %>
```
leenagupte: What happens here if `entry.edition_number` is nil? Originally it displayed `Current version`.

Is the check needed? Will `displayed_edition_number` ever be the same as `entry.edition_number` at this point in the loop?

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -24,26 +24,30 @@
...
+          <h3 class="app-c-timeline-entry__title"><%= "#{entry.edition_number.to_i.ordinalize} edition" %></h3>
```
leenagupte: What happens here if `entry.edition_number` is nil? Originally it displayed `Current version`.

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: Not anymore, I've removed it in one of the following commits when we changed the approach.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -28,6 +28,23 @@
 <% displayed_edition_number = nil %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <%= form_for(@internal_change_note, url: step_by_step_page_internal_change_notes_path) do |form| %>
```
leenagupte: Can this change be fixed the commit [Move internal notes to aside 1/3"](3dfb1c621d2613b0c12fde25b3b641d39f453485)

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: This is also not present in the final changes. It looks like this after the last commit `"#{records.first.edition_number ? "#{records.first.edition_number.to_i.ordinalize} edition" : "Current version"}"`

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: Yep, will merge a few commits as we pivoted on the approach.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -45,25 +44,27 @@
...
+              text: sanitize("#{records.first.edition_number ? "#{records.first.edition_number.to_i.ordinalize} edition" : "Current version"}")
+            },
+            content: {
+              html: sanitize(records.map { |note|
+                  "<div class='app-c-timeline-entry'>
+                    <h4 class='app-c-timeline-entry__type'>Internal note</h4>
+                    <p class='app-c-timeline-entry__dateline'>#{note.readable_created_date} by #{note.author}</p>
+                    <pre class='app-c-timeline-entry__description'>#{note.description}</pre>
+                  </div>"
+                }.join()
```
leenagupte: Nitpick: In Ruby you don't need the empty parentheses. You can just call `.join`.

app/views/step_by_step_pages/internal_change_notes.html.erb
leenagupte: Could you remove it from the history?

alex-ju:
Thanks for the review @leenagupte. Cleaned the commit history from the pivoting approach. It's only 3 clear commits now. Updated with your suggestions too.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -45,25 +44,27 @@
       } %>
     <% end %>
 
-    <% @step_by_step_page.internal_change_notes.each do |entry| %>
-      <div class="app-c-timeline-entry">
-        <% if displayed_edition_number != entry.edition_number %>
```
leenagupte: As this check always evaluated to true, and therefore was never needed, can you fix up it's removal to the commit it was added in.

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -25,56 +25,45 @@
...
-        </div>
-      </details>
-    </div>
-  </div>
-<% end %>
-
+<% displayed_edition_number = nil %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/accordion", {
```
leenagupte: Could you add a line to the commit message that clarifies that the accordion is being removed to make it easier to change the styling, but that it will be re-added in the next commit? I assume that's the reason the accordion is being removed here.

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: It is needed in this commit and it's doesn't always evaluate to true. It [starts with nil](https://github.com/alphagov/collections-publisher/pull/694/commits/c3105a241af4d66e663f6bc23c33c14c8d74abff#diff-5a189794e316cb715d3507edf2b23b8fR28) then [gets updated every time we reach a new edition](https://github.com/alphagov/collections-publisher/pull/694/commits/c3105a241af4d66e663f6bc23c33c14c8d74abff#diff-5a189794e316cb715d3507edf2b23b8fR52). We're using [a similar technique in content-publisher](https://github.com/alphagov/content-publisher/blob/master/app/views/documents/show/_history_tab.html.erb#L7).

app/views/step_by_step_pages/internal_change_notes.html.erb
leenagupte: To answer the original question, at this point in the code, the history looks like:

Screen Shot 2019-08-21 at 11 43 10

Which is fine, but it's not clear that the first title is being lost for step by steps with a draft from the commit description.

app/views/step_by_step_pages/internal_change_notes.html.erb
leenagupte: I don't get how this working, but I see that it does hide the title, there by "grouping" the change notes. so 👍

alex-ju:
@leenagupte updated commit messages as requested

PR #695, opened 2019-08-16
Lock 'Choose navigation' page when the step by step is scheduled

AgaDufrat:
## What
- User _cannot_ change the navigation setting
- User _can_ view the navigation setting

**Not scheduled**
not_scheduled

**Scheduled for publishing**
locked

## Why

[Trello card](https://trello.com/c/0yrygTBl/88-lock-a-step-by-step-that-has-been-scheduled-for-publication-on-the-choose-navigation-pages)

### DIFF
--- a/spec/models/navigation_rule_spec.rb
+++ b/spec/models/navigation_rule_spec.rb
@@ -139,4 +139,25 @@ RSpec.describe NavigationRule do
       expect(resource.smartanswer?).to be false
     end
   end
+
+  describe '#display_text' do
+    let(:resource) {
+      described_class.new(
+        title: 'A Title',
+        base_path: '/a-base-path',
+        content_id: 'A-CONTENT-ID-BOOM',
+      )
+    }
+    it "returns 'Always show navigation' if the value always" do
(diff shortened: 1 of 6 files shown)


spec/models/navigation_rule_spec.rb
```diff
@@ -139,4 +139,25 @@
       expect(resource.smartanswer?).to be false
     end
   end
+
+  describe '#display_text' do
+    let(:resource) {
```
leenagupte: Not a blocker: the `step_by_step_page` defined at the top of the test could be reused here rather than creating new object. 

e.g. `expect(step_by_step_page.display_text("always")).to eq("Always show navigation")`

AgaDufrat: But I'm creating a `navigation_rule` here and I'm reusing the `step_by_step_page` created at the top. I must say that the name `resource` is not very descriptive but that's what is used in all the other tests.

leenagupte: Sorry! I misread that. 🤦‍♂  You're right, `resource` is badly named.

PR #696, opened 2019-08-16
Set the time zone to London

leenagupte:
Trello: https://trello.com/c/WrpUNtsn

Config taken from: https://github.com/alphagov/content-publisher/commit/4ae1c4ed683bf0b24eb11d135aebd3cdd1a8ffa6

We store dates in the database in UTC, which means that when we're displaying
dates and times in the views they're an hour out of date during daylight saving
time. This sets the app's time zone to 'London' so that ActiveRecord
automatically converts dates and times to the UK's current time zone, whether
it's GMT or BST.

Also updates a test that was trying to get around this problem by converting the dates checked to "UTC +00:00"

### Before
Screen Shot 2019-08-16 at 18 24 54
Screen Shot 2019-08-16 at 18 25 08

### After
Screen Shot 2019-08-16 at 18 34 01
Screen Shot 2019-08-16 at 18 34 25

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -19,6 +19,8 @@ module CollectionsPublisher
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    config.time_zone = "London"
+
     unless Rails.env.production?
       ENV['JWT_AUTH_SECRET'] = '123'
     end
--- a/spec/models/step_spec.rb
+++ b/spec/models/step_spec.rb
@@ -131,8 +131,8 @@ RSpec.describe Step do
     end
 
     it 'should return the last date the links where checked' do
-      create(:link_report, batch_id: 2, step_id: step_item.id, created_at: "2018-08-07 10:30:38")
-      expect(step_item.links_last_checked_date.utc).to eq(Time.new(2018, 8, 7, 10, 30, 38).utc)
+      create(:link_report, batch_id: 2, step_id: step_item.id, created_at: Time.zone.parse("2018-08-07 10:30:38"))
+      expect(step_item.links_last_checked_date).to eq(Time.zone.local(2018, 8, 7, 10, 30, 38))
     end
 
     it 'should not fail if the saved batch id does not match a batch in link-checker-api' do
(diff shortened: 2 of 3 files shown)


config/application.rb
```diff
@@ -19,6 +19,8 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    config.time_zone = "London"
```
erino: What about `config.time_zone = "UTC"`? as won't London change between `GMT` and `BST`

spec/models/step_spec.rb
```diff
@@ -131,8 +131,8 @@
     end
 
     it 'should return the last date the links where checked' do
-      create(:link_report, batch_id: 2, step_id: step_item.id, created_at: "2018-08-07 10:30:38")
-      expect(step_item.links_last_checked_date.utc).to eq(Time.new(2018, 8, 7, 10, 30, 38).utc)
+      create(:link_report, batch_id: 2, step_id: step_item.id, created_at: Time.new(2018, 8, 7, 10, 30, 38))
```
ChrisBAshton: This line probably doesn't need changing.

config/application.rb
ChrisBAshton: That's what we're trying to achieve 👍 The current time in UTC is 1 hour behind BST, so if someone scheduled a publish for 10am on Tuesday it would actually be 9am UTC, which isn't the behaviour they want.

Setting to "London" accounts for both GMT and BST, so can schedule times predictably.

spec/models/step_spec.rb
```diff
@@ -131,8 +131,8 @@
...
+      expect(step_item.links_last_checked_date).to eq(Time.new(2018, 8, 7, 10, 30, 38))
```
kevindew: Note, this time will be in UTC (or whatever timezone the host OS is in). If you do `Time.zone.parse("2018-08-07 10:30:38")` you'll get it in London Time

leenagupte:
@AgaDufrat You're right, the line `in_time_zone` is no longer needed in the test. I've removed it and fixed up.

leenagupte:
@kevindew I've updated the test to use `Time.zone.parse` as recommended. `Time.new` was using `UTC` so I've changed the expectation as well.

PR #700, opened 2019-08-20
Check steps have content before publishing

leenagupte:
Trello: https://trello.com/c/3qZiI22D

## What
Hides the "Publish" and "Schedule Publish" buttons if there are no steps, and if any steps don't have content.

## Why
Step by steps will still be able to be saved and previewed, but hiding the buttons will prevent any incomplete step by steps from being accidentally published.

### Step by step with unpublished changes
Screen Shot 2019-08-20 at 18 10 56

### Draft step by step
Screen Shot 2019-08-20 at 18 10 36

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -64,7 +64,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def can_be_published?
-    has_draft? && !scheduled_for_publishing? && steps.any?
+    has_draft? && !scheduled_for_publishing? && steps_have_content?
   end
 
   def can_be_unpublished?
@@ -147,5 +147,9 @@ private
       text: "Unpublished changes",
       label_class: "label-primary"
     }
+  end
+
+  def steps_have_content?
+    steps.any? && steps.map(&:contents).all?(&:present?)
   end
 end
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -83,6 +83,10 @@ class StepByStepPage < ApplicationRecord
     !scheduled_for_publishing?
   end
 
+  def steps_have_content?
+    steps.any? && steps.map(&:contents).all?(&:present?)
+  end
+
   # Create a deterministic, but unique token that will be used to give one-time
   # access to a piece of draft content.
   # This token is created by using an id that should be unique so that there is
@@ -147,9 +151,5 @@ private
       text: "Unpublished changes",
       label_class: "label-primary"
     }
-  end
-
-  def steps_have_content?
-    steps.any? && steps.map(&:contents).all?(&:present?)
   end
 end
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -64,7 +64,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def can_be_published?
-    has_draft? && !scheduled_for_publishing?
+    has_draft? && !scheduled_for_publishing? && steps.any?
   end
 
   def can_be_unpublished?
--- a/app/views/step_by_step_pages/publish_or_delete.html.erb
+++ b/app/views/step_by_step_pages/publish_or_delete.html.erb
@@ -28,15 +28,6 @@
     <% if @step_by_step_page.can_be_published? || @step_by_step_page.can_discard_changes? ||  @step_by_step_page.scheduled_for_publishing?%>
       <div class="govuk-panel panel-default">
         <div class="panel-body">
-          <% if @step_by_step_page.can_be_published? %>
-            <p class="govuk-body">Before publishing check for broken links.</p>
-            <p class="govuk-body">After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
-            <%= render "govuk_publishing_components/components/button", {
-              text: "Publish changes",
-              href: step_by_step_page_publish_path(@step_by_step_page)
-            } %>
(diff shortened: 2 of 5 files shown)


spec/models/step_by_step_page_spec.rb
```diff
@@ -308,6 +308,14 @@
 
       expect(step_by_step_page.can_be_published?).to be false
     end
+
+    it 'cannot be published if there are no steps' do
+      step_by_step_page = create(:step_by_step_page, slug: "no-steps")
+      step_by_step_page.mark_draft_updated
+
+      expect(step_by_step_page.can_be_published?).to be false
+      expect(step_by_step_page.can_be_published?).to be false
```
ChrisBAshton: This is repeated twice.

spec/models/step_by_step_page_spec.rb
```diff
@@ -314,6 +314,12 @@
       step_by_step_page.mark_draft_updated
 
       expect(step_by_step_page.can_be_published?).to be false
+    end
+
+    it 'cannot be published if all steps do not have content' do
```
ChrisBAshton: Is it worth writing an additional test which has two steps - one with content, one without - and ensures that it cannot be published?

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -47,7 +38,14 @@
                 destructive: true
               } %>
             <% end %>
-          <% else %>
+          <% end %>
+          <% if @step_by_step_page.can_be_published? %>
```
ChrisBAshton: Worth making this an `elsif` instead?

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -57,7 +55,9 @@
           <% end %>
 
           <% if @step_by_step_page.can_discard_changes? %>
-            <p class="govuk-body govuk-!-margin-top-5">Or</p>
+            <% if @step_by_step_page.can_be_published? %>
```
ChrisBAshton: Why are we checking for this? As far as I know, we don't currently have any situation where you can discard changes but you can't publish 🤔

leenagupte: You could have added a new step that doesn't have content, to a published step by step, and then want to discard any and all new changes.

app/views/step_by_step_pages/publish_or_delete.html.erb
leenagupte: I don't think it makes much of a difference.

spec/models/step_by_step_page_spec.rb
leenagupte: This is covered by the first step, but I'll update the test name to make that clear.

app/views/step_by_step_pages/publish_or_delete.html.erb
leenagupte: I've added an `elsif` to handle showing the error message.

app/models/step_by_step_page.rb
```diff
@@ -83,6 +83,10 @@ def can_be_edited?
     !scheduled_for_publishing?
   end
 
+  def steps_have_content?
+    steps.any? && steps.map(&:contents).all?(&:present?)
+  end
```
ChrisBAshton: Can the movement of this code be fixed up to the 54db8c2c994add0711568d3530ab1bf5434679ef commit?

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -25,7 +25,7 @@
 
 <div class="govuk-grid-row publish-or-delete">
   <div class="govuk-grid-column-full">
-    <% if @step_by_step_page.can_be_published? || @step_by_step_page.can_discard_changes? ||  @step_by_step_page.scheduled_for_publishing?%>
+    <% if @step_by_step_page.can_be_published? || @step_by_step_page.can_discard_changes? ||  @step_by_step_page.scheduled_for_publishing? || !@step_by_step_page.steps_have_content? %>
```
ChrisBAshton: This is getting quite long - any chance it can be encapsulated in a controller method? 🤔 

If not, there's a double-space before `@step_by_step_page.scheduled_for_publishing?` that can be fixed.

app/models/step_by_step_page.rb
leenagupte: Do you mean the movement of `steps.any?`

app/views/step_by_step_pages/publish_or_delete.html.erb
leenagupte: This `if` determines whether or the grey background for the buttons is displayed and nothing else. Ideally presentation logic should be moved to a `Presenter` class an not the controller.

I think it's fine to leave this where it is for now as the tabbed layout is on it's way out, so it's not worth the effort of creating a new presenter.

I'll fix the double space problem and fix up.

app/models/step_by_step_page.rb
ChrisBAshton: I mean the movement of this method (`steps_have_content?`)

leenagupte: So I made `steps_have_content?` public at the moment that I needed to access it outside of the class. If it's hard to see from the diff why the method has moved, I can update the commit message.

ChrisBAshton: Ah ok, makes sense 👍

PR #702, opened 2019-08-21
Bypass changelog screen when first publishing step by steps

ChrisBAshton:
When first publishing a step by step, it makes little sense to provide a changenote or
to designate a change as Major or Minor.
This change makes the publish screen show a "Publish" button for draft step by steps,
which publishes immediately. Published step by steps will continue to have a "Publish
changes" button with an intermediary changenotes screen.

Trello: https://trello.com/c/VKLFfznn/120-update-the-publishing-flow-so-the-change-note-minor-major-only-shows-if-updating-a-live-version

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -64,9 +64,8 @@ class StepByStepPagesController < ApplicationController
     if request.post?
       @publish_intent = PublishIntent.new(publish_intent_params)
       if @publish_intent.valid?
+        note_description = publish_note_description
         publish_page(@publish_intent)
-        custom_note = " with note: #{@publish_intent.change_note}" unless @publish_intent.change_note.empty?
-        note_description = "#{@publish_intent.update_type.capitalize} update published by #{current_user.name}#{custom_note}"
         generate_internal_change_note(note_description)
         set_change_note_version
         redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been published."
@@ -201,6 +200,13 @@ private
     payload = Services.publishing_api.get_content(@step_by_step_page.content_id, version: published_version).to_hash
 
     StepByStepPageReverter.new(@step_by_step_page, payload).repopulate_from_publishing_api
+  end
+
+  def publish_note_description
+    return "First published by #{current_user.name}" unless @step_by_step_page.has_been_published?
+
+    custom_note = " with note: #{@publish_intent.change_note}" unless @publish_intent.change_note.empty?
+    "#{@publish_intent.update_type.capitalize} update published by #{current_user.name}#{custom_note}"
   end
 
   def generate_internal_change_note(note_description)
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -62,7 +62,7 @@ class StepByStepPagesController < ApplicationController
   def publish
     set_current_page_as_step_by_step
     if request.post?
-      @publish_intent = PublishIntent.new(params)
+      @publish_intent = PublishIntent.new(publish_intent_params)
       if @publish_intent.valid?
         publish_page(@publish_intent)
         custom_note = " with note: #{@publish_intent.change_note}" unless @publish_intent.change_note.empty?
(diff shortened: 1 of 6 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -123,6 +123,15 @@
     then_I_see_a_page_reverted_success_notice
   end
 
+  scenario "User publishes changes to a live step by step page" do
+    given_there_is_a_published_step_by_step_page_with_unpublished_changes
+    and_I_visit_the_publish_or_delete_page
+    and_I_click_button "Publish changes"
```
leenagupte: There is another method doing something similar: https://github.com/alphagov/collections-publisher/blob/f8401b53dd3f48766d63dc4b9f050a856d190fa7/spec/features/managing_step_by_step_pages_spec.rb#L373, perhaps they could be consolidated?

app/views/step_by_step_pages/publish_or_delete.html.erb
```diff
@@ -41,10 +41,18 @@
...
-            } %>
+            <% if @step_by_step_page.has_been_published? %>
+              <%= render "govuk_publishing_components/components/button", {
+                text: "Publish changes",
+                href: step_by_step_page_publish_path(@step_by_step_page)
+              } %>
+            <% else %>
+              <%= form_for(@step_by_step_page, url: step_by_step_page_publish_path(@step_by_step_page), method: :post) do %>
+                <input type="hidden" name="update_type" value="minor">
+                <input type="hidden" name="change_note" value="">
```
leenagupte: Rather than adding hidden fields here, I think it would be better for the `publish` action in the controller to set the default values for `PublishIntent`.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -64,9 +64,8 @@ def publish
     if request.post?
       @publish_intent = PublishIntent.new(params)
       if @publish_intent.valid?
+        note_description = publish_note_description(@publish_intent)
```
leenagupte: As an instance variable `@publish_intent` is accessible by the entire object, so it doesn't need to be passed to `publish_note_description`.

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -199,6 +198,15 @@ def revert_page
     StepByStepPageReverter.new(@step_by_step_page, payload).repopulate_from_publishing_api
   end
 
+  def publish_note_description(publish_intent)
+    if @step_by_step_page.has_been_published?
+      custom_note = " with note: #{publish_intent.change_note}" unless publish_intent.change_note.empty?
+      "#{publish_intent.update_type.capitalize} update published by #{current_user.name}#{custom_note}"
+    else
+      "First published by #{current_user.name}"
+    end
```
leenagupte: You could use an `unless` here rather than if/else:

```
return "First published by #{current_user.name}" unless @step_by_step_page.has_been_published?

custom_note = " with note: #{publish_intent.change_note}" unless publish_intent.change_note.empty?
 "#{publish_intent.update_type.capitalize} update published by #{current_user.name}#{custom_note}"
```

spec/support/step_nav_steps.rb
```diff
@@ -79,8 +79,8 @@ def given_there_is_a_published_step_by_step_page_with_unpublished_changes
     expect(@step_by_step_page.status[:name]).to eq 'unpublished_changes'
   end
 
-  def given_there_is_a_step_by_step_page_assigned_to_me
-    @step_by_step_page = create(:step_by_step_page_with_navigation_rules, assigned_to: stub_user.name)
+  def given_I_am_assigned_to_a_live_step_by_step_page_with_unpublished_changes
+    @step_by_step_page = create(:published_step_by_step_page, draft_updated_at: Time.zone.now, assigned_to: stub_user.name)
```
leenagupte: Why was this method changed? How does it related to setting the change note to "First published..."?

ChrisBAshton: The "User publishes and then makes more changes to a step by step page" scenario started to fail, because it was trying to assert:

`then_there_should_be_a_change_note "Minor update published by #{stub_user.name}"`

But the `create(:step_by_step_page_with_navigation_rules, assigned_to: stub_user.name)` creates a step by step without setting `published_at`, meaning it was being treated as a first publish, meaning the change note became "First published by Test author", therefore failing the change note assertion.

So I've now made sure it calls a factory method that creates a _published_ step by step, but with a draft (so that there is something to publish), and have renamed the method to make it clearer what it's doing.

leenagupte: Thanks for explaining! Can you add a version of that to the commit note?

PR #703, opened 2019-08-22
Upgrade govuk_publishing_components (19.0.0) & use summary_list

ChrisBAshton:
Trello card: https://trello.com/c/qXVCEouI/23-show-the-content-title-slug-introduction-description-using-the-summary-component

Depended on by https://github.com/alphagov/collections-publisher/pull/709.

## What

1. Upgrades govuk_publishing_components to 19.0.0, and
2. Uses summary_list component to render new sections on the 'step by step' admin page: 'Content', showing title/slug/intro/description (and an Edit link).
3. Adds tests so we can start having confidence that the right content is appearing on this page.

## Why upgrade govuk_publishing_components?

The summary list component was added to govuk_publishing_components
v18.2.0, hence the upgrade. As this is a major version change, there have
been some breaking changes (all detailed in the commit message).

A note from Alex:

> The legacy mode is not required in this app, but I'd keep it on for a while
> until we figure out how to use the new colours all over the app.

Since opening the PR, govuk_publishing_components v19 has been released,
and contains a number of changes we've made to components to feed into the
new step by step design, so rather than upgrade our dependency twice in 2 PRs,
I've just bumped straight to v19.
v19 is another breaking change but only affects apps that rely on mixins and variables
from govuk_frontend_toolkit (which is no longer included in v19 of govuk_publishing_components).
This doesn't appear to apply to collections-publisher, as evidenced in the screenshots below.

## Index before & after

There shouldn’t be any differences in the screenshots below. (NB there _were_ differences the first time I uploaded screenshots, but I’ve made some tweaks and now the output is about the same).

| Before | After |
|--------|------|
|Screen Shot 2019-09-03 at 11 39 17|Screen Shot 2019-09-03 at 11 25 58|

## Scheduling page before & after

(I've had to provide the now deprecated autocomplete module, so have taken screenshots of this page where it is used. Not quite sure why the 'Before' doesn't show an arrow next to the time, as I'm sure that used to be there... 🤔 )

| Before | After |
|--------|------|
|Screen Shot 2019-09-03 at 11 39 08|Screen Shot 2019-09-03 at 11 26 22|

## Steps page before & after

Now have a "Content" section with the title, slug, description & introduction and a link to edit them.

| Before | After |
|--------|------|
|Screen Shot 2019-09-03 at 11 38 54|Screen Shot 2019-09-03 at 11 26 42|

And here is the summary-list component with no 'Edit' link, which happens when the step by step is not editable (eg when scheduled):

Screen Shot 2019-09-03 at 12 10 23

### DIFF
--- a/app/assets/stylesheets/components/_markdown-editor.scss
+++ b/app/assets/stylesheets/components/_markdown-editor.scss
@@ -21,13 +21,13 @@ $app-toolbar-button-size: 50px;
   display: block;
   border: $govuk-border-width-form-element solid $govuk-input-border-colour;
   border-bottom: 0;
-  background-color: govuk-colour("white");
+  background-color: govuk-colour("white", $legacy: "white");
 }
 
 .app-c-markdown-editor__preview-toggle {
   width: 100%;
   border-bottom: $govuk-border-width-form-element solid $govuk-border-colour;
-  background-color: govuk-colour("grey-3");
+  background-color: govuk-colour("light-grey", $legacy: "grey-3");
 }
 
 .app-c-markdown-editor__button {
@@ -54,11 +54,11 @@ $app-toolbar-button-size: 50px;
   border-right: $govuk-border-width-form-element solid $govuk-border-colour;
   border-left: $govuk-border-width-form-element solid $govuk-border-colour;
   color: $govuk-text-colour;
-  background: govuk-colour("grey-4");
+  background: govuk-colour("light-grey", $legacy: "grey-4");
   text-decoration: none;
 
   &:focus {
-    background-color: govuk-colour("grey-4");
+    background-color: govuk-colour("light-grey", $legacy: "grey-4");
   }
 
   &:first-child {
@@ -78,7 +78,7 @@ $app-toolbar-button-size: 50px;
   .govuk-textarea {
     padding: govuk-spacing(6);
     overflow: scroll;
-    background-color: govuk-colour("grey-4");
+    background-color: govuk-colour("light-grey", $legacy: "grey-4");
   }
 
   .govuk-govspeak {
@@ -89,7 +89,7 @@ $app-toolbar-button-size: 50px;
 }
 
 .app-c-markdown-editor__toolbar {
-  background-color: govuk-colour("grey-4");
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
   line-height: 0;
 }
 
@@ -99,18 +99,19 @@ $app-toolbar-button-size: 50px;
 }
 
 .app-c-markdown-editor__toolbar-button {
-  @include govuk-focusable-fill;
   display: inline-block;
   width: $app-toolbar-button-size;
   height: $app-toolbar-button-size;
   background: none;
 
   &:focus {
-    outline-offset: -$govuk-focus-width;
+    @include govuk-focused-text;
+    outline: none;
+    box-shadow: none;
   }
 
   &:hover {
-    background-color: govuk-colour("grey-3");
+    background-color: govuk-colour("light-grey", $legacy: "grey-3");
     cursor: pointer;
   }
 }
@@ -127,7 +128,7 @@ $app-toolbar-button-size: 50px;
   width: auto;
   padding: govuk-spacing(3);
   border: 0;
-  border-left: 4px solid govuk-colour("white");
+  border-left: 4px solid govuk-colour("white", $legacy: "white");
   background: transparent;
   vertical-align: top;
 }
--- a/app/assets/stylesheets/utilities/_overrides.scss
+++ b/app/assets/stylesheets/utilities/_overrides.scss
@@ -1,7 +1,3 @@
-.gem-c-layout-for-admin {
-  @include _govuk-font-face-nta;
-}
-
 .gem-c-phase-banner {
   .gem-c-phase-banner__banner-item {
     display: none;
@@ -34,12 +30,12 @@
 
 .govuk-tag {
   &.govuk-tag--warning {
-    background-color: govuk-colour('yellow');
-    color: govuk-colour('black');
+    background-color: govuk-colour("yellow", $legacy: "yellow");
+    color: govuk-colour("black", $legacy: "black");
   }
 
   &.govuk-tag--success {
-    background-color: govuk-colour('green');
+    background-color: govuk-colour("green", $legacy: "green");
   }
 }
 
--- a/app/views/step_by_step_pages/show.html.erb
+++ b/app/views/step_by_step_pages/show.html.erb
@@ -18,6 +18,8 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/summary_list", @step_by_step_page_presenter.summary_list_params %>
+
     <table class="govuk-table step-list">
       <thead class="govuk-table__head">
         <tr class="govuk-table__row">
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -39,6 +39,14 @@ RSpec.feature "Managing step by step pages" do
     then_I_see_the_new_step_by_step_page
   end
 
+  scenario "User visits an existing step by step page" do
+    given_there_is_a_step_by_step_page
+    when_I_view_the_step_by_step_page
+    then_I_can_see_a_summary_section
+    and_I_can_edit_the_summary_section
+    and_I_can_see_a_metadata_section
+  end
+
   scenario "Validation fails" do
     when_I_visit_the_new_step_by_step_form
     and_I_fill_in_the_form_with_invalid_data
@@ -299,6 +307,37 @@ RSpec.feature "Managing step by step pages" do
 
   def then_I_see_the_step_by_step_page
     expect(page).to have_content("How to be amazing")
+  end
+
+  def then_I_can_see_a_summary_section
+    within_summary_section do
+      expect(page).to have_content "Content"
+      expect(page).to have_content "Title #{@step_by_step_page.title}"
+      expect(page).to have_content "Slug #{@step_by_step_page.slug}"
+      expect(page).to have_content "Introduction #{@step_by_step_page.introduction}"
+      expect(page).to have_content "Description #{@step_by_step_page.description}"
+    end
+  end
+
+  def and_I_can_edit_the_summary_section
+    within_summary_section do
+      expect(page).to have_link("Edit", :href => edit_step_by_step_page_path(@step_by_step_page))
+    end
+  end
+
+  def within_summary_section
+    expect(page).to have_css(".gem-c-summary-list")
+    within(".gem-c-summary-list") do
+      yield
+    end
+  end
+
+  def and_I_can_see_a_metadata_section
+    within(".gem-c-metadata") do
+      expect(page).to have_content("Status: Draft")
+      expect(page).to have_content("Last saved")
+      expect(page).to have_content("Created")
+    end
   end
 
   def and_I_visit_the_publish_or_delete_page
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -171,6 +171,8 @@ RSpec.feature "Managing step by step pages" do
       and_there_should_be_a_change_note "Minor update scheduled by Test author for publishing at 10:26am on 20 April 2030"
       and_the_step_by_step_is_not_editable
       when_I_view_the_step_by_step_page
+      then_I_can_see_a_summary_section
+      but_I_cannot_edit_the_summary_section
       then_I_can_preview_the_step_by_step
       and_the_steps_can_be_checked_for_broken_links
     end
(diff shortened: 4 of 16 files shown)


app/assets/stylesheets/components/_markdown-editor.scss
```diff
@@ -99,7 +99,7 @@ $app-toolbar-button-size: 50px;
 }
 
 .app-c-markdown-editor__toolbar-button {
-  @include govuk-focusable-fill;
+  // @include govuk-focusable-fill;
```
ChrisBAshton: TODO

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -1,5 +1,5 @@
 .gem-c-layout-for-admin {
-  @include _govuk-font-face-nta;
+  // @include _govuk-font-face-nta;
```
ChrisBAshton: TODO

app/views/step_by_step_pages/show.html.erb
```diff
@@ -90,15 +90,10 @@
...
-        metadata = {}
-        metadata.store("Status", @step_by_step_page.status[:text])
-        metadata.store("Last saved", format_full_date_and_time(@step_by_step_page.updated_at))
-        metadata["Last saved"] = "#{metadata["Last saved"]} by #{@step_by_step_page.assigned_to}" if @step_by_step_page.assigned_to.present?
-        metadata.store("Created", format_full_date_and_time(@step_by_step_page.created_at))
-        metadata.store("Links checked", format_full_date_and_time(@step_by_step_page.links_last_checked_date)) if @step_by_step_page.links_checked?
-      %>
-      <%= render "govuk_publishing_components/components/metadata", { other: metadata } %>
+      <%= render "govuk_publishing_components/components/summary_list", {
+        title: "Overview",
```
leenagupte: The designs attached to the Trello card don't have "Overview" as a title. Have they changed?

app/views/step_by_step_pages/show.html.erb
```diff
@@ -90,15 +90,10 @@
...
+        items: StepByStepPresenter.new(@step_by_step_page).summary_list
```
leenagupte: Creating a presenter is great idea! It can be called from the `show` method in the controller, and still be accessed here.

app/presenters/step_by_step_presenter.rb
```diff
@@ -0,0 +1,33 @@
+class StepByStepPresenter
```
leenagupte: I like the idea of a presenter, I'm just concerned that the name is too similar to `StepNavPresenter`  and could cause confusion. Perhaps `StepByStepPagePresenter` so it's named similarly to the model?

spec/presenters/step_by_step_presenter_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "rails_helper"
+require "gds_api/test_helpers/link_checker_api"
+
+RSpec.describe StepByStepPresenter do
+  include GdsApi::TestHelpers::LinkCheckerApi
+  include AuthenticationControllerHelpers
```
leenagupte: I don't think you need to explicitly include `AuthenticationControllerHelpers` here as it's already being included here: https://github.com/alphagov/collections-publisher/blob/master/spec/support/authentication.rb#L30

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: I wrongly assumed you have to provide a title, [according to the component docs](https://components.publishing.service.gov.uk/component-guide/select). Have just tried it without the title and it renders fine!

Will edit this PR and have raised a [PR to add a title-less example to the summary list component](https://github.com/alphagov/govuk_publishing_components/pull/1070).

spec/presenters/step_by_step_presenter_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    subject { described_class.new(step_nav).summary_list }
+
+    it "presents a list of key/value metadata" do
+      expect(subject).to eq(default_summary)
+    end
+
+    context "assigned_to is set" do
+      let(:step_nav) { create(:step_by_step_page_with_navigation_rules, assigned_to: stub_user.name) }
```
leenagupte: The original `step_nav` could be re-used here rather than creating a new object.
For example: `step_nav.assigned_to = stub_user.name`. The assignment would have to be made inside the test, but as there's only one test, the context isn't really needed.

spec/presenters/step_by_step_presenter_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      it "shows who made the most recent change" do
+        expect(subject[1]).to eq(
```
leenagupte: `subject[1]` could start failing if the order of the information changes. Perhaps a `select` could be used to find the hash with a `field` that equals "Last saved". 

See: https://stackoverflow.com/questions/2244915/how-do-i-search-within-an-array-of-hashes-by-hash-values-in-ruby

spec/presenters/step_by_step_presenter_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          field: "Last saved",
+          value: "#{format_full_date_and_time(Time.zone.now)} by #{stub_user.name}"
+        )
+      end
+    end
+
+    context "links have been checked" do
+      before :each do
+        stub_link_checker_api_create_batch(
```
leenagupte: I don't think link-checker-api needs to be stubbed. I think you should be able to create a `link-report` factory, and link it to a step.

For example:
```
step = create(:step, step_by_step_page: step_nav)
create(link_report, step: step)
```

spec/presenters/step_by_step_presenter_spec.rb
ChrisBAshton: We're testing a presenter rather than a controller, so it isn't getting pulled in automatically. Without this line I get errors along the lines of `unknown method stub_user`.

spec/presenters/step_by_step_presenter_spec.rb
ChrisBAshton: Thanks, much nicer!

app/views/step_by_step_pages/show.html.erb
```diff
@@ -97,6 +98,19 @@
...
+      title: "Sidebar settings",
+      id: "sidebar-settings",
+      edit: {
+        href: step_by_step_page_navigation_rules_path(@step_by_step_page),
+        data_attributes: {
+          gtm: "edit-sidebar-settings"
+        }
+      }
+    } do %>
+      <p class="govuk-body">Guidance around sidebar settings</p>
```
leenagupte: This section feels incomplete. Is there guidance to link to or actual guidance that can be displayed here? 

Is there a plan to display the current sidebar settings here?

leenagupte: Should the sidebar settings section be hidden if there aren't any navigation rules?

ChrisBAshton: Ha, I copied the text exactly as per the design, but didn't stop to read it through properly. You're right, this looks like placeholder text, not sure what's actually supposed to go here.

I'll put a comment on the card when I'm back on my work laptop (or you could, in my absence, to speed things along!)

Tests pass locally & on CI, I wonder if you've run `govuk-docker build` yet? As this PR updates the govuk_publishing_components gem, and adds accessible-autocomplete NPM module. (It's possible `govuk-docker` doesn't account for this but that somehow the Jenkinsfile does - so you may need to run a `yarn install`).

leenagupte: I've removed and rebuilt the image and the container, and run `yarn install`, but the tests still fail.

leenagupte:
The drop down arrow is configured here: https://github.com/alphagov/collections-publisher/blob/master/app/assets/javascripts/components/autocomplete.js#L41 I'm not sure why it has started showing up now though.

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: Have moved the addition of those sections to a separate PR so that we can get this PR merged.

https://github.com/alphagov/collections-publisher/pull/709

Sounds like we just need to get to the bottom of why the tests are failing on your machine before we proceed 🤔 come find me when you're in the office!

ChrisBAshton: Thanks for clarifying the horizontal lines comment. I've added a commit to override the component styles so that the line is only added to the last row item. I'm not sure if this was intended to be the default when Alex raised the [PR adding summary_list to govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/1061) (and that therefore the style should be changed there) or whether collections-publisher is the only app that will want summary_list without the horizontal lines. It seems like the pragmatic option to perform the override in a commit here, which we can revert in future if the summary_list component is changed.

As for the tests, I've tried deleting my containers and force-deleted the collections-publisher image, and removed my `node_modules` directory. I then rebuilt the collections-publisher image; the tests are still passing.

I notice the [Jenkinsfile has a `beforeTest` step](https://github.com/alphagov/collections-publisher/blob/master/Jenkinsfile#L7-L9) that executes `yarn install`, which is why CI is fine. And [there was a change to govuk-docker](https://github.com/alphagov/govuk-docker/commit/047c905cba5d107b2496e685a8908bf21a8c6523) to install yarn dependancies for collections publisher. Is your `~/govuk/govuk-docker` on `master` and up to date?

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -75,3 +75,15 @@
...
 
 }
+
+.govuk-summary-list__key,
+.govuk-summary-list__value,
+.govuk-summary-list__actions {
+  border-bottom: 0;
+
+  // only apply border to the last row of the summary list
+  .govuk-summary-list .govuk-summary-list__row:last-child & {
```
leenagupte: I wonder if this is an oversight in the gem. This component was copied from content-publisher, which does not display a line between each summary item, so perhaps rather than having an override here, this styling should be added to the gem?

ChrisBAshton: Ah, looks like the summary_list component in [govuk_publishing_components inherits from govuk-frontend](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_summary-list.scss#L1) (i.e. Design System). In Design System, it [_does_ have the horizontal lines in between rows](https://design-system.service.gov.uk/components/summary-list/) (although there is the [option to override it with `govuk-summary-list--no-border`](https://design-system.service.gov.uk/components/summary-list/#summary-list-without-borders).

Unfortunately there's no way to pass that class to the component, so I'm still going to have to make a PR to govuk_publishing_components 😅

ChrisBAshton: Have raised https://github.com/alphagov/govuk_publishing_components/pull/1073.

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: For the record, we paired on this IRL and couldn't figure out the root cause, but it was fixed on Leena's machine by deleting and re-cloning the repo, and deleting & rebuilding the images & containers.

app/assets/stylesheets/utilities/_overrides.scss
ChrisBAshton: Merged PR. Now blocked on getting the gem bumped & published.

Once that's done, can pull in latest gem, remove the override commit, take new screenshots for the PR description and get this merged 🎉

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,12 +21,12 @@
     <%= render "govuk_publishing_components/components/summary_list", {
       borderless: true,
       title: "Content",
-      edit: {
+      edit: @step_by_step_page.can_be_edited? ? {
```
leenagupte: I think this logic should be moved to a method in the presenter and can be simplified. Rather than setting `edit` to `nil` it can be excluded completely:

```
    summary_list_params = {
      borderless: true,
      title: "Content",
      items: [
         .
         .
         .
      ]
    }

    if @step_by_step_page.can_be_edited?
      summary_list_params.merge!(
        edit: {
          href: edit_step_by_step_page_path(@step_by_step_page),
          data_attributes: {
            gtm: "edit-title-summary-body"
          }
        }
      )
    end
```

Then the component would only need to call:
`<%= render "govuk_publishing_components/components/summary_list", @step_by_step_page_presenter.summary_list_params %>`

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -309,14 +310,27 @@ def then_I_see_the_step_by_step_page
...
     expect(page).to have_css(".gem-c-summary-list")
     within(".gem-c-summary-list") do
       expect(page).to have_content "Content"
-      expect(page).to have_link("Edit", :href => edit_step_by_step_page_path(@step_by_step_page))
       expect(page).to have_content "Title #{@step_by_step_page.title}"
       expect(page).to have_content "Slug #{@step_by_step_page.slug}"
       expect(page).to have_content "Introduction #{@step_by_step_page.introduction}"
       expect(page).to have_content "Description #{@step_by_step_page.description}"
+
+      if editable
```
leenagupte: I think for readability this should be split into two separate steps:

`and_I_can_edit_the_metadata` and `and_I_cannot_edit_the_metadata`, with each step being explicit about what can and can't be viewed.

I'm using "metadata" rather than "content" because I think the term content is going to become overloaded and it can also refer to individual and collective step content.

ChrisBAshton:
(Noticed I had some `@step_by_step_page`s in the presenter that could become `step_by_step_page`s - have pushed up a tiny tweak)

PR #704, opened 2019-08-23
Update index page filters to enable pre-filtered results

leenagupte:
Trello: https://trello.com/c/eo8mwZlv

## What's changed
Remove filtering results by javascript, and filter using database queries to allow direct linking to a pre-filtered page of results. 

## Why
This means that content designers can send each other links to a pre-filtered results, rather than asking people who are unfamiliar with the tool to filter the results themselves. It also allows the list to be linked to from other sources of guidance.

Also fixes a bug where it wasn't possible to filter results filtering by both "Title" and "Status" as the JS submitted the query on key stroke rather than waiting for a button press.

### Before
Screen Shot 2019-08-27 at 15 49 39

### After
Screen Shot 2019-08-27 at 15 41 55
Screen Shot 2019-08-27 at 15 48 28

URL: `http://collections-publisher.dev.gov.uk/step-by-step-pages?title_or_url=one&status=scheduled`

### Note to reviewer
There is already a [card in the backlog](https://trello.com/c/C4HHCrAX) to rename and move the position of the "Create new" button, so it hasn't been attempted here.


app/services/step_by_step_filter.rb
```diff
@@ -1,11 +1,25 @@
 class StepByStepFilter
-  attr_reader :status
+  attr_reader :status, :title_or_url
```
ChrisBAshton: All the example test cases in this (second) commit test against the title, but not the URL. I think it would be cleaner to have `:title` only in this commit, and expand it out to be `:title_or_url` in a future commit. (That, or leave the code as-is but write additional tests against the URL in this commit)

~I'd also suggest renaming URL to `slug`, it's a little more specific and is the field that is actually getting matched against.~ EDIT: now I've gone through the other commits, I can see it matches on full URLs as well as slugs.

app/services/step_by_step_filter.rb
```diff
@@ -7,10 +7,15 @@ def initialize(params)
   end
 
   def results
-    results = []
-    results += filter_by_status if status.present?
-    results += filter_by_title_or_url if title_or_url.present?
-    results
+    if status.present? && title_or_url.present?
+      filter_by_status & filter_by_title_or_url
```
ChrisBAshton: Nice!

app/services/step_by_step_filter.rb
```diff
@@ -26,6 +26,10 @@ def filter_by_status
 
   def filter_by_title_or_url
     StepByStepPage.where("title LIKE ?", "%#{title_or_url}%")
-      .or(StepByStepPage.where(slug: title_or_url))
+      .or(StepByStepPage.where(slug: slug_from_title_or_url))
```
ChrisBAshton: 🎉

spec/services/step_by_step_filter_spec.rb
```diff
@@ -166,4 +166,22 @@
...
+      filter_params = {
+        status: "",
+        title_or_url: ""
+      }
+      results = described_class.new(filter_params).results
+
+      expect(results.count).to eq(3)
+    end
+
+    it "returns all step by steps if there aren't any filters params" do
```
ChrisBAshton:
```suggestion
    it "returns all step by steps if there aren't any filter params" do
```

app/services/step_by_step_filter.rb
```diff
@@ -32,4 +32,8 @@ def filter_by_title_or_url
   def slug_from_title_or_url
     title_or_url.split('/').last
   end
+
+  def ordered_results(results)
+    results.sort_by(&:title)
```
ChrisBAshton: I wonder if it would be cleaner to call _this_ `results`, and to rename the method on line 9 to `unordered_results`? That way we wouldn't have to make all those `ordered_results` calls, and this method would simply be:

```ruby
def results
  unordered_results.sort_by(&:title)
end

def unordered_results
  if status.present? && title_or_url.present?
    filter_by_status & filter_by_title_or_url
  elsif ...etc
end
```

spec/services/step_by_step_filter_spec.rb
```diff
@@ -203,5 +206,29 @@
...
+        order_by: "scheduled_at"
+      }
+      results = described_class.new(filter_params).results
+
+      expect(results.count).to eq(2)
+      expect(results.first.title).to eq(scheduled_step_by_step.title)
+      expect(results.last.title).to eq(new_step_by_step.title)
+    end
+
+    it "returns all results if order by attribute is invalid" do
```
ChrisBAshton: Smart 👍

app/views/step_by_step_pages/index.html.erb
```diff
@@ -28,56 +28,83 @@
...
+            text: "Status",
+            html_for: "status",
+            bold: true
+          } %>
+          <%= render "govuk_publishing_components/components/select", {
+            id: "status",
+            name: "status",
+            label: "",
+            full_width: true,
+            value: params[:status],
```
ChrisBAshton: `value` isn't used outside of the `options` array.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -28,56 +28,83 @@
...
+        <div class="govuk-grid-column-full">
+          <%= render "govuk_publishing_components/components/label", {
+            text: "Status",
+            html_for: "status",
+            bold: true
+          } %>
+          <%= render "govuk_publishing_components/components/select", {
+            id: "status",
+            name: "status",
+            label: "",
```
ChrisBAshton: Are we not specifying `label` here because it doesn't come out as bold, so we're defining our own label separately?

I wonder if we can just remove the `label: ""` altogether.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -28,56 +28,83 @@
...
+              {
+                text: "All",
+                data_attributes: {
+                  show: "all"
+                },
+              },
+              {
+                text: "Draft",
+                value: "draft",
+                data_attributes: {
```
ChrisBAshton: I don't think these `data_attributes` are needed anymore - only the `value` gets used in generating the filter URL 🤔

app/views/step_by_step_pages/index.html.erb
leenagupte: Yes. "label" in the select component is not bold. I tried removing, but the component doesn't render properly, so I think it's a required field.

app/views/step_by_step_pages/index.html.erb
leenagupte: I thought `data_attributes` were for tracking? https://github.com/alphagov/govuk_publishing_components/blob/3fdfd33366502f61d548f45399b026860e3ffafd/lib/govuk_publishing_components/presenters/select.rb#L11

ChrisBAshton: Ah, true 👍

app/services/step_by_step_filter.rb
leenagupte: I thought about calling it `title` and then renaming it, but thought it was unnecessary as the filtering by slug/base_path/url would start to be built in the next commit, and renaming variables is extra noise when looking at changes. I'd rather keep the URL tests separate and they need specific changes to the search query to work.

spec/services/step_by_step_filter_spec.rb
leenagupte: Updated and fixed up.

app/services/step_by_step_filter.rb
leenagupte: I went with `filtered_results` rather than `unordered` because strictly speaking not all of the results are unordered at this point (`by_title` returns a list of all SBS ordered by title).

Updated and fixed up.

app/views/step_by_step_pages/index.html.erb
leenagupte: Updated and fixed up.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -107,6 +107,7 @@
   </div>
 
   <div class="govuk-grid-column-three-quarters">
+    <p class="govuk-body govuk-!-margin-bottom-1"><%=pluralize(@step_by_step_pages.count, "step by step")%></p>
```
ChrisBAshton: Looks like we're missing spaces after the `%=` and before the `%>` here.

leenagupte: Update and fixed up.

PR #707, opened 2019-08-28
Move metadata summary content into new StepByStepPresenter

ChrisBAshton:
Moved the business logic into a presenter rather than muddying the
view.

Did this as part of #703 then realised it was out of scope, so moving into its own PR.
There's no Trello card dependent on this, but it's probably a good idea to clean up
our view for maintainability.

### DIFF
--- a/app/presenters/step_by_step_page_presenter.rb
+++ b/app/presenters/step_by_step_page_presenter.rb
@@ -1,0 +1,34 @@
+class StepByStepPagePresenter
+  include TimeOptionsHelper
+  attr_reader :step_by_step_page
+
+  def initialize(step_by_step_page)
+    @step_by_step_page = step_by_step_page
+  end
+
+  def summary_metadata
+    items = {
+      "Status" => step_by_step_page.status[:text],
+      "Last saved" => last_saved,
+      "Created" => format_full_date_and_time(step_by_step_page.created_at),
+    }
+
+    if step_by_step_page.links_checked?
+      items.merge!(
+        "Links checked" => format_full_date_and_time(step_by_step_page.links_last_checked_date)
+      )
+    end
+    items
+  end
+
+  def last_saved
+    last_saved_time = format_full_date_and_time(step_by_step_page.updated_at)
+    return "#{last_saved_time} by #{step_by_step_page.assigned_to}" if assigned?
+
+    last_saved_time
+  end
+
+  def assigned?
+    step_by_step_page.assigned_to.present?
+  end
+end
--- a/spec/presenters/step_by_step_page_presenter_spec.rb
+++ b/spec/presenters/step_by_step_page_presenter_spec.rb
@@ -1,0 +1,42 @@
+require "rails_helper"
+require "gds_api/test_helpers/link_checker_api"
+
+RSpec.describe StepByStepPagePresenter do
+  include TimeOptionsHelper
+
+  before do
+    allow(Services.publishing_api).to receive(:lookup_content_id)
+  end
+
+  describe "#summary_list" do
+    let(:step_nav) { create(:step_by_step_page_with_steps) }
+    let(:default_summary) {
+      {
+        "Status" => "Draft",
+        "Last saved" => format_full_date_and_time(Time.zone.now),
+        "Created" => format_full_date_and_time(Time.zone.now)
+      }
+    }
+
+    subject { described_class.new(step_nav).summary_metadata }
+
+    it "presents a list of key/value metadata" do
+      expect(subject).to eq(default_summary)
+    end
+
+    it "shows who made the most recent change" do
+      step_nav.assigned_to = "Firstname Lastname"
+
+      expect(subject["Last saved"]).to eq("#{format_full_date_and_time(Time.zone.now)} by #{step_nav.assigned_to}")
+    end
+
+    it "has additional metadata showing when links were checked" do
+      create(:link_report, step: step_nav.steps.first)
+      summary_with_links_checked = default_summary.merge(
+        "Links checked" => format_full_date_and_time(Time.zone.now)
+      )
+
+      expect(subject).to eq(summary_with_links_checked)
+    end
+  end
+end
--- a/app/assets/stylesheets/utilities/_overrides.scss
+++ b/app/assets/stylesheets/utilities/_overrides.scss
@@ -77,5 +77,4 @@
 
     @include govuk-clearfix;
   }
-
 }
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -14,7 +14,9 @@ class StepByStepPagesController < ApplicationController
     @step_by_step_page = StepByStepPage.new
   end
 
-  def show; end
+  def show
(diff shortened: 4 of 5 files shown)


spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    subject { described_class.new(step_nav).summary_metadata }
+
+    it "presents a list of key/value metadata" do
+      expect(subject).to eq(default_summary)
+    end
+
+    it "shows who made the most recent change" do
+      step_nav.assigned_to = stub_user.name
```
leenagupte: This is the only test that `stub_user` is used. All that's really needed is to assign `step_nav.assigned_to` a string. So rather than including the AuthenticationControllerHelpers and creating a user, you could get away with just doing:

```
step_nav.assigned_to = "Firstname Lastname"
expect(subject["Last saved".to_sym]).to eq("#{format_full_date_and_time(Time.zone.now)} by #{step_nav.assigned_to}")
```

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      expect(subject["Last saved".to_sym]).to eq("#{format_full_date_and_time(Time.zone.now)} by #{stub_user.name}")
```
leenagupte: Nice use of `subject["Last saved".to_sym]`.

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    context "links have been checked" do
+      before :each do
+        step = create(:step, step_by_step_page: step_nav)
```
leenagupte: After suggesting this to you yesterday, I realised that you don't need to create a new step. You can just do:
```
create(:link_report, step: step_nav.steps.first)
```

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+require "gds_api/test_helpers/link_checker_api"
+
+RSpec.describe StepByStepPagePresenter do
+  include GdsApi::TestHelpers::LinkCheckerApi
```
leenagupte: This line isn't needed.

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  include AuthenticationControllerHelpers
+  include GovukContentSchemaTestHelpers
```
leenagupte: This line isn't needed.

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -0,0 +1,23 @@
+class StepByStepPagePresenter
+  include TimeOptionsHelper
+
+  def initialize(step_by_step_page)
+    @step_by_step_page = step_by_step_page
+  end
+
+  def summary_metadata
```
leenagupte: I know that this method was copied wholesale from the view, but I think it could do with a refactor in the next commit to make it more readable.

For example:
an `attr_reader` could be used to access `@step_by_step_page`
the last saved text could be split into a method, 
e.g.
```
def last_saved
  last_saved_time = format_full_date_and_time(step_by_step_page.updated_at)
  return "#{last_saved_time} by #{@step_by_step_page.assigned_to}" if assigned?

  last_saved_time
end

def assigned?
  step_by_step_page.assigned_to.present?
end
```

Then in the `items` hash, set `"Last saved": last_saved`

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+    last_saved_time = format_full_date_and_time(@step_by_step_page.updated_at)
+    last_saved = "#{last_saved_time} by #{@step_by_step_page.assigned_to}" if @step_by_step_page.assigned_to.present?
+    items = {
+      "Status": @step_by_step_page.status[:text],
+      "Last saved": last_saved || last_saved_time,
+      "Created": format_full_date_and_time(@step_by_step_page.created_at),
+    }
+    if @step_by_step_page.links_checked?
```
leenagupte: Can you add a line break here.

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+      items = items.merge(
```
leenagupte: Warning: if any other key/value pairs need to be added, this will need to change to a `merge!`. That's because `merge` returns a new hash rather than mutating the original object, so any new elements added after this will be added to the original `items`.

Another option is to do `items["Links checked"] = format_full_date_and_time(@step_by_step_page.links_last_checked_date)` as that also mutates the original object.

app/presenters/step_by_step_page_presenter.rb
ChrisBAshton: Nice, thanks 👍

app/presenters/step_by_step_page_presenter.rb
ChrisBAshton: Hmm this would still work, as I reassign `items = items.merge`, so an additional key/value pair would just be an additional `items = items.merge`.

In the alternative suggestion, I think it would need to be `items["Links checked".to_sym] =`, which doesn't read very nicely. But I think `items.merge!` would be cleaner 👍

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+  include TimeOptionsHelper
+  attr_reader :step_by_step_page
+
+  def initialize(step_by_step_page)
+    @step_by_step_page = step_by_step_page
+  end
+
+  def summary_metadata
+    items = {
+      "Status": step_by_step_page.status[:text],
```
leenagupte: I didn't spot this before, but in Ruby the convention is to use hash rocket (`=>`) for string keys.
So `"Status" => step_by_step_page.status[:text]` etc.

Doing that means that the you can access data with out having to call `to_sym` on the key, so things like `items["Status"]` and `items["Last saved"]` will return a value.

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    subject { described_class.new(step_nav).summary_metadata }
+
+    it "presents a list of key/value metadata" do
+      expect(subject).to eq(default_summary)
+    end
+
+    it "shows who made the most recent change" do
+      step_nav.assigned_to = "Firstname Lastname"
+
+      expect(subject["Last saved".to_sym]).to eq("#{format_full_date_and_time(Time.zone.now)} by #{step_nav.assigned_to}")
```
leenagupte: See my comments above about hash rockets for string keys.

app/presenters/step_by_step_page_presenter.rb
ChrisBAshton: Good shout!

PR #708, opened 2019-08-28
Overhaul the steps section

ChrisBAshton:
Current WIP:

Screen Shot 2019-08-28 at 11 20 56

Trello card: https://trello.com/c/UlgsMO0Q/45-overhaul-the-steps-section-and-move-the-add-step-button-to-it-and-show-it-in-the-summary-component


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -340,6 +351,28 @@ def within_summary_section
...
+    end
+  end
+
+  def and_I_can_see_an_empty_steps_overview_section
+    expect(page).not_to have_css('.govuk-table.step-list')
+    expect(page).to have_content('No steps have been added yet.')
+  end
+
+  def and_I_can_edit_and_reorder_and_delete_steps
+    expect(page).to have_link("Reorder", :href => step_by_step_page_reorder_path(@step_by_step_page))
```
ChrisBAshton: Move this to a separate step and make it part of the commit where 'Reorder' is added.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -367,14 +369,17 @@ def step_text_in_row(row)
...
-    expect(page).to_not have_link("Reorder")
+  def but_I_cannot_edit_or_delete_steps
     all('.govuk-table.step-list .govuk-table__cell').each do |step|
       expect(step).not_to have_link("Edit")
       expect(step).not_to have_link("Delete")
     end
   end
 
+  def and_I_cannot_reorder_steps
+    expect(page).to_not have_link("Reorder", exact: true)
```
ChrisBAshton: Not sure why previous commit passes tests but current commit needs `exact: true` 🤔

app/assets/stylesheets/_steps.scss
```diff
@@ -163,6 +163,16 @@ $dark-grey-2: #6f777b;
 }
 
 .step-list {
+  th {
+    font-weight: normal;
```
ChrisBAshton: If we have to override font-weight here, I wonder if `th` is the most appropriate markup to use. `td` may be better.

app/views/shared/steps/_broken-links-summary.html.erb
```diff
@@ -1,9 +1,15 @@
 <% if step.broken_links? %>
-  <span class="govuk-tag govuk-tag--warning">
-    <%= step.broken_links.count %> broken link<%= "(s)" if step.broken_links.count > 1 %> found
+  <span class="broken-link-summary broken-link-summary--fail">
+    <svg aria-hidden="true" role="img" focusable="false" width="12" height="12" viewBox="0 0 352 512" xmlns="http://www.w3.org/2000/svg">
+      <path d="M242.72 256l100.07-100.07c12.28-12.28 12.28-32.19 0-44.48l-22.24-22.24c-12.28-12.28-32.19-12.28-44.48 0L176 189.28 75.93 89.21c-12.28-12.28-32.19-12.28-44.48 0L9.21 111.45c-12.28 12.28-12.28 32.19 0 44.48L109.28 256 9.21 356.07c-12.28 12.28-12.28 32.19 0 44.48l22.24 22.24c12.28 12.28 32.2 12.28 44.48 0L176 322.72l100.07 100.07c12.28 12.28 32.2 12.28 44.48 0l22.24-22.24c12.28-12.28 12.28-32.19 0-44.48L242.72 256z"></path>
+    </svg>
+    <%= step.broken_links.count %> broken link<%= "(s)" if step.broken_links.count > 1 %>
```
ChrisBAshton: Can move this text logic to a presenter.

spec/support/custom_rspec_matchers.rb
```diff
@@ -0,0 +1,9 @@
+module CustomRSpecMatchers
+  def have_table_heading(text)
+    have_selector(".govuk-table__header", :text => text)
+  end
+
+  def have_step_text_in_row(row:, text:)
+    have_selector(".govuk-table__body .govuk-table__row:nth-child(#{row}) .step-list__title", :text => text)
```
leenagupte: This is check is specific to step-by-steps, and as it's only need for the steps list, I think this methods would be better placed in spec/features/managing_step_by_step_pages_spec.rb

spec/support/custom_rspec_matchers.rb
```diff
@@ -0,0 +1,9 @@
+module CustomRSpecMatchers
+  def have_table_heading(text)
+    have_selector(".govuk-table__header", :text => text)
```
leenagupte: I'd question whether this method is needed at all.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -60,6 +58,10 @@
                   <%= link_to("Edit", edit_step_by_step_page_step_path(@step_by_step_page.id, step.id), class: "gem-c-button govuk-button gem-c-button--inline") %>
                   <%= link_to("Delete", step_by_step_page_step_path(@step_by_step_page.id, step.id), method: 'delete', data: { confirm: "Are you sure?" }, class: "gem-c-button govuk-button govuk-button--warning gem-c-button--inline") %>
                 </td>
+              <% else %>
+                <td class="govuk-table__cell govuk-table__cell--numeric">
+                  <%= link_to("View", edit_step_by_step_page_step_path(@step_by_step_page.id, step.id), class: "gem-c-button govuk-button gem-c-button--inline") %>
```
leenagupte: As part of the scheduling work we decided not to allow users to visit the "Edit step" page, so this "View" link isn't needed.

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -58,7 +58,7 @@ def summary_list_params
   end
 
   def steps_title
-    return { title: "Steps" } unless step_by_step_page.can_be_edited?
+    return { title: "Steps" } unless step_by_step_page.can_be_edited? && step_by_step_page.steps.present?
```
leenagupte: Hasn't it been decided to always show the "Steps" heading?

leenagupte: Hasn't it been decided to always show the "Steps" title, even if there aren't any steps?

ChrisBAshton: Yes - if you look further down, we return Steps with Reorder link if this conditional fails.

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: Ah interesting, because they can still view the step content via the step by step preview?

Because if we don't provide them with a 'View' link, they'll have no way of looking at the step markup (URLs etc). But I suppose that's not an issue so long as they can still see the output.

ChrisBAshton: Have removed 👍

app/presenters/step_by_step_page_presenter.rb
ChrisBAshton: Have now refactored it to work the same way as other methods in this presenter (ie a hash which we `merge!` edit links onto if needed)

app/views/shared/steps/_broken-links-summary.html.erb
```diff
@@ -1,5 +1,14 @@
...
+  <span class="broken-link-summary broken-link-summary--<%= step.broken_links? ? 'fail' : 'pass' %>">
+    <% if step.broken_links? %>
+      <svg aria-hidden="true" role="img" focusable="false" width="12" height="12" viewBox="0 0 352 512" xmlns="http://www.w3.org/2000/svg">
+        <path d="M242.72 256l100.07-100.07c12.28-12.28 12.28-32.19 0-44.48l-22.24-22.24c-12.28-12.28-32.19-12.28-44.48 0L176 189.28 75.93 89.21c-12.28-12.28-32.19-12.28-44.48 0L9.21 111.45c-12.28 12.28-12.28 32.19 0 44.48L109.28 256 9.21 356.07c-12.28 12.28-12.28 32.19 0 44.48l22.24 22.24c12.28 12.28 32.2 12.28 44.48 0L176 322.72l100.07 100.07c12.28 12.28 32.2 12.28 44.48 0l22.24-22.24c12.28-12.28 12.28-32.19 0-44.48L242.72 256z"></path>
+      </svg>
+    <% else %>
+      <svg aria-hidden="true" role="img" focusable="false" width="12" height="12" viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
+        <polygon points="434.8,49 174.2,309.7 76.8,212.3 0,289.2 174.1,463.3 196.6,440.9 196.6,440.9 511.7,125.8 434.8,49"/>
+      </svg>
+    <% end %>
```
ChrisBAshton: TODO - refactor this somehow.

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -28,17 +28,28 @@
   }
 }
 
-.govuk-tag {
-  &.govuk-tag--warning {
-    background-color: govuk-colour("yellow", $legacy: "yellow");
-    color: govuk-colour("black", $legacy: "black");
+.broken-link-summary {
```
alex-ju: As discussed, these styles should be moved into `objects/_broken-link-status.scss` as they're not overwrites anymore, but new BEM classes and we should prefix them with `.app-` (e.g. `.app-broken-link-status`, `.app-broken-link-status--fail`, `.app-broken-link-status--pass`) to make it clear where the styles come from.

I'm happy to keep them as partials for now.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -375,10 +378,21 @@ def and_I_can_edit_and_delete_steps
     end
   end
 
+  def but_I_cannot_edit_or_delete_steps
+    all('.govuk-table.step-list .govuk-table__cell').each do |step|
```
ChrisBAshton: Edit

app/assets/stylesheets/utilities/_overrides.scss
ChrisBAshton: Done 👍

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -57,6 +57,18 @@ def summary_list_params
     params
   end
 
+  def broken_link_status_text(step)
+    if step.broken_links?
+      "#{step.broken_links.count} broken #{'link'.pluralize(step.broken_links.count)} found"
```
leenagupte: I think using pluralise is a good idea to craft the text 👍

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -57,6 +57,18 @@ def summary_list_params
...
+    elsif step.link_report?
+      "No broken links"
+    end
+  end
+
+  def broken_link_status_classes(step)
+    "govuk-tag govuk-tag--#{step.broken_links? ? 'warning' : 'success'}"
```
leenagupte: I'm not convinced of the value of moving all of the css code into the presenter. It feels like this should live in the view as it feels like it's less clear what options are being presented to users.

Presenters are usually just used to [manipulate data](https://stackoverflow.com/questions/36458688/what-is-the-rails-presenters-folder-for) for presentation in the view.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -44,6 +44,7 @@
     when_I_view_the_step_by_step_page
     then_I_can_see_a_summary_section
     and_I_can_edit_the_summary_section
+    and_I_can_see_an_empty_steps_overview_section
```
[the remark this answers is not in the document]
ChrisBAshton: Renamed to `and_I_can_see_a_steps_overview_section_with_no_steps`.

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -63,3 +63,16 @@
     @include govuk-clearfix;
   }
 }
+
+// scss-lint:disable IdSelector
+.gem-c-summary-list#steps {
```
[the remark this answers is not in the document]
ChrisBAshton: Have added a Why to the commit 👍

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -40,4 +39,46 @@
...
+      allow(step).to receive(:link_report?).and_return(true)
+    end
+
+    it "should use singular 'link' if only one link is broken" do
+      allow(step).to receive(:broken_links?).and_return(false)
+
+      expect(subject.broken_link_status_text(step)).to eq "No broken links"
+    end
+
+    it "should use singular 'link' if only one link is broken" do
```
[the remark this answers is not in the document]
ChrisBAshton: Use of `pluralize` is an implementation detail - someone could come along and change that and break the pluralising behaviour, having these tests is a good safeguard against that.

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -57,6 +57,20 @@ def summary_list_params
     params
   end
 
+  def broken_link_status(step)
+    return step.title unless step.link_report?
+
+    "
+      #{step.title}
+      <div class=\"step-broken-links__summary\">
```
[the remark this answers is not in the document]
ChrisBAshton: Agreed, this wasn't the way I wanted to approach it.

The problem is, the design has SVG icons and red/green status text:

Screen Shot 2019-09-09 at 15 49 56

...so we can no longer just pass the broken link summary like this:

```ruby
step_config = {
  field: "1",
  value: "1 broken link"
}
```

The component would need to take something like this:

```ruby
step_config = {
  field: step.logic == 'number' ? step_number : step.logic,
  value: "
    <div><svg>...</svg> <span class='error'>1 broken link</span></div>
  ".html_safe
}
```

Ideally this would live in _broken-links-summary.html something like:

```ruby
step_config = {
  field: step.logic == 'number' ? step_number : step.logic,
  value: render 'shared/steps/broken-links-summary', step: step
}
```

But I haven't figured out how to make `render` accessible from the presenter 🤔 any ideas?

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -57,6 +57,36 @@ def summary_list_params
     params
   end
 
+  def steps_settings
```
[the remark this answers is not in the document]
ChrisBAshton: What would be a better name?

`steps_summary_list_params`? `steps_overview_section_params`?

Beginning to think `settings` not a good word choice, as we have 'Sidebar settings' elsewhere. And unfortunately we have a `summary_list_params` method above this, for the overall 'summary' at the top of the page.

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -108,4 +117,27 @@ def last_saved
...
+
+private
+
+  def steps
+    step_number = 0
+    step_by_step_page.steps.map do |step|
+      step_number = step_number + 1 if step.logic == 'number'
+      {
+        field: step.logic == 'number' ? step_number : step.logic,
+        value: "#{step.title} #{broken_link_status_text(step)}", # TODO: render broken-links-summary.html
```
ChrisBAshton: Want to do something like this:

```ruby
value: "#{step.title} #{render 'shared/steps/broken-links-summary', step: step}"
```

...but `render` is not available to the presenter.

app/presenters/step_by_step_page_presenter.rb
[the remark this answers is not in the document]
leenagupte: I don't understand why you need to call render from the presenter.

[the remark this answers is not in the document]
ChrisBAshton: We want the content of `broken-links-summary.html` as our `value`. This is what I was hoping to pair on.

Yes I think in hindsight the best approach is to put it all in the view 👍 just means all the logic for the `step` method will be in the view too.

app/presenters/step_by_step_page_presenter.rb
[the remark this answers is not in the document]
ChrisBAshton: Have now pulled this back out of the presenter and into the view. 👍

app/presenters/step_by_step_page_presenter.rb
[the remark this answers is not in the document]
ChrisBAshton: Have now pulled this back out of the presenter and into the view.

ChrisBAshton:
Thanks for the review. Agree with the comments and have addressed 👍 

> Should "Add step" be a link that sits next to "Reorder"?
> When seen in the context of the page, the "Check for broken links" button looks a bit odd floating about by itself when the step by step can't be edited.

I'm following the design attached to the Trello card. Personally think the "Add step" button looks good where it is. Agreed "Check for broken links" looks a little odd when in scheduled state, but we need to be able to check for broken links at any time (which is itself a bit of an odd thing - you'd think a step by step that is approved and scheduled had been checked already. But I know this is because links that work today may not necessarily work tomorrow)

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -40,4 +39,46 @@
...
+      }
+    }
+
+    subject { described_class.new(step_nav) }
+
+    before do
+      allow(step).to receive(:link_report?).and_return(true)
+    end
+
+    it "should use singular 'link' if only one link is broken" do
```
brucebolt: I think this test is named incorrectly.  Something like `should pluralise 'links' if no links are broken`.

ChrisBAshton: Whoops, well spotted. A rogue copy & paste. Will fix 👍

brucebolt: That's fine by me

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -57,6 +57,14 @@ def summary_list_params
     params
   end
 
+  def broken_link_status_text(step)
```
leenagupte: This feels like this is in the wrong place. It feels like this should be on the step, no the step-by-step-page.

spec/presenters/step_by_step_page_presenter_spec.rb
```diff
@@ -40,4 +39,46 @@
...
+      allow(step).to receive(:broken_links).and_return([broken_link])
+
+      expect(subject.broken_link_status_text(step)).to eq "1 broken link found"
+    end
+
+    it "should pluralise 'links' if multiple links are broken" do
+      allow(step).to receive(:broken_links).and_return([broken_link, broken_link])
```
leenagupte: The content of the broken links doesn't really matter for this test, so is `broken_link` needed, or can `broken_links` just return strings?

app/views/shared/steps/_broken-links-summary.html.erb
```diff
@@ -1,5 +1,8 @@
 <% if step.link_report? %>
-  <span class="govuk-tag govuk-tag--<%= step.broken_links? ? 'warning' : 'success' %>">
-    <%= @step_by_step_page_presenter.broken_link_status_text(step) %>
-  </span>
+  <div class="step-broken-links__summary">
+    <span class="app-broken-link-status app-broken-link-status--<%= step.broken_links? ? 'fail' : 'pass' %>">
+      <%= @step_by_step_page_presenter.broken_link_status_icon(step) %>
```
leenagupte: It feels strange to be calling `@step_by_step_page_presenter` here as we're in an individual step.

leenagupte: As the icon is static, perhaps instead this can be:

```
<% if step.broken_links? %>
  render "cross.svg"
<%  else %> 
  render "tick.svg"
<% end %>
```

I think for this to work, `cross.svg` and `tick.svg` need to be moved somewhere accessible by the view.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -20,57 +20,37 @@
...
-        <% end %>
-      </tbody>
-    </table>
+    <%
+      step_number = 0
+      steps = @step_by_step_page.steps.map do |step|
+        step_number = step_number + 1 if step.logic == 'number'
+        {
+          field: step.logic == 'number' ? step_number : step.logic,
+          value: "#{step.title} #{render 'shared/steps/broken-links-summary', step: step}".html_safe,
```
leenagupte: What is `value` supposed to be? Why is "render" being called here?

leenagupte: It feels like we're trying to force summary-list component to act in a way it's not designed to. Is there scope for updating summary list to accept an html block, similar to inset-text?
cc @alex-ju

app/views/step_by_step_pages/show.html.erb
```diff
@@ -59,7 +59,9 @@
             href: new_step_by_step_page_step_path(@step_by_step_page)
           } %>
         <% end %>
-        <%= button_to "Check for broken links", { action: "check_links", step_by_step_page_id: @step_by_step_page.id}, method: :post, class: "govuk-button gem-c-button gem-c-button--inline gem-c-button--secondary-quiet" %>
+        <% if @step_by_step_page.steps.present? %>
```
leenagupte: Nitpick: This check could be `.any?`

app/views/shared/steps/_broken-links-summary.html.erb
leenagupte: I've got this working:

```
<% if step.link_report? %>
  <div class="step-broken-links__summary">
    <% if step.broken_links? %>
      <span class="app-broken-link-status app-broken-link-status--fail">
        <%= render "svgs/cross.svg" %>
        <%= pluralize(step.broken_links.count, "broken link") %> found
      </span>
    <% else %>
      <span class="app-broken-link-status app-broken-link-status--pass">
        <%= render "svgs/tick.svg" %>
        No broken links
      </span>
    <% end %>
  </div>
<% end %>
```

The svg files are in `app/views/svgs` they are being rendered as partials, so they need a `_` prepended to the file name, e.g. `_cross.svg`

app/views/step_by_step_pages/show.html.erb
leenagupte: After talking IRL, we've decided to:

1. Move `step.title` to the `broken-links-summary` partial,
2. Stop rendering `broken-links-summary` on the Step page as the text is very small and doesn't seem to be adding much value. If there are broken links, the broken link report is rendered, so the users are aware if something needs fixing.

This will allow `value` to be set to: `render('shared/steps/broken-links-summary', step: step)`

leenagupte: I think `_broken-links-summary` should probably be renamed to `_summary` now as that's a more accurate description of what it is.

PR #709, opened 2019-08-29
Add 'Sidebar settings' and 'Secondary links' sections

ChrisBAshton:
Trello cards: https://trello.com/c/h6Oj4hno/46-move-sidebar-settings-and-secondary-links-inside-the-summary-component

Dependent on https://github.com/alphagov/collections-publisher/pull/703

## What

1. Uses summary_list component to render new sections on the 'step by step' admin page:
   - 'Sidebar settings' (and Edit/View link)
   - 'Secondary links' (and Edit/View link), with guidance content taken from the secondary links page
2. Adds tests to verify that the correct content is appearing on this page.

Here are the additional sections added to the bottom of the page:

------

Screen Shot 2019-08-29 at 17 00 14

### DIFF
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -44,6 +44,7 @@ RSpec.feature "Managing step by step pages" do
     when_I_view_the_step_by_step_page
     then_I_can_see_a_summary_section
     and_I_can_edit_the_summary_section
+    and_I_can_see_a_sidebar_settings_section_with_link "Edit"
     and_I_can_see_a_metadata_section
   end
 
@@ -173,6 +174,7 @@ RSpec.feature "Managing step by step pages" do
       when_I_view_the_step_by_step_page
       then_I_can_see_a_summary_section
       but_I_cannot_edit_the_summary_section
+      and_I_can_see_a_sidebar_settings_section_with_link "View"
       then_I_can_preview_the_step_by_step
       and_the_steps_can_be_checked_for_broken_links
     end
@@ -337,6 +339,13 @@ RSpec.feature "Managing step by step pages" do
     expect(page).to have_css(".gem-c-summary-list#content")
     within(".gem-c-summary-list#content") do
       yield
+    end
+  end
+
+  def and_I_can_see_a_sidebar_settings_section_with_link(link_text)
+    within(".gem-c-summary-list#sidebar-settings") do
+      expect(page).to have_content "Sidebar settings"
+      expect(page).to have_link(link_text, :href => step_by_step_page_navigation_rules_path(@step_by_step_page))
     end
   end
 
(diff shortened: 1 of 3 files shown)


ChrisBAshton:
@leenagupte thanks for the approval. I've added a feature to conditionally display 'Edit' or 'View' depending on the editable status of the step by step. Could you re-review please?

As before, this is branched off `update-govuk_publishing_components` as it needs the latest gem, but I intend for it to be approved & merged separately (I'll change the base branch of the PR to `master` once the parent branch has been merged).

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -44,6 +44,7 @@
     when_I_view_the_step_by_step_page
     then_I_can_see_a_summary_section
     and_I_can_edit_the_summary_section
+    and_I_can_see_a_sidebar_settings_section_with_link "Edit"
```
leenagupte: I think there should be a step after this that checks that when the "Edit" link is clicked, the user is taken to the "Choose Navigation" page, so we can be sure that we're linking to the right place.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -45,6 +45,7 @@
     then_I_can_see_a_summary_section
     and_I_can_edit_the_summary_section
     and_I_can_see_a_sidebar_settings_section_with_link "Edit"
+    and_I_can_see_a_secondary_links_section_with_link "Edit"
```
leenagupte: Same here, I think there should be a step after this that checks that when the "Edit" link is clicked, the user is taken to the "Secondary content" page, so we can be sure that we're linking to the right place.

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: 👍 I'll add that as an expectation inside `and_I_can_see_a_sidebar_settings_section_with_link` (don't want to bloat the scenario too much)

ChrisBAshton: Hmmm I'm already asserting that the link has the right URL:

`expect(page).to have_link(link_text, :href => step_by_step_page_navigation_rules_path(@step_by_step_page))`

I want to test that the link text says "Edit" or "View" (unlike the `and_I_can_edit_the_summary_section` above, which should only have "Edit" or nothing at all). So we'd get into a situation where when it's editable, we write:

```
and_I_can_see_a_sidebar_settings_section
and_it_should_have_a_link "Edit"
and_it_should_take_me_to_the_sidebar_settings_page
```

vs when it's not editable

```
and_I_can_see_a_sidebar_settings_section
and_it_should_have_a_link "View"
and_it_should_take_me_to_the_sidebar_settings_page
but_I_should_not_be_able_to_edit_it
# which duplicates some of our other tests
```

I think what we have already is less verbose and easier to understand.

ChrisBAshton:
I've reworded the first commit 👍

ChrisBAshton:
Have reworded it again 👍

PR #713, opened 2019-09-02
Store the status of a step by step

leenagupte:
Trello: https://trello.com/c/0H0pphSZ
Implements: PR #714 

## What's changed?
Creates a new field to store the status of a step by step. The field is validated against the current list of possible statuses:

- draft
- published
- scheduled

It also removes "unpublished_changes" as a status as, in terms of workflow, this is the same as being in "draft".

The "live" status has also been renamed to "published" to to be consistent with publishing-api the other publishing apps.

A rake task has been created to populate the current status of all existing step by steps.

## Why
The "status" of a step by step was calculated from data about the step by step, e.g. if there is a draft_updated_at date but not a published_at date, then it's status was a draft etc.

That worked until now, when we only had 4 statuses:

- Draft
- Live (published)
- Scheduled
- Unpublished changes

To add the 2i workflow, we'll need at least three more statuses:

- Submitted for 2i
- In review
- 2i approved

Which will make calculating them very complicated.

## Post deployment task

Run task: `rake step_by_step:set_status`


app/models/step_by_step_page.rb
```diff
@@ -15,6 +15,7 @@ class StepByStepPage < ApplicationRecord
   validates :slug, format: { with: /\A([a-z0-9]+-)*[a-z0-9]+\z/ }, uniqueness: true
   validates :slug, slug: true, on: :create
   validates :status, inclusion: { in: STATUSES }, presence: true
+  validates :status, status: true
```
erino: What about renaming the validator as it took me quite a while to realise that the second `:status` was the name of the validator?

app/models/step_by_step_page.rb
```diff
@@ -30,7 +30,7 @@ def has_draft?
   end
 
   def scheduled_for_publishing?
-    has_draft? && scheduled_at.present?
+    has_draft? && scheduled_at.present? && status == "scheduled"
```
erino: What about using a [`StringInquirer`](https://api.rubyonrails.org/v5.2/classes/ActiveSupport/StringInquirer.html) to turn status into methods? such as `status.scheduled?`

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -9,7 +9,7 @@ def initialize(step_by_step_page)
 
   def summary_metadata
     items = {
-      "Status" => step_by_step_page.status[:text],
+      "Status" => step_by_step_page.status.humanize,
```
erino: Is the capitalisation of `unpublished_changes` to "Unpublished changes" correct? or should it be "Unpublished Changes"

app/views/step_by_step_pages/index.html.erb
```diff
@@ -132,7 +132,7 @@
               <td class="govuk-table__cell">
                 <% if step_by_step_page.has_draft? %>
                   <%= link_to "/#{step_by_step_page.slug}", preview_url(step_by_step_page.slug, auth_bypass_id: step_by_step_page.auth_bypass_id), target: "_blank" %>
-                <% elsif step_by_step_page.status == "live" %>
+                <% elsif step_by_step_page.status == "published" %>
                   <%= link_to "/#{step_by_step_page.slug}", live_url(step_by_step_page.slug), target: "_blank" %>
```
erino: What about changing `live_url` to `public_url`? as `live` is being purged from the codebase.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -51,35 +51,7 @@
...
-              },
-              {
-                text: "Scheduled",
-                value: "scheduled",
-                data_attributes: {
-                  show: "scheduled"
-                },
-              }
-            ].each{ |item| item[:selected] = item[:value] == params[:status] }
+            options: StepByStepFilter::Options.statuses.each{ |item| item[:selected] = item[:value] == params[:status] }
```
ChrisBAshton: Nice!

Think it should have a space between `each` and `{`.

spec/validators/status_validator_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      step_by_step_page.scheduled_at = nil
+      step_by_step_page.status = "scheduled"
+
+      expect(step_by_step_page).not_to be_valid
+      expect(step_by_step_page.errors.messages[:status]).to eq([error_message])
+    end
+
+    it "does not allow status to be scheduled if there is no draft" do
+      allow(step_by_step_page).to receive(:has_draft?).and_return(false)
+      step_by_step_page.scheduled_at = Time.zone.now + 1.hour
```
erino: `1.hour.since` is a bit more Railsy

app/models/step_by_step_page.rb
```diff
@@ -49,6 +49,7 @@ def mark_as_published
     update_attribute(:draft_updated_at, now)
     update_attribute(:scheduled_at, nil)
     update_attribute(:assigned_to, nil)
+    update_attribute(:status, "published")
```
erino: Can all of these (and others from other methods) calls to `update_attribute` be rewritten one call to `update { published_at: now, draft_updated_at: now ... }`  saving a few database calls

app/services/step_by_step_page_reverter.rb
```diff
@@ -12,6 +12,7 @@ def repopulate_from_publishing_api
     step_by_step_page.introduction = introduction
     step_by_step_page.description = payload_from_publishing_api[:description]
     step_by_step_page.draft_updated_at = step_by_step_page.published_at
+    step_by_step_page.status = "published"
```
erino: Can all of these updates be replaced with a single call to `update`?

app/views/step_by_step_pages/index.html.erb
erino: and what moving this to a method in `StepByStepFilter::Options` to keep the view cleaner?

leenagupte: @erino That's an interesting idea. I'll investigate how to do that.

app/presenters/step_by_step_page_presenter.rb
leenagupte: The [original text](https://github.com/alphagov/collections-publisher/blob/master/app/views/step_by_step_pages/index.html.erb#L83) is "Unpublished changes", so I assumed that's it's alright.

app/services/step_by_step_page_reverter.rb
leenagupte: Isn't that what happens when `save!` is called?

app/models/step_by_step_page.rb
```diff
@@ -53,7 +53,7 @@ def self.validate_redirect(redirect_url)
 
   def status
     return "scheduled" if scheduled_for_publishing?
-    return "unpublished_changes" if unpublished_changes?
+    return "draft" if has_draft?
```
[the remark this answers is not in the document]
leenagupte: This shouldn't be a problem as it should be picked up by the tests.

spec/services/step_by_step_filter_spec.rb
```diff
@@ -40,16 +32,6 @@
...
 
-    it "returns a list of step by steps with a status of published" do
-      filter_params = {
-        status: "published"
-      }
-      results = described_class.new(filter_params).results
-
-      expect(results.count).to eq(1)
-      expect(results.first.title).to eq(published_step_by_step.title)
-    end
```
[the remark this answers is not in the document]
leenagupte: > The Filter specs were using `unpublished_changes` to check for what happens when filtering by a
status with nothing assigned. This has been changed to use `published` instead as that required the
least amount of test changes.

To do that `published_step_by_step_page` had to be removed. So this test will never pass.

app/models/step_by_step_page.rb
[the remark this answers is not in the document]
leenagupte: I do that in a later commit.

app/models/step_by_step_page.rb
```diff
@@ -52,6 +52,10 @@ def mark_as_published
   def mark_as_unpublished
     update_attribute(:published_at, nil)
     update_attribute(:draft_updated_at, nil)
+
+  def mark_as_scheduled
+    update_attribute(:status, "scheduled")
+  end
```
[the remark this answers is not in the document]
leenagupte: Ah, the `end` for `mark_as_unpublished` is in the wrong place. Will fix.

app/presenters/step_by_step_page_presenter.rb
erino: 👍

app/services/step_by_step_page_reverter.rb
erino: I mean, could the entire methods be rewritten like this?

```ruby 
def repopulate_from_publishing_api
  update(
    title: payload_from_publishing_api[:title],
    slug: payload_from_publishing_api[:base_path].tr('/', ''),
    introduction: introduction,
    description: payload_from_publishing_api[:description],
    draft_updated_at: step_by_step_page.published_at,
    status: "published"
  )
end
```

spec/services/step_by_step_page_reverter_spec.rb
```diff
@@ -68,6 +68,10 @@
       expect(step_by_step.draft_updated_at).to eq(published_at)
     end
 
+    it "sets the status to published" do
+      expect(step_by_step_page.status).to eq("published")
```
erino: If `StringInquirer` is used for status, then I think this spec could use a predicate matcher

```ruby
expect(step_by_step_page.status).to be_published
```

app/models/step_by_step_page.rb
leenagupte: I've renamed the validator to `StatusPrerequisiteValidator`.

app/models/step_by_step_page.rb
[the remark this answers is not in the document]
leenagupte: Updated and fixed up.

app/models/step_by_step_page.rb
leenagupte: Updated and fixed-up

app/models/step_by_step_page.rb
leenagupte: Updated and fixed-up

app/views/step_by_step_pages/index.html.erb
leenagupte: Updated and fixed-up.

app/models/step_by_step_page.rb
leenagupte: Updated and fixed-up

spec/validators/status_validator_spec.rb
leenagupte: Updated and fixed-up

app/services/step_by_step_page_reverter.rb
leenagupte: Update and fixed-up

spec/services/step_by_step_page_reverter_spec.rb
leenagupte: Update and fixed-up

app/views/step_by_step_pages/index.html.erb
leenagupte: @erino Refactored and fixed-up

PR #716, opened 2019-09-03
Random clean up

leenagupte:
Removes a rogue "binding.pry" and makes one of the tests more consistent


ChrisBAshton:
(Good spot on the tests!)

PR #721, opened 2019-09-05
Update index page

alex-ju:
This PR updates the summary page as follows:
- Update the heading and the "Create new" button
- Splits the index page into `_filters` and `_results` for maintainability and consistency with other publishing apps
- Updates the header of the page to use the admin_layout yield blocks and show the add button next to the title
- Updates the results table to use the table helper and document list component to generate the results consistently across publishing apps.

<table>
<tr><td valign="top">

index-before

</td><td valign="top">

index-after

</td></tr>
</table>

co-author: @AgaDufrat

[Trello card](https://trello.com/c/C4HHCrAX)

### DIFF
--- a/app/assets/stylesheets/objects/_grid.scss
+++ b/app/assets/stylesheets/objects/_grid.scss
@@ -1,0 +1,3 @@
+.app-grid-column--align-right {
+  text-align: right;
+}
--- a/app/helpers/time_options_helper.rb
+++ b/app/helpers/time_options_helper.rb
@@ -20,6 +20,10 @@ module TimeOptionsHelper
     datetime.strftime('%-l:%M%P on %-d %B %Y')
   end
 
+  def format_full_date(datetime)
+    datetime.strftime('%-d %B %Y')
+  end
+
   def default_datetime_placeholder(
     year: 1.day.from_now.year,
     month: 1.day.from_now.month,
--- a/app/views/step_by_step_pages/index.html.erb
+++ b/app/views/step_by_step_pages/index.html.erb
@@ -6,24 +6,18 @@
   ]
 %>
 
-<%= render 'shared/steps/page_title', links: links %>
+<% content_for :title, render("govuk_publishing_components/components/heading", {
+  text: "Step by step navigation publisher",
+  margin_bottom: 6
+}) %>
 
 <%= render 'shared/steps/form_notice', message: notice, status: "success" %>
 
-
-<%= render "govuk_publishing_components/components/heading", {
-  text: "Step by step navigation publisher",
-  margin_bottom: 6
-} %>
-
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-    <%= render "govuk_publishing_components/components/button", {
-      text: "Create new",
-      href: new_step_by_step_page_path
-    } %>
-  </div>
-</div>
+<% content_for :title_side, render("govuk_publishing_components/components/button", {
+  text: "Create new step by step",
+  href: new_step_by_step_page_path,
+  margin_bottom: true
+}) %>
 
 <div class="govuk-grid-row govuk-!-margin-top-5">
   <div class="govuk-grid-column-one-quarter">
--- a/app/views/step_by_step_pages/index.html.erb
+++ b/app/views/step_by_step_pages/index.html.erb
@@ -6,12 +6,9 @@
   ]
 %>
 
-<% content_for :title, render("govuk_publishing_components/components/heading", {
-  text: "Step by step navigation publisher",
-  margin_bottom: 6
-}) %>
-
 <%= render 'shared/steps/form_notice', message: notice, status: "success" %>
+
+<% content_for :title, "Step by steps" %>
 
 <% content_for :title_side, render("govuk_publishing_components/components/button", {
   text: "Create new step by step",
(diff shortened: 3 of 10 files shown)


app/views/step_by_step_pages/index/_results.erb
```diff
@@ -1,41 +1,42 @@
...
-              <%= link_to "/#{step_by_step_page.slug}", live_url(step_by_step_page.slug), target: "_blank" %>
-            <% else %>
-              /<%= step_by_step_page.slug %>
-            <% end %>
-          </td>
-        </tr>
+  <%= t.head do %>
+    <%= t.header "Title" %>
+    <%= t.header "Status" %>
+    <%= t.header "Last updated" %>
```
ChrisBAshton: This is a different heading, probably worth clarifying that in this commit (or better yet, splitting it out into a separate commit).

app/views/step_by_step_pages/index/_results.erb
```diff
@@ -1,41 +1,42 @@
...
+              },
+              metadata: {
+                preview_link: preview_link
+              }
+            }
+          ]
+        } %>
+
+        <%= t.cell step_by_step_page.status.humanize %>
+        <%= t.cell step_by_step_page.created_at.strftime('%-d %B %Y') %>
```
ChrisBAshton: Would probably be best to create a new helper method in TimeOptionsHelper and call that from here. There's a [`format_full_date_and_time`](https://github.com/alphagov/collections-publisher/blob/f8401b53dd3f48766d63dc4b9f050a856d190fa7/app/helpers/time_options_helper.rb#L19-L21) you can use for inspiration, probably call this one `format_full_date` or something?

app/views/step_by_step_pages/index/_results.erb
```diff
@@ -1,41 +1,42 @@
...
+
+      <% preview_link = capture do %>
+        <% if step_by_step_page.has_draft? %>
+          <%= link_to "/#{step_by_step_page.slug}", preview_url(step_by_step_page.slug, auth_bypass_id: step_by_step_page.auth_bypass_id), target: "_blank", class: "govuk-link" %>
+        <% elsif step_by_step_page.status.published? %>
+          <%= link_to "/#{step_by_step_page.slug}", published_url(step_by_step_page.slug), target: "_blank", class: "govuk-link" %>
+        <% else %>
+          /<%= step_by_step_page.slug %>
+        <% end %>
+      <% end %>
```
ChrisBAshton: I wonder if this can be refactored? I'd be tempted to move the preview URL logic to a presenter so we can call it like:

```ruby
<%= link_to "/<%= step_by_step_page.slug %>", preview_link_url, target: "_blank", class: "govuk-link" %>
```

app/views/step_by_step_pages/index/_results.erb
```diff
@@ -0,0 +1,41 @@
...
+
+  <tbody class="govuk-table__body">
+    <% if @step_by_step_pages.present? %>
+      <% @step_by_step_pages.each do |step_by_step_page| %>
+        <tr class="govuk-table__row" data-status="<%= step_by_step_page.status[:name] %>">
+          <td class="govuk-table__cell">
+            <strong><%= link_to step_by_step_page.title, step_by_step_page %></strong>
+          </td>
+          <td class="govuk-table__cell">
+            <span class="h4">
```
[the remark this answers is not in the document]
alex-ju: I copied this to split the index page. Markup is was so bad that I don't even know where to start 😄 . It was updated in the following commit to use the table helper which generates nice semantic elements and classes.

app/views/step_by_step_pages/index/_results.erb
```diff
@@ -0,0 +1,41 @@
...
+              <span class="label <%= step_by_step_page.status[:label_class] %>" title="<%= step_by_step_page.status[:text] %>"><%= step_by_step_page.status[:text] %></span>
+            </span>
+          </td>
+          <td class="govuk-table__cell">
+            <% if step_by_step_page.has_draft? %>
+              <%= link_to "/#{step_by_step_page.slug}", preview_url(step_by_step_page.slug, auth_bypass_id: step_by_step_page.auth_bypass_id), target: "_blank" %>
+            <% elsif step_by_step_page.status[:name] == "live" %>
```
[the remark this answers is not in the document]
alex-ju: This was removed in the next commit after rebase as [it's replaced with `step_by_step_page.status.published?`](https://github.com/alphagov/collections-publisher/pull/713/files#diff-3ee31a41ddc772d410b5bcef3348884dR100). Rebase made it hard for this PR to be reviewed unfortunately as the template split commit preserved the old code 😞.

app/views/step_by_step_pages/index.html.erb
```diff
@@ -6,107 +6,35 @@
...
-            <td class="govuk-table__cell" colspan="3">
-              <p class="govuk-hint">No step by steps have been created yet.</p>
-            </tr>
-          </tr>
-        <% end %>
-      </tbody>
-    </table>
+    <% if @step_by_step_pages.none? %>
+      <div class="govuk-body">
+        <p>Improve your search results:</p>
```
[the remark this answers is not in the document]
alex-ju: Good suggestion, copied over from the other publisher. 👍

app/assets/stylesheets/objects/_grid.scss
```diff
@@ -0,0 +1,3 @@
+.app-grid-column--align-right {
+  text-align: right;
```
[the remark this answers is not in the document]
alex-ju: We have the numeric feature which automatically aligns cells to the right, but also applies the font for tabular numbers - which I wasn't sure we want for the dates.

app/views/step_by_step_pages/index/_results.erb
alex-ju: Updated to use the time_options helper

app/views/step_by_step_pages/index/_results.erb
alex-ju: Mm.. the use of presenters seems to be ambiguous at the team level (or even wider) so I'd say this refactor suggestions is not in the scope of this PR 😁 .
I agree there are means of making this cleaner, but since it wasn't introduced with this PR (but ported from the `index.html.erb` file, I'm tempted to leave it as is.

alex-ju:
@ChrisBAshton As you've probably figured by the end of the review the 'live' reference was present in a intermediate commit which wasn't updated at rebase (the joys of splitting files into partials). I tried to address all the comments except for one refactor suggestions which I consider out of scope for this PR. I should've suggested reviewing files instead of commits as it's much easier to get over the split into partials bit.

app/helpers/time_options_helper.rb
```diff
@@ -20,6 +20,10 @@ def format_full_date_and_time(datetime)
     datetime.strftime('%-l:%M%P on %-d %B %Y')
   end
 
+  def format_full_date(datetime)
```
ChrisBAshton: Could you add some tests for this method? (There are equivalents for format_full_date_and_time)

alex-ju: Sure, have added now.

app/views/step_by_step_pages/index/_results.erb
leenagupte: I agree with this. Can you add a line to the commit message about why this changed.

spec/helpers/time_options_helper_spec.rb
```diff
@@ -49,6 +49,13 @@
     end
   end
 
+  describe "#format_full_date" do
+    it "formats the date correctly" do
+      date = Time.new(2030, 4, 20, 10, 26, 0, '+01:00') # London timezone
```
ChrisBAshton: Can probably simplify to:

```suggestion
      date = Time.new(2030, 4, 20)
```

As we're only concerned with the date. Should at least be able to remove the London timezone comment and `+01:00` modifier 👍

PR #723, opened 2019-09-06
Clarify major and minor updates

AgaDufrat:
**What**
- View is updated in line with the [card](https://trello.com/c/ZB4lH4G8/28-publish-delete-tab-clarify-what-major-and-minor-updates-mean) description
- References to minor/major are removed from UI

**Why**
Content designers were not sure what does the minor and major update mean.

**Example**
change_note

### DIFF
--- a/app/views/step_by_step_pages/_change_notes.html.erb
+++ b/app/views/step_by_step_pages/_change_notes.html.erb
@@ -1,0 +1,32 @@
+<% textarea = capture do %>
+    <%= render "govuk_publishing_components/components/textarea", {
+      label: {
+        text: "Public change note",
+      },
+      name: "change_note",
+      rows: 4,
+      hint: "Describe the change for users. This change note will be emailed to subscribers."
+    } %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds" id="change-note">
+    <%= render "govuk_publishing_components/components/radio", {
+      heading: "Notify users about this change?",
+      name: "update_type",
+      id: "update-type",
+      items: [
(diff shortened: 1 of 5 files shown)


app/views/step_by_step_pages/_change_notes.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    } %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds" id="change-note">
+    <% legend_text = tag.p(
+      tag.strong("Notify users about this change?"))
+    %>
+    <%= render "govuk_publishing_components/components/fieldset", legend_text: legend_text do %>
+      <%= render "govuk_publishing_components/components/radio", {
```
alex-ju: The radio component already generates the markup for fieldset (as it always contains more than one input and they need to be grouped). I suggest you remove the wrapping fieldset component from L17 and pass its content as a 'heading' on the radios component (which gets rendered as `<legend>`: `heading: "Notify users about this change?"`).

AgaDufrat: Thanks! I was wondering about that but was worried that `heading` won't be enough :)
I will also need to resolve headers in the Scheduling page. What would be the best way to do so? Make "Date" bigger?
Screen Shot 2019-09-09 at 09 38 43

alex-ju: As discussed. I'm aware of the differences in font-size between labels and legends but that needs to be fixed in the component (or coped with with an app-level overwrite) so I'll treat it separately.

PR #728, opened 2019-09-10
Add submit for 2i page

leenagupte:
Trello: https://trello.com/c/NO5JO4Q1

## What's changed?
Adds a form to allow the user to submit a step by step for 2i, with additional notes for the reviewer.

The status of the step-by-step changes to to `submitted_for_2i`.

The name of the submitter is also being stored, so that in a future change, we can ensure that then same user can't review the step-by-step.

Access to the Submit for 2i form is limited to users who have "Unreleased feature" permissions until the whole 2i feature is ready to be deployed. 

## Why?
This will ensure that reviewers know when a step-by-step is ready to be reviewed and that they can see a pre-filtered list of every thing that is ready for review.

### Submit for 2i page
Screen Shot 2019-09-10 at 16 30 15

### Success message
Screen Shot 2019-09-10 at 16 31 02

### Changes notes
Screen Shot 2019-09-10 at 16 31 12

### Filterable status
Screen Shot 2019-09-10 at 16 31 28

## Note to reviewer
At the moment the "Submit for 2i" form can only be accessed by hacking the URL, and going to the route directly: 
`http://collections-publisher.dev.gov.uk/step-by-step-pages/1/submit-for-2i`

This is because the overview page is currently being drafted, but when it's ready, a "Submit for 2i" button will appear in the "Action" box.


config/routes.rb
```diff
@@ -4,22 +4,22 @@
...
+    post :check_links
+    get 'internal-change-notes', to: 'interal_change_notes'
+    post 'internal-change-notes', to: 'internal_change_notes#create'
     get 'navigation-rules', to: 'navigation_rules#edit'
     put 'navigation-rules', to: 'navigation_rules#update'
     get :publish
     post :publish
-    get :schedule
-    post :schedule
+    get  'publish-or-delete', to: 'publish_or_delete'
```
brucebolt: You've got two spaces between the `get` and `'`, but one on the other lines

app/validators/status_prerequisite_validator.rb
```diff
@@ -1,15 +1,25 @@
...
     end
   end
+
+private
+
+  def can_be_scheduled?(record)
+    record.has_draft? && record.scheduled_at.present?
+  end
+
+  def can_be_summited_for_2i?(record)
```
brucebolt: There's a typo in the name of this method

app/controllers/review_controller.rb
```diff
@@ -5,7 +5,19 @@ class ReviewController < ApplicationController
   before_action :require_unreleased_feature_permissions!
   before_action :set_step_by_step_page
 
-  def submit_for_2i; end
+  def submit_for_2i
+    if request.post?
+      if @step_by_step_page.update(
+        review_requester: current_user.name,
```
brucebolt: Instead of storing the actual name of the requester, would it be cleaner to store the user's ID, then retrieve their name from the `users` table when displayed?

It would also future-proof this, e.g. if we ever wanted to have the email address of the requester displayed.

app/validators/status_prerequisite_validator.rb
leenagupte: I cannot spell "submitted"!

app/controllers/review_controller.rb
leenagupte: It would. I was following the pattern that's used for setting "assigned_to", though that's no reason not to make it better here.

brucebolt: Looks like the same pattern has been used previously in the change notes too.  It's not something I'm massively keen on since it feels like we're duplicating data (and then showing out of date names when someone's name has changed).

db/schema.rb
```diff
@@ -109,6 +109,7 @@
     t.string "status", null: false
     t.string "review_requester"
     t.index ["content_id"], name: "index_step_by_step_pages_on_content_id", unique: true
+    t.index ["review_requester"], name: "fk_rails_91a467475a"
```
ChrisBAshton: Where does this value come from? It looks automatically generated, unlike anything else in this file.

app/controllers/review_controller.rb
```diff
@@ -10,7 +10,7 @@ def submit_for_2i
       status = "submitted_for_2i"
 
       if @step_by_step_page.update(
-        review_requester: current_user.name,
+        review_requester: current_user.uid,
```
ChrisBAshton: Is it worth fixing this up to the `Set status to submitted_for_2i when form submitted` commit? I can't think of any value in retaining the history that you temporarily opted for `name` before refactoring to `uid`.
Probably worth describing in the commit why `uid` was chosen too.

db/migrate/20190909151119_add_review_requester_to_step_by_step_pages.rb
```diff
@@ -0,0 +1,5 @@
+class AddReviewRequesterToStepByStepPages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :step_by_step_pages, :review_requester, :string
```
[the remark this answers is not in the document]
leenagupte: No. The uid is stored as a string in the User table, and I'll be adding a foreign key relationship between them.

db/schema.rb
leenagupte: It was autogenerated by the migration to add a foreign key.

app/controllers/review_controller.rb
leenagupte: I'd hadn't finished working through Bruce's comments, when I pushed this change. It can't be fixed up until the migration is as well.

leenagupte:
@brucebolt @ChrisBAshton Thanks for your comments. Fixing the migrations has gotten a little complicated, so I'm going to move to a new branch.

PR #730, opened 2019-09-11
Flag actions to do before publishing step by step

ChrisBAshton:
Trello card: https://trello.com/c/IEGDwZch/49-use-inset-text-component-to-flag-actions-that-need-to-be-done-before-publishing

This PR:

- Uses `inset-text` component to show the important actions a user must take before they can
  publish their step by step.
- Adds new pre-requisite: "Add at least one step"
  Whilst it was never possible to publish an empty step by step (would fail the `has_content?`
  check), it used to only have the generic error message "Every step must have content", which
  doesn't make much sense if you haven't even added a step.
- Adds new pre-requisite: "Check for broken links"
  Cannot publish a step by step without checking for broken links since your last update.
- Displays multiple links if step by step is affected by multiple things. Example:
  One step may have no content (fail # 1) and also the draft may have been updated since you last checked for broken links (fail # 2).

Note that if any links are broken, we should ideally prevent publishing and we should have another inset-text prompt that says links need fixing, but this needs fleshing out (e.g. should there be a prompt for every step where there is a broken link, linking out to that step, or should there be one prompt to cover them all, linking out just to the `#steps` section?). This is out of scope for this PR and will be done in https://trello.com/c/9hBzk8lJ.

Screen Shot 2019-09-13 at 13 29 03

### DIFF
--- a/app/assets/stylesheets/utilities/_overrides.scss
+++ b/app/assets/stylesheets/utilities/_overrides.scss
@@ -102,3 +102,7 @@
   }
 }
 // scss-lint:enable IdSelector
+
+.gem-c-inset-text {
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+}
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -86,7 +86,10 @@ class StepByStepPage < ApplicationRecord
   end
 
   def can_be_published?
-    has_draft? && !scheduled_for_publishing? && steps_have_content? && links_checked?
+    has_draft? &&
+      !scheduled_for_publishing? &&
+      steps_have_content? &&
+      links_checked_since_last_update?
   end
 
   def can_be_unpublished?
@@ -126,6 +129,10 @@ class StepByStepPage < ApplicationRecord
 
   def links_last_checked_date
     steps.map(&:links_last_checked_date).reject(&:blank?).max
+  end
+
+  def links_checked_since_last_update?
+    links_checked? && links_last_checked_date > draft_updated_at
   end
 
   def links_checked?
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -86,7 +86,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def can_be_published?
-    has_draft? && !scheduled_for_publishing? && steps_have_content?
+    has_draft? && !scheduled_for_publishing? && steps_have_content? && links_checked?
   end
 
   def can_be_unpublished?
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -131,6 +131,10 @@ class StepByStepPage < ApplicationRecord
     steps.map(&:links_last_checked_date).reject(&:blank?).max
   end
 
+  def should_show_required_prepublish_actions?
+    has_draft? && !scheduled_for_publishing? && !can_be_published?
+  end
+
   def links_checked_since_last_update?
     links_checked? && links_last_checked_date > draft_updated_at
   end
--- a/app/views/step_by_step_pages/show.html.erb
+++ b/app/views/step_by_step_pages/show.html.erb
@@ -18,6 +18,8 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <%= render "step_by_step_pages/show/required_actions" %>
+
     <%= render "govuk_publishing_components/components/summary_list", @step_by_step_page_presenter.summary_list_params %>
 
     <%
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -26,6 +26,12 @@ FactoryBot.define do
     status { "published" }
   end
 
+  factory :step_by_step_with_unpublished_changes, parent: :step_by_step_page_with_steps do
+    published_at { 3.hours.ago }
+    draft_updated_at { 2.hours.ago }
+    status { "draft" }
+  end
+
   factory :scheduled_step_by_step_page, parent: :step_by_step_page_with_steps do
     scheduled_at { 3.hours.from_now }
     status { "scheduled" }
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -13,7 +13,7 @@ FactoryBot.define do
     status { "draft" }
     draft_updated_at { 3.hours.ago }
 
-    factory :step_by_step_page_with_steps do
+    factory :step_by_step_page_with_steps, aliases: [:draft_step_by_step_page] do
       after(:create) do |step_by_step_page|
         create(:step, step_by_step_page: step_by_step_page)
         create(:or_step, step_by_step_page: step_by_step_page)
@@ -26,13 +26,7 @@ FactoryBot.define do
     status { "published" }
   end
 
-  factory :draft_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 1.day.ago }
-    status { "draft" }
-  end
-
   factory :scheduled_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 3.hours.ago }
     scheduled_at { 3.hours.from_now }
     status { "scheduled" }
   end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -11,6 +11,7 @@ FactoryBot.define do
     introduction { "Find out the steps to become amazing" }
     description { "How to be amazing - find out the steps to become amazing" }
     status { "draft" }
+    draft_updated_at { 3.hours.ago }
 
     factory :step_by_step_page_with_steps do
       after(:create) do |step_by_step_page|
@@ -21,7 +22,6 @@ FactoryBot.define do
   end
 
   factory :published_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 3.hours.ago }
     published_at { 3.hours.ago }
     status { "published" }
   end
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -291,6 +291,15 @@ RSpec.feature "Managing step by step pages" do
     and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
+  scenario "Multiple required actions are listed in the inset prompt" do
+    given_a_step_by_step_has_an_empty_step_added_after_links_last_checked
+    when_I_view_the_step_by_step_page
+    then_I_should_see_an_inset_prompt
+    and_the_prompt_should_contain_link_to_steps_section "Add content to all your steps"
+    and_the_prompt_should_contain_link_to_steps_section "Check for broken links"
+    and_I_cannot_publish_or_schedule_the_step_by_step
+  end
+
   def and_it_has_change_notes
     create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
   end
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -286,6 +286,15 @@ RSpec.feature "Managing step by step pages" do
     and_there_should_be_no_schedule_button
   end
 
+  scenario "Step by step has been updated since it was last checked for broken links" do
+    given_a_step_by_step_has_been_updated_after_links_last_checked
+    when_I_view_the_step_by_step_page
+    then_I_should_see_an_inset_prompt
+    and_the_prompt_should_contain_link_to_steps_section "Check for broken links"
+    and_there_should_be_no_publish_button
+    and_there_should_be_no_schedule_button
+  end
+
   def and_it_has_change_notes
     create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
   end
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -238,6 +238,8 @@ RSpec.feature "Managing step by step pages" do
     given_there_is_a_step_by_step_page_with_steps_missing_content
     when_I_view_the_step_by_step_page
     then_I_should_see "Step by steps cannot be published until all steps have content."
+    and_I_should_see_an_inset_prompt
+    and_the_prompt_should_contain "Add content to all your steps"
     and_there_should_be_no_publish_button
     and_there_should_be_no_schedule_button
   end
@@ -622,6 +624,16 @@ RSpec.feature "Managing step by step pages" do
   end
 
   alias_method :then_there_should_be_no_schedule_button, :and_there_should_be_no_schedule_button
+
+  def and_I_should_see_an_inset_prompt
+    expect(page).to have_css('.govuk-inset-text', text: "To publish this step by step you need to")
+  end
+
+  def and_the_prompt_should_contain(prompt)
+    within('.govuk-inset-text') do
+      expect(page).to have_content prompt
+    end
+  end
 
   def then_I_should_see(content, scope = nil)
     scope_selector = case scope
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -300,6 +300,15 @@ RSpec.feature "Managing step by step pages" do
     and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
+  scenario "Publishing/scheduling a step by step does not prompt to check for broken links" do
+    given_there_is_a_step_that_has_no_broken_links
+    when_I_view_the_step_by_step_page
+    then_there_should_be_no_inset_prompt
+    and_when_I_click_button "Publish"
+    then_I_am_told_that_it_is_published
+    and_there_should_continue_to_be_no_inset_prompt
+  end
+
   def and_it_has_change_notes
     create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
   end
@@ -660,6 +669,12 @@ RSpec.feature "Managing step by step pages" do
   end
 
   alias_method :then_I_should_see_an_inset_prompt, :and_I_should_see_an_inset_prompt
+
+  def then_there_should_be_no_inset_prompt
+    expect(page).not_to have_css('.govuk-inset-text', text: "To publish this step by step you need to")
+  end
+
+  alias_method :and_there_should_continue_to_be_no_inset_prompt, :then_there_should_be_no_inset_prompt
 
   def and_the_prompt_should_contain_link_to_steps_section(prompt)
     within('.govuk-inset-text') do
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -85,9 +85,9 @@ RSpec.feature "Managing step by step pages" do
   end
 
   scenario "User publishes a page" do
-    given_there_is_a_draft_step_by_step_page
+    given_there_is_a_step_by_step_page_with_a_link_report
     when_I_view_the_step_by_step_page
-    and_when_I_click_button "Publish"
+    when_I_click_button "Publish"
     then_I_am_told_that_it_is_published
     then_I_see_the_step_by_step_page
     and_I_see_an_unpublish_button
@@ -151,7 +151,7 @@ RSpec.feature "Managing step by step pages" do
   end
 
   scenario "User publishes changes to a published step by step page" do
-    given_there_is_a_published_step_by_step_page_with_unpublished_changes
+    given_there_is_a_step_by_step_page_with_unpublished_changes_whose_links_have_been_checked
     when_I_view_the_step_by_step_page
     and_when_I_click_button "Publish"
     then_I_should_see_a_publish_form_with_changenotes
@@ -646,6 +646,8 @@ RSpec.feature "Managing step by step pages" do
   def and_I_should_see_an_inset_prompt
     expect(page).to have_css('.govuk-inset-text', text: "To publish this step by step you need to")
   end
+
+  alias_method :then_I_should_see_an_inset_prompt, :and_I_should_see_an_inset_prompt
 
   def and_the_prompt_should_contain_link_to_steps_section(prompt)
     within('.govuk-inset-text') do
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -240,8 +240,7 @@ RSpec.feature "Managing step by step pages" do
     then_I_should_see "Step by steps cannot be published until all steps have content."
     and_I_should_see_an_inset_prompt
     and_the_prompt_should_contain_link_to_steps_section "Add content to all your steps"
-    and_there_should_be_no_publish_button
-    and_there_should_be_no_schedule_button
+    and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
   scenario "A step has not been tested for broken links" do
@@ -273,8 +272,7 @@ RSpec.feature "Managing step by step pages" do
     when_I_view_the_step_by_step_page
     then_I_should_see_an_inset_prompt
     and_the_prompt_should_contain_link_to_steps_section "Add at least one step"
-    and_there_should_be_no_publish_button
-    and_there_should_be_no_schedule_button
+    and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
   scenario "Step by step has not been checked for broken links" do
@@ -282,8 +280,7 @@ RSpec.feature "Managing step by step pages" do
     when_I_view_the_step_by_step_page
     then_I_should_see_an_inset_prompt
     and_the_prompt_should_contain_link_to_steps_section "Check for broken links"
-    and_there_should_be_no_publish_button
-    and_there_should_be_no_schedule_button
+    and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
   scenario "Step by step has been updated since it was last checked for broken links" do
@@ -291,8 +288,7 @@ RSpec.feature "Managing step by step pages" do
     when_I_view_the_step_by_step_page
     then_I_should_see_an_inset_prompt
     and_the_prompt_should_contain_link_to_steps_section "Check for broken links"
-    and_there_should_be_no_publish_button
-    and_there_should_be_no_schedule_button
+    and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
   def and_it_has_change_notes
@@ -650,8 +646,6 @@ RSpec.feature "Managing step by step pages" do
     expect(page).not_to have_link("Schedule")
   end
 
-  alias_method :then_there_should_be_no_schedule_button, :and_there_should_be_no_schedule_button
-
   def and_I_should_see_an_inset_prompt
     expect(page).to have_css('.govuk-inset-text', text: "To publish this step by step you need to")
   end
@@ -806,7 +800,10 @@ RSpec.feature "Managing step by step pages" do
     end
   end
 
-  alias_method :and_there_should_be_no_publish_button, :then_there_should_be_no_publish_button
+  def and_I_cannot_publish_or_schedule_the_step_by_step
+    then_there_should_be_no_publish_button
+    and_there_should_be_no_schedule_button
+  end
 
   def then_there_should_be_no_discard_changes_button
     within(".app-side__actions") do
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -239,7 +239,7 @@ RSpec.feature "Managing step by step pages" do
     when_I_view_the_step_by_step_page
     then_I_should_see "Step by steps cannot be published until all steps have content."
     and_I_should_see_an_inset_prompt
-    and_the_prompt_should_contain "Add content to all your steps"
+    and_the_prompt_should_contain_link_to_steps_section "Add content to all your steps"
     and_there_should_be_no_publish_button
     and_there_should_be_no_schedule_button
   end
@@ -629,9 +629,9 @@ RSpec.feature "Managing step by step pages" do
     expect(page).to have_css('.govuk-inset-text', text: "To publish this step by step you need to")
   end
 
-  def and_the_prompt_should_contain(prompt)
+  def and_the_prompt_should_contain_link_to_steps_section(prompt)
     within('.govuk-inset-text') do
-      expect(page).to have_content prompt
+      expect(page).to have_link(prompt, href: '#steps')
     end
   end
 
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -39,8 +39,8 @@ RSpec.feature "Managing step by step pages" do
     then_I_see_the_new_step_by_step_page
   end
 
-  scenario "User visits an existing step by step page" do
-    given_there_is_a_step_by_step_page
+  scenario "User visits a step by step page with no steps" do
+    given_there_is_a_draft_step_by_step_page_with_no_steps
     when_I_view_the_step_by_step_page
     then_I_can_see_a_summary_section
     and_I_can_edit_the_summary_section
@@ -51,7 +51,7 @@ RSpec.feature "Managing step by step pages" do
     and_I_can_see_a_metadata_section
   end
 
-  scenario "User visits an existing step by step page with steps" do
+  scenario "User visits a step by step page with steps" do
     given_there_is_a_step_by_step_page_with_steps
     when_I_view_the_step_by_step_page
     then_I_can_see_a_summary_section
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -277,6 +277,15 @@ RSpec.feature "Managing step by step pages" do
     and_there_should_be_no_schedule_button
   end
 
+  scenario "Step by step has not been checked for broken links" do
+    given_there_is_a_draft_step_by_step_page
+    when_I_view_the_step_by_step_page
+    then_I_should_see_an_inset_prompt
+    and_the_prompt_should_contain_link_to_steps_section "Check for broken links"
+    and_there_should_be_no_publish_button
+    and_there_should_be_no_schedule_button
+  end
+
   def and_it_has_change_notes
     create(:internal_change_note, step_by_step_page_id: @step_by_step_page.id)
   end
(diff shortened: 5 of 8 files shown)


spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -598,6 +615,11 @@ def and_the_prompt_should_link_to_the_check_links_page
     end
   end
 
+  def but_when_I_edit_a_step
+    # Unable to do this through UI interaction without getting `ActiveRecord::RecordInvalid`
```
ChrisBAshton: Tried this:

```ruby
    click_on "Add a new step"
    fill_in "Step title", with: "Step title"
    fill_in "Content, tasks and links in this step", with: "Some content"
    click_on "Save step"
    expect(page).to have_content("Step was successfully created.")
```

But I get:

```
     ActiveRecord::RecordInvalid:
       Validation failed: Title can't be blank, Base path can't be blank, Content can't be blank, Publishing app can't be blank, Schema name can't be blank
```

Then tried copying the stubbing (and its associated methods) from step_by_step_pages_controller_spec.rb:

```ruby
    allow(Services.publishing_api).to receive(:get_content)
      .with(step_by_step_page.content_id)
      .and_return(content_item(step_by_step_page))
```

But still get the same error message.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -40,7 +40,7 @@
     then_I_see_the_new_step_by_step_page
   end
 
-  scenario "User visits an existing step by step page" do
+  scenario "User visits an existing step by step page with no steps" do
     given_there_is_a_step_by_step_page
```
ChrisBAshton: Should change this to `given_there_is_a_step_by_step_page_with_no_steps`, now that I've made an alias method for it.

app/models/step_by_step_page.rb
```diff
@@ -128,6 +128,10 @@ def links_last_checked_date
     steps.map(&:links_last_checked_date).reject(&:blank?).max
   end
 
+  def links_checked_since_last_update?
+    links_checked? && links_last_checked_date > draft_updated_at
```
ChrisBAshton: Have just done some more testing locally and it looks as though scheduling updates the `draft_updated_at` value.

So you would:

- Have an inset prompt stating you should check links
- You check links, getting rid of the prompt
- You successfully schedule the step by step
- You now have another inset prompt stating you should check links

Is there another field I can use instead I wonder...? 🤔

spec/factories.rb
```diff
@@ -20,19 +21,18 @@
...
-    draft_updated_at { 3.hours.ago }
     published_at { 3.hours.ago }
     status { "published" }
   end
 
-  factory :draft_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 1.day.ago }
-    status { "draft" }
-  end
+  factory :draft_step_by_step_page, parent: :step_by_step_page_with_steps
```
brucebolt: Is there a need for this factory?  It seems to be identical to `step_by_step_page_with_steps`.

ChrisBAshton: It is identical - think of one as an alias for the other. Each are referenced in several test files, in whichever way reads most naturally, with the emphasis either on the fact it's a draft or the fact it has steps.

It feels like it would be nice to go in at some point and reduce to just one (`draft_step_by_step_page_with_steps`?), but I'd have to make a few changes in around 7 files and some of the original intent might be lost. I'd be tempted to leave as-is, giving the two implementations freedom to grow apart if needs arise.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,6 +18,13 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <% unless @step_by_step_page.steps_have_content? %>
```
leenagupte: Where has this moved from? Can you update the commit note to say where it was before?

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,7 +21,11 @@
     <% unless @step_by_step_page.steps_have_content? %>
       <%= render "govuk_publishing_components/components/inset_text" do %>
         <h2 class="govuk-heading-m">To publish this step by step you need to</h2>
-        <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
+        <% if !@step_by_step_page.steps.any? %>
```
leenagupte: A truthy `.empty?` could be used here rather than `!....any?`

spec/factories.rb
brucebolt: FactoryBot allows you to [alias factories](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#aliases).  This may be a more optimal way of having both names for the same factory.

app/views/step_by_step_pages/show.html.erb
leenagupte: Also, I think it would be better to use `can_be_published?` as the guard clause, because that's what's being checked.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -39,8 +39,8 @@
     then_I_see_the_new_step_by_step_page
   end
 
-  scenario "User visits an existing step by step page" do
-    given_there_is_a_step_by_step_page
+  scenario "User visits an existing step by step page with no steps" do
```
leenagupte: Perhaps remove "existing" as well, as that's implied in the steps.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,13 +18,15 @@
...
+    <% unless @step_by_step_page.steps_have_content? && @step_by_step_page.links_checked? %>
       <%= render "govuk_publishing_components/components/inset_text" do %>
         <h2 class="govuk-heading-m">To publish this step by step you need to</h2>
         <% if !@step_by_step_page.steps.any? %>
           <%= link_to('Add at least one step', new_step_by_step_page_step_path(@step_by_step_page), class: 'govuk-link govuk-link--no-visited-state') %>
-        <% else %>
+        <% elsif !@step_by_step_page.steps_have_content? %>
           <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
+        <% elsif !@step_by_step_page.links_checked? %>
+          <%= link_to "Check for broken links", { action: "check_links", step_by_step_page_id: @step_by_step_page.id}, method: :post, class: 'govuk-link govuk-link--no-visited-state' %>
```
leenagupte: Why doesn't this do the same as "Add at least one step" and show link users to the steps section where the "Check for broken links" button is?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -280,6 +280,15 @@
     and_there_should_be_no_schedule_button
   end
 
+  scenario "Step by step has not been checked for broken links" do
+    given_there_is_a_draft_step_by_step_page
+    when_I_view_the_step_by_step_page
+    then_I_should_see_an_inset_prompt
+    and_the_prompt_should_link_to_the_check_links_page
```
leenagupte: There isn't a "check links" page. What's happening at the moment is that a broken link report is being submitted. Perhaps the step split into two: 
```
and_I_see_a_prompt_to_check_for_broken_links
and_I_when_click_on_the_prompt_a_broken_link_report_is_submitted`
```
Though as stated above, I think this second step should just drop the user down to the steps section.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -15,6 +15,7 @@
     given_I_am_a_GDS_editor
     setup_publishing_api
     stub_default_publishing_api_put_intent
+    stub_link_checking_for_default_fixtures
```
leenagupte: Is this needed? The `link_report` factory can be used to create a link_report against a step.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -88,6 +89,7 @@
 
   scenario "User publishes a page" do
     given_there_is_a_draft_step_by_step_page
+    and_I_have_checked_for_broken_links
```
leenagupte: This and the previous step can be combined to use `given_there_is_a_step_by_step_page_with_a_link_report`

spec/factories.rb
```diff
@@ -20,8 +21,11 @@
     end
   end
 
+  factory :step_by_step_page_with_steps_but_no_draft, parent: :step_by_step_page_with_steps do
+    draft_updated_at { nil }
+  end
+
   factory :published_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 3.hours.ago }
```
leenagupte: This should be left in. When `mark_as_published` is called on a step-by-step, the `draft_updated_at` is set to the same date and time as `published_at`.

spec/factories.rb
```diff
@@ -20,8 +21,11 @@
     end
   end
 
+  factory :step_by_step_page_with_steps_but_no_draft, parent: :step_by_step_page_with_steps do
+    draft_updated_at { nil }
```
leenagupte: Is this possible? Unless the step-by-step has been unpublished, there should always be a `draft_updated_at` date.

leenagupte: What do you mean by "no draft"? Does that mean the step by step is published?

spec/models/step_by_step_page_spec.rb
```diff
@@ -306,7 +306,7 @@
   end
 
   describe '#can_be_published?' do
-    let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
+    let(:step_by_step_page) { create(:step_by_step_page_with_steps_but_no_draft) }
```
leenagupte: Why has this changed to "no draft"?

leenagupte: Should this be using `published_step_by_step_page` instead so that the tests still make sense?

app/models/step_by_step_page.rb
```diff
@@ -86,7 +86,7 @@ def unpublished_changes?
   end
 
   def can_be_published?
-    has_draft? && !scheduled_for_publishing? && steps_have_content? && links_checked?
+    has_draft? && !scheduled_for_publishing? && steps_have_content? && links_checked_since_last_update?
```
leenagupte: Nitpick: Can each condition be moved to a separate line to make it easier to read?

e.g.
```
condition_one? &&
  condition_two? &&
  condition_three?
```

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -292,6 +292,19 @@
     and_there_should_be_no_schedule_button
   end
 
+  scenario "Step by step has been updated since it was last checked for broken links" do
```
leenagupte: I think this scenario can be greatly reduced. The first 5 steps could be combined into one new one:
`given_there_is_step_by_step_updated_after_links_last_checked`

leenagupte: `then_there_should_be_no_inset_prompt` could be used in another test where the step-by-step page is eligible to be published.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -680,6 +697,17 @@ def and_the_prompt_should_link_to_the_check_links_page
     end
   end
 
+  def but_when_I_edit_a_step
```
leenagupte: As mentioned abveI think this can be removed. 
There are already steps in `managing_steps_spec.rb` that test editing a step.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -879,4 +907,8 @@ def then_I_should_see_a_filtered_list_of_step_by_steps
       expect(page).to_not have_content("A published step nav")
     end
   end
+
+  def stub_downstream_navigation_rules_updating
+    allow(StepLinksForRules).to receive(:update)
```
leenagupte: Why is this being stubbed?

spec/models/step_by_step_page_spec.rb
```diff
@@ -192,6 +192,30 @@
     end
   end
 
+  describe 'links_checked_since_last_update?' do
+    let(:step_by_step_with_step) { create(:step_by_step_page) }
```
leenagupte: Could this create a `step_by_step_page_with_steps` so a step doesn't need to be created in every step?

spec/models/step_by_step_page_spec.rb
```diff
@@ -310,6 +334,7 @@
 
     before do
       allow(step_by_step_page).to receive(:links_checked?) { true }
+      allow(step_by_step_page).to receive(:links_checked_since_last_update?) { true }
```
leenagupte: Does `links_checked?` still need to be stubbed?

spec/models/step_by_step_page_spec.rb
```diff
@@ -347,6 +372,15 @@
     it 'cannot be published if links have not been checked' do
       step_by_step_page.mark_draft_updated
       allow(step_by_step_page).to receive(:links_checked?) { false }
+      allow(step_by_step_page).to receive(:links_checked_since_last_update?) { false }
```
leenagupte: Also here, does `links_checked?` need to be stubbed?

spec/factories.rb
```diff
@@ -30,13 +30,9 @@
...
   end
 
-  factory :draft_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 1.day.ago }
-    status { "draft" }
-  end
+  factory :draft_step_by_step_page, parent: :step_by_step_page_with_steps
 
   factory :scheduled_step_by_step_page, parent: :step_by_step_page_with_steps do
-    draft_updated_at { 3.hours.ago }
```
leenagupte: Why has this been removed?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -852,6 +848,11 @@ def then_there_should_be_no_publish_button
 
   alias_method :and_there_should_be_no_publish_button, :then_there_should_be_no_publish_button
```
leenagupte: Is this alias still required?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -852,6 +848,11 @@ def then_there_should_be_no_publish_button
...
 
+  def and_I_cannot_publish_or_schedule_the_step_by_step
+    then_there_should_be_no_publish_button
+    and_there_should_be_no_schedule_button
```
leenagupte: There's an alias for this method, `then_there_should_be_no_schedule_button` that should also be removed.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,12 +21,17 @@
...
         <h2 class="govuk-heading-m">To publish this step by step you need to</h2>
+
         <% if !@step_by_step_page.steps.any? %>
           <%= link_to('Add at least one step', new_step_by_step_page_step_path(@step_by_step_page), class: 'govuk-link govuk-link--no-visited-state') %>
-        <% elsif !@step_by_step_page.steps_have_content? %>
-          <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
-        <% elsif !@step_by_step_page.links_checked_since_last_update? %>
-          <%= link_to "Check for broken links", { action: "check_links", step_by_step_page_id: @step_by_step_page.id}, method: :post, class: 'govuk-link govuk-link--no-visited-state' %>
+        <% else %>
+          <% unless @step_by_step_page.steps_have_content? %>
```
leenagupte: Don't use `else` and `unless` together as it's very hard to read. I think it would be better to have an elsif:

```
if @step_by_step_page.steps.empty?
elsif !@step_by_step_page.steps_have_content?
```

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,12 +21,17 @@
...
+            <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
+          <% end %>
+
+          <% unless @step_by_step_page.links_checked_since_last_update? %>
```
leenagupte: Same here. Using a nested `unless` is hard to read. `links_checked_since_last_update?` would be better being renamed something like `check_links?` as I suggested yesterday, and then it could be called with an `if` and moved out of the nested conditionals.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -301,6 +301,16 @@
     and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
+  scenario "Multiple check failures are each listed in the inset prompt" do
```
leenagupte: I wasn't sure what "check failures" meant. Perhaps it could be renamed to "required actions"

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,20 +18,27 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% unless @step_by_step_page.can_be_published? %>
+    <% if !@step_by_step_page.can_be_published? && !@step_by_step_page.scheduled_for_publishing? %>
```
leenagupte: Does the whole `inset` need to be wrapped in this? Surely, it only needs to be incorporated into the link check, as the step by step can't be published or scheduled if it doesn't have any steps etc.

leenagupte: `can_be_published?` already does the `!@step_by_step_page.scheduled_for_publishing?` so it doesn't need to be repeated here.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -311,6 +311,18 @@
...
   end
 
+  scenario "Fixing issues identified in the inset prompt, clears the inset prompt" do
+    given_there_is_a_step_by_step_page_with_steps
+    and_it_has_been_edited_since_its_links_were_last_checked
+    then_I_should_see_an_inset_prompt
+    and_the_prompt_should_link_to_the_check_links_page
+    but_when_I_check_for_broken_links
+    then_there_should_be_no_inset_prompt
+    and_when_I_schedule_the_step_by_step
```
leenagupte: Scheduling isn't part of the "fixing issues" flow. Perhaps checking if the publish button appears would be better?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -311,6 +311,18 @@
...
+    then_the_step_by_step_is_scheduled
+    and_there_should_continue_to_be_no_inset_prompt
```
leenagupte: Perhaps this should be added to one of the scheduling tests that check that a step-by-step can't be edited.

spec/models/step_by_step_page_spec.rb
```diff
@@ -216,6 +216,27 @@
...
   end
 
+  describe 'should_check_links?' do
+    it 'should return false if step by step is scheduled' do
+      step_by_step_with_step = create(:step_by_step_page, status: 'scheduled', scheduled_at: 1.day.from_now)
+
+      expect(step_by_step_with_step.should_check_links?).to be false
+    end
+
+    it 'should return false if step by step has not been edited since publish' do
```
leenagupte: I think this should be renamed to `should return false if the status of the step by step is published`

spec/models/step_by_step_page_spec.rb
```diff
@@ -216,6 +216,27 @@
     end
   end
 
+  describe 'should_check_links?' do
+    it 'should return false if step by step is scheduled' do
+      step_by_step_with_step = create(:step_by_step_page, status: 'scheduled', scheduled_at: 1.day.from_now)
```
leenagupte: The `step_by_step_page` could be created once in a `let` block, and the status updated in each test as needed. e.g. here the arrangement could be `step_by_step_page.status == "scheduled"`.

leenagupte: Or use the `scheduled_step_by_step_page` factory.

spec/models/step_by_step_page_spec.rb
```diff
@@ -216,6 +216,27 @@
...
+
+      expect(step_by_step_with_step.should_check_links?).to be false
+    end
+
+    it 'should return false if step by step has not been edited since publish' do
+      publish_time = Time.zone.now
+      step_by_step_with_step = create(:step_by_step_page, status: 'published', draft_updated_at: publish_time, published_at: publish_time)
```
leenagupte: The `published_step_by_step_page` could be used here.

spec/models/step_by_step_page_spec.rb
```diff
@@ -216,6 +216,27 @@
...
+
+      expect(step_by_step_with_step.should_check_links?).to be false
+    end
+
+    it 'should return true if step by step is in draft' do
+      step_by_step_with_step = create(:step_by_step_page, status: 'draft')
```
leenagupte: The status doesn't need to be set to "draft" as that's the default for this factory.

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -102,3 +102,7 @@
   }
 }
 // scss-lint:enable IdSelector
+
+.gem-c-inset-text {
```
leenagupte: Where is this being used?

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,20 +18,27 @@
...
+            <li>
+              <%= link_to('Add at least one step', new_step_by_step_page_step_path(@step_by_step_page), class: 'govuk-link govuk-link--no-visited-state') %>
+            </li>
+          <% else %>
+            <% unless @step_by_step_page.steps_have_content? %>
+              <li>
+                <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
+              </li>
+            <% end %>
+            <% if @step_by_step_page.should_check_links? && !@step_by_step_page.links_checked_since_last_update? %>
```
leenagupte: Can `!@step_by_step_page.links_checked_since_last_update?` be part of the `should_check_links?` check?

app/models/step_by_step_page.rb
```diff
@@ -128,6 +128,10 @@ def links_last_checked_date
     steps.map(&:links_last_checked_date).reject(&:blank?).max
   end
 
+  def should_check_links?
```
leenagupte: Should this also return false if there are no links to check? E.g if there aren't any `navigation_rules`?

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: I'll clarify with Tobi - I would have thought this should actually submit the 'Check for broken links' form.

spec/factories.rb
ChrisBAshton: I did this because of [this test](https://github.com/alphagov/collections-publisher/blob/master/spec/models/step_by_step_page_spec.rb#L345-L347):

```ruby
    it 'cannot be published if it does not have a draft' do
      expect(step_by_step_page.can_be_published?).to be false
    end
```

Now that the default factory sets the `draft_updated_at` field, this was returning `true`. But what I could do is call `mark_as_published` in this step before calling `can_be_published?`?

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: The structure is roughly:

```
if ...empty?
  link 1
else
        if !steps_have_content?
          link 2
        elsif some_other_condition?
          link 3
end
```

To allow for link 1 and link 2/3 to be displayed at the same time, this needs to be nested as above. But can do `if !` 👍

app/assets/stylesheets/utilities/_overrides.scss
ChrisBAshton: It's on the `inset-text` component we pull into `show.html`.

app/views/step_by_step_pages/show.html.erb
ChrisBAshton: I'll clarify with Tobi - I would have thought this should actually submit the 'Check for broken links' form.

ChrisBAshton: Have double-checked, she agrees it should link to the section 👍 this may change after user research. I'll go ahead and do that now - simplifies it a lot!

spec/factories.rb
leenagupte: This test expects `has_draft?` to return false.

app/views/step_by_step_pages/show.html.erb
leenagupte: I think it might be nicer to use a list of if statements. Once for each action.
e.g.
`if empty?`
`if no content?`
`if links not checked` 
each if could contain all of the logic it needs, e.g. should_check_links? could check whether there are any steps, and if the steps have content. This keeps the nested logic out of the view.

app/assets/stylesheets/utilities/_overrides.scss
leenagupte: Can this changed be added where it's used to make that clear?

spec/factories.rb
ChrisBAshton: `mark_as_published` would mean there is no draft, therefore `can_be_published?` becomes `false`.

Can you explain what your proposal is?

spec/factories.rb
ChrisBAshton: Nice!

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: Good spot.

spec/factories.rb
ChrisBAshton: Is it that we should be using `published_step_by_step_page` instead? (Agreed that makes more sense as a step by step with steps but no draft is not possible).

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,7 +21,11 @@
     <% unless @step_by_step_page.steps_have_content? %>
       <%= render "govuk_publishing_components/components/inset_text" do %>
         <h2 class="govuk-heading-m">To publish this step by step you need to</h2>
-        <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
+        <% if @step_by_step_page.steps.empty? %>
+          <%= link_to('Add at least one step', new_step_by_step_page_step_path(@step_by_step_page), class: 'govuk-link govuk-link--no-visited-state') %>
```
ChrisBAshton: Fix

app/views/step_by_step_pages/show.html.erb
```diff
@@ -21,11 +21,16 @@
...
           <%= link_to('Add at least one step', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
-        <% elsif !@step_by_step_page.steps_have_content? %>
+        <% end %>
+
+        <% if @step_by_step_page.steps.any? && !@step_by_step_page.steps_have_content? %>
           <%= link_to('Add content to all your steps', '#steps', class: 'govuk-link govuk-link--no-visited-state') %>
-        <% elsif !@step_by_step_page.links_checked_since_last_update? %>
+        <% end %>
+
+        <% if @step_by_step_page.steps.any? && !@step_by_step_page.links_checked_since_last_update? %>
```
ChrisBAshton: Move `<ul><li>` stuff into this commit.

spec/models/step_by_step_page_spec.rb
```diff
@@ -214,6 +214,27 @@
...
 
+  describe 'should_check_links?' do
+    it 'should return false if step by step is scheduled' do
+      step_by_step_with_step = create(:step_by_step_page, status: 'scheduled', scheduled_at: 1.day.from_now)
+
+      expect(step_by_step_with_step.should_check_links?).to be false
+    end
+
+    it 'should return false if step by step has not been edited since publish' do
+      publish_time = Time.zone.now
```
ChrisBAshton: Remove this

spec/features/managing_step_by_step_pages_spec.rb
ChrisBAshton: Note to self to remove `and_the_prompt_should_link_to_the_check_links_page` from this commit.

spec/factories.rb
leenagupte: This test is checking the first part of this condition: https://github.com/alphagov/collections-publisher/blob/cd98e0e72ff212144d04242b487f58d61d906c37/app/models/step_by_step_page.rb#L94 that `has_draft?` is false. 

So, yes using `published_step_by_step_page` should work, as long as the conditions that `has_draft?` relies on haven't changed.

spec/support/step_nav_steps.rb
```diff
@@ -103,6 +103,23 @@ def given_there_is_a_step_by_step_page_with_a_link_report
...
+    step = create(:step)
+    create(:link_report, step_id: step.id)
+    @step_by_step_page = create(
+      :published_step_by_step_page,
+      steps: [step],
+      slug: "step-by-step-with-link-report-and-unpublished-changes",
+      draft_updated_at: 1.day.ago,
+      status: "draft"
+    )
+  end
```
ChrisBAshton: Look at refactoring this.

spec/support/step_nav_steps.rb
```diff
@@ -120,6 +120,23 @@ def given_there_is_a_step_by_step_page_with_unpublished_changes_whose_links_have
...
+    step = create(:step)
+    create(:link_report, step_id: step.id)
+    @step_by_step_page = create(
+      :published_step_by_step_page,
+      steps: [step],
+      slug: "step-by-step-with-link-report-and-unpublished-changes-since-links-checked",
+      draft_updated_at: Time.zone.now,
+      status: "draft"
+    )
+  end
```
ChrisBAshton: Find way of reducing duplication between this and the factory helper above it.

spec/support/step_nav_steps.rb
```diff
@@ -137,6 +137,24 @@ def given_a_step_by_step_has_been_updated_after_links_last_checked
...
+      :published_step_by_step_page,
+      slug: "step-by-step-with-link-report-and-empty-step-added-since-links-checked",
+      draft_updated_at: Time.zone.now,
+      status: "draft"
+    )
+
+    step_with_link_report = create(:step, step_by_step_page: @step_by_step_page)
+    create(:link_report, step_id: step_with_link_report.id)
+    create(:step, contents: "", step_by_step_page: @step_by_step_page)
+  end
```
ChrisBAshton: And this one.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -312,6 +323,22 @@ def and_I_visit_the_publish_page
...
+        "https://www.gov.uk/also/good/stuff",
+        "https://www.gov.uk/not/as/great"
+      ],
+      webhook_uri: "https://collections-publisher.test.gov.uk/link_report"
+    )
+    stub_link_checker_api_get_batch(id: 0)
+
+    when_I_view_the_step_by_step_page
+    click_button "Check for broken links"
+  end
```
ChrisBAshton: Find way of doing this without having to stub.

spec/models/step_by_step_page_spec.rb
```diff
@@ -306,7 +306,7 @@
   end
 
   describe '#can_be_published?' do
-    let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
+    let(:step_by_step_page) { create(:published_step_by_step_page) }
```
brucebolt: Is this the correct factory to be using here?  I think a step by step in draft state might be more suited?

spec/models/step_by_step_page_spec.rb
```diff
@@ -309,7 +331,7 @@
     let(:step_by_step_page) { create(:published_step_by_step_page) }
 
     before do
-      allow(step_by_step_page).to receive(:links_checked?) { true }
+      allow(step_by_step_page).to receive(:links_checked_since_last_update?) { true }
     end
 
     it 'can be published if it has a draft, is not scheduled, all steps have content and links have been checked' do
```
brucebolt: I'd suggest updating this (and all other similar test cases) to say something like `...links have been checked since last update` (in commit c48bea5f91a0c1de86be463a79cfa44d09bfea3f) since you're now testing `links_checked_since_last_update?` rather than `links_checked?`.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -806,7 +800,10 @@ def then_there_should_be_no_publish_button
     end
   end
 
-  alias_method :and_there_should_be_no_publish_button, :then_there_should_be_no_publish_button
+  def and_I_cannot_publish_or_schedule_the_step_by_step
```
brucebolt: Combining this is a nice idea 👍

spec/support/step_nav_steps.rb
```diff
@@ -122,6 +122,19 @@ def given_a_step_by_step_has_been_updated_after_links_last_checked
     )
   end
 
+  def given_a_step_by_step_has_had_an_empty_step_added_after_links_last_checked
```
brucebolt: Is there a way this could be shortened (it took me a while to process it)?

Although I'm struggling to think of anything much shorter 🤔: 
`given_a_step_by_step_has_an_empty_step_added_without_links_checked`
`given_a_step_by_step_has_an_empty_step_added_after_links_last_checked`

app/models/step_by_step_page.rb
```diff
@@ -131,6 +131,10 @@ def links_last_checked_date
     steps.map(&:links_last_checked_date).reject(&:blank?).max
   end
 
+  def should_show_required_prepublish_actions?
+    has_draft? && !scheduled_for_publishing? && !can_be_published?
```
brucebolt: Is the latter part of this conditional (`&& !can_be_published?`) required?  `has_draft?` checks if the `draft_updated_at` is not equal to `published_at`, so I think this should be enough to determine whether the `draft_updated_at` date changed because it was published.

The `!scheduled_for_publishing?` is definitely required.

spec/models/step_by_step_page_spec.rb
brucebolt: I agree with Leena's comment here, since the test is relying on a published status doing what it should do.

app/models/step_by_step_page.rb
ChrisBAshton: I think this is still required because `can_be_published?` also checks for `steps_have_content?`. Something might have a draft but also have empty steps.

brucebolt: Cool, that makes sense.  I didn't consider that scenario.

spec/models/step_by_step_page_spec.rb
```diff
@@ -306,9 +355,13 @@
   end
 
   describe '#can_be_published?' do
-    let(:step_by_step_page) { create(:step_by_step_page_with_steps) }
+    let(:step_by_step_page) { create(:published_step_by_step_page) }
```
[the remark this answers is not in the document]
ChrisBAshton: Leena and I had a chat about this in https://github.com/alphagov/collections-publisher/pull/730#discussion_r324759634.

I originally used a draft state but that meant having to do something funny with one of the tests. We opted for choosing a published state instead.

ChrisBAshton:
Re-ran the tests and they passed. ✅

PR #736, opened 2019-09-13
Show submission confirmation to user when they check for broken links

ChrisBAshton:
When a user checks for broken links, they should see a message telling
them that their check is in progress. Previously we provided no feedback.
If the user refreshes the page, the message goes away.

Trello: https://trello.com/c/qensq3Jw/57-provide-feedback-when-clicking-check-for-broken-links

Screen Shot 2019-09-13 at 09 44 53

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -145,6 +145,7 @@ class StepByStepPagesController < ApplicationController
   def check_links
     set_current_page_as_step_by_step
     @step_by_step_page.steps.each(&:request_broken_links)
+    redirect_to @step_by_step_page, notice: "Links are currently being checked. Please refresh the page to check progress. When all links have been checked, you'll see a message below each step."
   end
 
   def internal_change_notes
--- a/spec/features/managing_step_by_step_pages_spec.rb
+++ b/spec/features/managing_step_by_step_pages_spec.rb
@@ -271,6 +271,24 @@ RSpec.feature "Managing step by step pages" do
     and_I_cannot_publish_or_schedule_the_step_by_step
   end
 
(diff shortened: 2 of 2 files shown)


app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,6 +18,19 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <% unless @step_by_step_page.can_be_published? %>
+      <%= render "govuk_publishing_components/components/inset_text" do %>
+        <h2 class="govuk-heading-m">To publish this step by step you need to</h2>
+        <% if !@step_by_step_page.steps.any? %>
+          <a href="<%= new_step_by_step_page_step_path(@step_by_step_page) %>">Add at least one step</a>
```
alex-ju: I suggest we use the Rails helpers (`link_to`) consistently to generate links and add `govuk-link` class to them.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,6 +18,19 @@
...
+        <% elsif !@step_by_step_page.steps_have_content? %>
+          <a href="#steps">Add content to all your steps</a>
```
alex-ju: This needs `govuk-link` class to be added to it.

app/views/step_by_step_pages/show.html.erb
```diff
@@ -18,6 +18,19 @@
...
+        <% elsif !@step_by_step_page.links_checked_since_last_update? %>
+          <%= link_to "Check for broken links", { action: "check_links", step_by_step_page_id: @step_by_step_page.id}, method: :post %>
```
alex-ju: This needs `govuk-link` class too.

PR #741, opened 2019-09-16
Update tabs

alex-ju:
This updates the tabs section to only show the Summary and History (as all the rest of them became now redundant, being accessed using the links in the summary-list component or the actions block on the right-hand side).

Based on #711, but a lot less ambitious.

### Before
Screen Shot 2019-09-16 at 14 28 34

### After
Screen Shot 2019-09-16 at 14 27 49

[Trello card](https://trello.com/c/ss4RmXoT)

### DIFF
--- a/app/assets/stylesheets/_steps.scss
+++ b/app/assets/stylesheets/_steps.scss
@@ -122,11 +122,6 @@ $dark-grey-2: #6f777b;
   margin: 20px 0 20px;
 }
 
-.last-saved-by--text {
-  margin: 20px 0 20px;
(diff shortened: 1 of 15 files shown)


alex-ju:
@leenagupte: I updated the tests to extract the `when_I_visit_the_history_page` into its own step and updates the tests suite to reflect the fact that navigation to the History page is required to check the change notes. Hopefully this makes it clearer in tests when navigation to the history page happens and we already have the `when_I_view_the_step_by_step_page` and `then_I_can_see_a_summary_section` which covers the Summary page.

PR #746, opened 2019-09-18
Add "notify users about change" to step-by-step scheduling workflow

brucebolt:
Updates scheduling workflow to let content designers choose if they want to tell users about the change to a step-by-step.

This is an option when publishing a step-by-step immediately, but was not available when scheduling.

First view in the workflow:
Screen Shot 2019-09-12 at 15 08 54

Second view in the workflow:
Screen Shot 2019-09-12 at 15 09 01

Trello card: https://trello.com/c/XM1zjRfM/62-add-notify-users-about-change-to-scheduling-workflow

This is a much simplified version of https://github.com/alphagov/collections-publisher/pull/735.

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -96,6 +96,7 @@ class StepByStepPagesController < ApplicationController
       elsif @step_by_step_page.update_attributes(scheduled_at: scheduled_at)
         schedule_to_publish(session[:update_type], session[:public_change_note])
         note_description = "Scheduled by #{current_user.name} for publishing at #{format_full_date_and_time(scheduled_at)}"
+        note_description << " with change note: #{session[:public_change_note]}" if session[:public_change_note].present?
         generate_internal_change_note(note_description)
         set_change_note_version
         redirect_to @step_by_step_page, notice: "'#{@step_by_step_page.title}' has been scheduled to publish."
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -78,7 +78,11 @@ class StepByStepPagesController < ApplicationController
 
   def schedule
     set_current_page_as_step_by_step
-    if request.post?
+  end
+
+  def schedule_datetime
+    set_current_page_as_step_by_step
+    if params[:schedule]
       date_params = params[:schedule][:date].permit(:year, :month, :day).to_h.symbolize_keys
       time_param = params[:schedule][:time]
       @schedule_placeholder = default_datetime_placeholder(date_params.merge(time: time_param))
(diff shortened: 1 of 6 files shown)


app/controllers/step_by_step_pages_controller.rb
```diff
@@ -78,7 +78,11 @@ def publish
 
   def schedule
     set_current_page_as_step_by_step
-    if request.post?
+  end
+
+  def schedule_datetime
```
ChrisBAshton: This method is quite long and difficult to read. I wonder if it could be split out into well named methods that document what it's doing? Something like:

```ruby
def schedule
  if !changenotes_supplied?
    render :schedule_changenotes
  elsif !datetime_inputs_supplied? || parse_datetime_inputs.issues.any?
    render :schedule_datetime
  else
    schedule!(changenotes, datetime)
  end
end
```

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -153,6 +153,21 @@ def schedule_for_future
...
     end
+
+    it "creates an internal change note for minor change" do
+      schedule_for_future
+
+      expected_description = "Scheduled by Name Surname for publishing at 10:26am on 20 April 2030"
+      expect(step_by_step_page.internal_change_notes.first.description).to eq expected_description
+    end
+
+    it "creates an internal change note for major change" do
```
ChrisBAshton: I'd take inspiration from the scenario names from the `publish` tests here, i.e.

`it "generates an internal change note with change note text for major changes"`

and

`it "generates an internal change note without change note text for minor changes"`

app/controllers/step_by_step_pages_controller.rb
```diff
@@ -78,7 +78,11 @@ def publish
...
+    set_current_page_as_step_by_step
+    if request.post? && params[:schedule]
```
ChrisBAshton: `schedule_datetime` is only accessible over POST, so this can be simplified to `params[:schedule]`.

spec/controllers/step_by_step_pages_controller_spec.rb
brucebolt: I'll update the major change description.  I'll leave the minor change description since the description is technically correct (i.e. a change note has been created).

app/controllers/step_by_step_pages_controller.rb
brucebolt: Indeed, this method does need breaking up (a particularly prime candidate would be the parsing code).  I've chosen to just add the changes needed here to deal with the change note parameters, leaving the rest of the method untouched.  We can do a refactor later on, if there's time at the end of the mission.

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -184,6 +184,28 @@
...
+      and_I_visit_the_scheduling_page
+      then_I_should_see_a_publish_form_with_changenotes
+      and_I_fill_in_the_changenote_form
+      and_when_I_click_button "Continue"
+      then_inputs_should_have_tomorrows_date
+      and_I_fill_in_the_scheduling_form
+      when_I_submit_the_form
+      then_I_should_see "has been scheduled to publish"
+      and_the_step_by_step_should_have_the_status "Scheduled"
+      and_there_should_be_a_change_note "Scheduled by Test author for publishing at 10:26am on 20 April 2030 with change note: We made some changes"
```
ChrisBAshton: I think everything below this line can be removed (and put in a less complex scenario if it hasn't been covered yet) as this scenario is very long.

brucebolt: Good point, I've removed those lines now

PR #747, opened 2019-09-18
Allow 2i reviews to be approved, or changes requested

brucebolt:
Once a 2i review has been requested, the reviewer needs to be able to either approve the review or request changes.

This PR carries the backend work.  Frontend work will be done separately.

Trello cards: https://trello.com/c/R85nupi5/38-2i-reviewer-can-approve-step-by-steps-backend and https://trello.com/c/rlPS1Kyo/36-2i-reviewer-can-request-change-to-step-by-step-backend

### DIFF
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -3,8 +3,22 @@ class ReviewController < ApplicationController
 
   before_action :require_gds_editor_permissions!
   before_action :require_unreleased_feature_permissions!
-  before_action :require_2i_reviewer_permissions!, only: %i(claim_2i_review)
+  before_action :require_2i_reviewer_permissions!, only: %i(approve_2i_review claim_2i_review)
   before_action :set_step_by_step_page
+
+  def approve_2i_review
+    status = "2i_approved"
+
+    if @step_by_step_page.update(
+      review_requester_id: nil,
+      reviewer_id: nil,
+      status: status
+    )
+      generate_change_note(status)
+
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i_approved."
+    end
+  end
 
   def claim_2i_review
     status = "in_review"
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -3,7 +3,7 @@ class ReviewController < ApplicationController
 
   before_action :require_gds_editor_permissions!
   before_action :require_unreleased_feature_permissions!
-  before_action :require_2i_reviewer_permissions!, only: %i(approve_2i_review claim_2i_review)
+  before_action :require_2i_reviewer_permissions!, only: %i(approve_2i_review claim_2i_review request_change_2i_review)
   before_action :set_step_by_step_page
 
   def approve_2i_review
@@ -30,6 +30,20 @@ class ReviewController < ApplicationController
       generate_change_note(status)
 
       redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully claimed for review."
+    end
+  end
+
+  def request_change_2i_review
+    status = "draft"
+
+    if @step_by_step_page.update(
+      reviewer_id: nil,
+      review_requester_id: nil,
+      status: status
+    )
+      generate_change_note("Changes requested", params[:requested_change])
+
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested."
     end
   end
 
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -1,5 +1,6 @@
 class StepByStepPage < ApplicationRecord
   STATUSES = %w(
+    2i_approved
     draft
     in_review
     published
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,6 +7,7 @@ Rails.application.routes.draw do
     post :check_links
     post 'approve-2i-review', to: 'review#approve_2i_review'
     post 'claim-2i-review', to: 'review#claim_2i_review'
+    post 'request-change-2i-review', to: 'review#request_change_2i_review'
     get 'internal-change-notes', to: 'interal_change_notes'
     post 'internal-change-notes', to: 'internal_change_notes#create'
     get 'navigation-rules', to: 'navigation_rules#edit'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,7 @@ Rails.application.routes.draw do
 
   resources :step_by_step_pages, path: 'step-by-step-pages' do
     post :check_links
+    post 'approve-2i-review', to: 'review#approve_2i_review'
     post 'claim-2i-review', to: 'review#claim_2i_review'
     get 'internal-change-notes', to: 'interal_change_notes'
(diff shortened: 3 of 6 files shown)


spec/controllers/review_controller_spec.rb
```diff
@@ -71,6 +71,72 @@
...
+
+      it "cannot be accessed by users with only signin permissions" do
+        stub_user.permissions = %w(signin)
+        post :approve_2i_review, params: { step_by_step_page_id: step_by_step_page.id }
+
+        expect(response.status).to eq(403)
+      end
+
+      it "sets status to approved" do
+        stub_user.permissions = ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"]
```
ChrisBAshton: Could you extract this out to a `let` variable at the top of this `describe` block?

```ruby
let(:required_permissions) { ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"] }
```

It's used in a few places and if/when it changes in future we should only have to update it once.

config/routes.rb
```diff
@@ -6,6 +6,7 @@
   resources :step_by_step_pages, path: 'step-by-step-pages' do
     post :check_links
     post 'claim-2i-review', to: 'review#claim_2i_review'
+    post 'request-change-2i-review', to: 'review#request_change_2i_review'
```
ChrisBAshton: Could this be made alphabetical like you've done in `before_action` in the `review_controller.rb`?

spec/controllers/review_controller_spec.rb
```diff
@@ -137,6 +137,74 @@
...
+      end
+
+      it "can be accessed by users with GDS editor, Unreleased feature and 2i reviewer permissions" do
+        stub_user.permissions = ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"]
+        post :request_change_2i_review, params: { step_by_step_page_id: step_by_step_page.id }
+
+        expect(response).to redirect_to step_by_step_page_path(step_by_step_page)
+      end
+
+      it "cannot be accessed by users with only GDS editor permissions" do
```
ChrisBAshton: And rather than testing where users only have one permission, I think we could have more confidence in our tests if we're testing for the absence of a required permission. For example:

```ruby
it "cannot be accessed by users missing the GDS editor permission" do
  stub_user.permissions = required_permissions - "GDS Editor"
  # ...etc
```

I'd even be tempted to take it a step further with something like this, as all the fail cases are identical:

```ruby
it "cannot be accessed by users missing any of the required permissions" do
  required_permissions.each.do |required_permission|
    stub_user.permissions = required_permissions - required_permission
    # ...etc
  end
end
```

spec/controllers/review_controller_spec.rb
```diff
@@ -137,6 +137,74 @@
...
+
+      it "sets status to draft, removes reviewer_id and review_requester_id" do
+        stub_user.permissions = ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"]
+        post :request_change_2i_review, params: { step_by_step_page_id: step_by_step_page.id }
+
+        step_by_step_page.reload
+
+        expect(step_by_step_page.status).to eq("draft")
+        expect(step_by_step_page.reviewer_id).to be_nil
+        expect(step_by_step_page.review_requester_id).to be_nil
```
ChrisBAshton: This is nice 👍 We don't seem to be resetting `reviewer_id` and `review_requester_id` where the 2i was _approved_ - should we be? It seems like it would be good housekeeping.

app/models/step_by_step_page.rb
```diff
@@ -1,5 +1,6 @@
 class StepByStepPage < ApplicationRecord
   STATUSES = %w(
+    approved
```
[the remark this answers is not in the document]
brucebolt: I agree, however I kept this consistent with `in_review` which doesn't mention 2i.

I think the statuses should be stored in the database as integers anyway and use Rails enum feature to refer to these as strings in the code.  That way, it would be easy to change the status text (e.g. when we implement factcheck) without having to do a database migration.  I'll see how much work would be involved in changing this value.

app/controllers/review_controller.rb
```diff
@@ -3,9 +3,21 @@ class ReviewController < ApplicationController
...
-  before_action :require_2i_reviewer_permissions!, only: %i(claim_2i_review)
+  before_action :require_2i_reviewer_permissions!, only: %i(approve_2i_review claim_2i_review)
   before_action :set_step_by_step_page
 
+  def approve_2i_review
+    status = "approved"
+
+    if @step_by_step_page.update(
+      status: status
+    )
```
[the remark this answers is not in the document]
brucebolt: Yes, but I'll update based on https://github.com/alphagov/collections-publisher/pull/747#discussion_r326043323, so there will now be additional values here.

spec/controllers/review_controller_spec.rb
brucebolt: The combination of users if slightly different in various tests, so I'd rather leave the full list present each time, so we're not switching between different ways of setting the permissions

spec/controllers/review_controller_spec.rb
brucebolt: I really like this idea.  Implementing it as an additional commit for the other routes too.

spec/validators/status_prerequisite_validator_spec.rb
```diff
@@ -37,6 +37,36 @@
...
+      allow(step_by_step_page).to receive(:has_draft?).and_return(false)
+      step_by_step_page.reviewer_id = SecureRandom.uuid
+      step_by_step_page.status = "approved"
+
+      expect(step_by_step_page).not_to be_valid
+      expect(step_by_step_page.errors.messages[:status]).to eq([error_message])
+    end
+
+    it "does not allow status to be approved if the current status is not submitted_for_2i" do
+      allow(step_by_step_page).to receive(:has_draft?).and_return(true)
```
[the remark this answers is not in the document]
brucebolt: I've decided not to change this.  We're using the same step by step for all tests here, so it would result in unnecessary extra code if I were to add an additional create step here.

spec/controllers/review_controller_spec.rb
```diff
@@ -10,31 +10,30 @@
...
-          stub_user.permissions << "GDS Editor"
-          get :submit_for_2i, params: { step_by_step_page_id: step_by_step_page.id }
-
-          expect(response.status).to eq(403)
-        end
-
-        it "cannot be accessed by users with neither GDS editor and Unreleased feature permissions" do
-          stub_user.permissions = %w(signin)
-          get :submit_for_2i, params: { step_by_step_page_id: step_by_step_page.id }
+        required_permissions.drop(1).each do |required_permission|
```
ChrisBAshton: Instead of `drop(1)` which depends on the order of the permissions, might I suggest removing it by value?

```ruby
(required_permissions - ["signon"]).each do |required_permission|
```

spec/controllers/review_controller_spec.rb
```diff
@@ -10,31 +10,30 @@
...
+          it "cannot be accessed by users without the #{required_permission} permissions" do
```
ChrisBAshton: Nitpick - I'd make this singular, as it only deals with one permission at a time:

```suggestion
          it "cannot be accessed by users without the #{required_permission} permission" do
```

Same for the ones below.

PR #760, opened 2019-09-23
Make step markdown link-checking more robust (fix sidebar settings & preview)

ChrisBAshton:
Fixes two bugs:

1) "Sidebar settings" page on step by step pages would break if there was a broken markdown URL in any step. They are now safely ignored (first commit).
2) Broken markdown links would also prevent the preview draft from updating. They are now safely ignored (and are simply rendered as broken markdown in the preview, making it obvious to content designers that there is a problem).

Trello card: https://trello.com/c/2hNZqWUY/67-preview-does-not-update-if-there-is-invalid-markdown-in-a-step

### DIFF
--- a/app/services/step_content_parser.rb
+++ b/app/services/step_content_parser.rb
@@ -39,10 +39,15 @@ class StepContentParser
   end
 
   def base_paths(step_text)
-    relative_paths(step_text).map do |path|
-      uri = URI.parse(path)
-      uri.path
+    paths = relative_paths(step_text).map do |path|
+      begin
+        uri = URI.parse(path)
+        uri.path
+      rescue URI::InvalidURIError
+        nil
+      end
     end
+    paths.compact
   end
 
   def all_paths(step_text)
--- a/app/services/step_content_parser.rb
+++ b/app/services/step_content_parser.rb
@@ -39,15 +39,7 @@ class StepContentParser
   end
 
   def base_paths(step_text)
-    paths = relative_paths(step_text).map do |path|
-      begin
-        uri = URI.parse(path)
-        uri.path
-      rescue URI::InvalidURIError
-        nil
-      end
-    end
-    paths.compact
+    relative_paths(step_text).map { |path| safely_parse_path(path) }.compact
   end
 
   def all_paths(step_text)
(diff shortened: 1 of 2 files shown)


app/services/step_content_parser.rb
```diff
@@ -87,16 +87,23 @@ def bulleted_list?(section)
...
-      if line.scan(/\[/).empty?
-        { "text": line[2..-1] }
-      elsif /\[(?<text>(.+))\]\((?<href>(.+))\)((?<context>.*))$/ =~ line
-        payload = {
-          "text": text,
-          "href": href
-        }
-        payload[:context] = context.strip unless context.blank?
-        payload
+      payload = {
```
brucebolt: There's a couple of nested `if` statements, with the `else` outside of the `if`, which made it difficult to fully understand what the different scenarios are.

Could this be written as (for example):
```
  def link_content(section)
    section.map do |line|
      payload = {}
      if /\[(?<text>(.+))\]\((?<href>(.+))\)((?<context>.*))$/ =~ line && safely_parse_path(href)
        payload = {
          "text": text,
          "href": href,
        }
        payload[:context] = context.strip unless context.blank?
      else
        payload[:text] = line[2..-1]
      end
      payload
    end
  end
```

If the assumption I've made that the two conditionals can be joined with an `and` is incorrect, then could a test be added to cater for cases where the first condition is true but the second is false.  My assumption has resulted in all the tests passing with the proposed code.

app/services/step_content_parser.rb
```diff
@@ -87,16 +87,23 @@ def bulleted_list?(section)
...
+            "href": href,
+          }
+          payload[:context] = context.strip unless context.blank?
+        end
       end
+      payload
     end
   end
+
+  def remove_bullet_prefix(line)
```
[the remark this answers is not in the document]
ChrisBAshton: Good spot. It was at some point being used twice (inside the `if line.scan(/\[/).empty?` conditional, and then inside the next conditional somewhere in its nest). I can remove it now 👍

app/services/step_content_parser.rb
ChrisBAshton: Good idea, will rewrite 👍 wasn't sure about using the `href` variable in the same conditional as it is created in.

PR #761, opened 2019-09-23
Add submit for 2i review action

alex-ju:
This work introduces the "Submit for 2i review" and the results or that action. I'm trying to avoid a massive PR, so since we're behind a feature flag I'm raising early to check the direction of this work with the rest of the team. Planning to raise a follow up PR that enables "claiming" after this gets merged.

### Preview
In draft state or with changes after being published
2i-draft

After submitted for 2i
2i-submitted-for-2i

[Trello card](https://trello.com/c/Fwo562zl)

### DIFF
--- a/app/views/step_by_step_pages/show/_actions.erb
+++ b/app/views/step_by_step_pages/show/_actions.erb
@@ -15,8 +15,6 @@
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
     <% elsif @step_by_step_page.can_be_published? %>
-      <p class="govuk-body">Before publishing check for broken links.</p>
-      <p class="govuk-body">After publishing add this step by step to a mainstream browse category and a taxonomy topic.</p>
       <% if @step_by_step_page.has_been_published? %>
         <%= render "govuk_publishing_components/components/button", {
           text: "Publish",
@@ -32,9 +30,6 @@
         step_by_step_page_schedule_path(@step_by_step_page),
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
-    <% elsif !@step_by_step_page.steps_have_content? %>
-      <p class="govuk-body">Step by steps cannot be published until all steps have content.</p>
-      <%= preview_link %>
     <% elsif @step_by_step_page.has_draft? %>
       <%= preview_link %>
     <% end %>
--- a/app/views/step_by_step_pages/show/_actions.erb
+++ b/app/views/step_by_step_pages/show/_actions.erb
@@ -14,7 +14,9 @@
         method: :post,
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
-    <% elsif @step_by_step_page.can_be_published? %>
+    <% elsif
+      @step_by_step_page.can_be_published? && !(current_user.has_permission? "Unreleased feature") || # remove this line when the feature flag is removed
+      @step_by_step_page.status.approved_2i? && (current_user.has_permission? "Unreleased feature") %>
       <% if @step_by_step_page.has_been_published? %>
         <%= render "govuk_publishing_components/components/button", {
           text: "Publish",
@@ -31,6 +33,13 @@
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
     <% elsif @step_by_step_page.has_draft? %>
+      <% if !@step_by_step_page.status.submitted_for_2i? &&
+        (current_user.has_permission? "Unreleased feature") # remove this line when the feature flag is removed %>
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Submit for 2i review",
+          href: step_by_step_page_submit_for_2i_path(@step_by_step_page)
+        } %>
+      <% end %>
       <%= preview_link %>
     <% end %>
 
--- a/app/views/step_by_step_pages/show/_required_actions.html.erb
+++ b/app/views/step_by_step_pages/show/_required_actions.html.erb
@@ -29,3 +29,9 @@
     </ul>
   <% end %>
 <% end %>
+
+<% if @step_by_step_page.status.submitted_for_2i? %>
+  <%= render "govuk_publishing_components/components/inset_text" do %>
+    <p class="govuk-body">Not yet claimed for 2i</p>
+  <% end %>
+<% end %>
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -66,6 +66,10 @@ FactoryBot.define do
     end
   end
 
+  factory :step_by_step_page_submitted_for_2i, parent: :step_by_step_page_with_steps do
+    status { "submitted_for_2i" }
+  end
+
   factory :step do
     title { "Check how awesome you are" }
     logic { "number" }
--- a/spec/features/create_step_by_step_spec.rb
+++ b/spec/features/create_step_by_step_spec.rb
@@ -9,6 +9,7 @@ RSpec.feature "Create new step by step page" do
 
   before do
     given_I_am_a_GDS_editor
+    given_I_can_access_unreleased_features
     setup_publishing_api
     stub_default_publishing_api_put_intent
   end
@@ -18,6 +19,7 @@ RSpec.feature "Create new step by step page" do
     and_I_fill_in_the_form
     and_I_see_a_page_created_success_notice
     and_I_see_I_saved_it_last
+    and_I_can_submit_the_step_by_step_for_2i_review
     and_I_can_preview_the_step_by_step
     when_I_visit_the_step_by_step_pages_index
     then_I_see_the_new_step_by_step_page
--- a/spec/features/reviewing_step_by_steps_spec.rb
+++ b/spec/features/reviewing_step_by_steps_spec.rb
@@ -15,6 +15,8 @@ RSpec.feature "Reviewing step by step pages" do
     when_I_visit_the_submit_for_2i_page
     and_I_submit_the_form
     then_I_see_a_submitted_for_2i_success_notice
+    and_I_see_a_not_yet_claimed_for_2i_prompt
+    and_I_cannot_see_a_submit_for_2i_button
     when_I_view_internal_change_notes
     then_I_can_see_an_automated_change_note
   end
@@ -25,13 +27,11 @@ RSpec.feature "Reviewing step by step pages" do
     and_I_fill_in_additional_comments
     and_I_submit_the_form
     then_I_see_a_submitted_for_2i_success_notice
+    and_I_see_a_not_yet_claimed_for_2i_prompt
+    and_I_cannot_see_a_submit_for_2i_button
     when_I_view_internal_change_notes
     then_I_can_see_an_automated_change_note
     and_I_can_see_additional_comments_in_the_change_note
-  end
-
-  def given_I_can_access_unreleased_features
-    stub_user.permissions << "Unreleased feature"
   end
 
   def when_I_visit_the_submit_for_2i_page
@@ -52,6 +52,16 @@ RSpec.feature "Reviewing step by step pages" do
 
   def then_I_see_a_submitted_for_2i_success_notice
     expect(page).to have_content("Step by step page was successfully submitted for 2i")
+  end
+
+  def and_I_see_a_not_yet_claimed_for_2i_prompt
+    expect(page).to have_css(".govuk-inset-text", text: "Not yet claimed for 2i")
+  end
+
+  def and_I_cannot_see_a_submit_for_2i_button
+    within(".app-side__actions") do
+      expect(page).to_not have_link("Submit for 2i review")
+    end
   end
 
   def then_I_can_see_an_automated_change_note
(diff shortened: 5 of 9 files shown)


spec/support/step_nav_steps.rb
```diff
@@ -187,6 +187,14 @@ def then_I_see_the_new_step_by_step_page
     expect(page).to have_content("How to bake a cake")
   end
 
+  def then_I_can_submit_the_step_by_step_for_2i_review
+    within(".app-side__actions") do
+      expect(page).to have_link("Submit for 2i review")
+    end
+  end
+
+  alias_method :and_I_can_submit_the_step_by_step_for_2i_review, :then_I_can_submit_the_step_by_step_for_2i_review
```
ChrisBAshton: This alias method isn't needed at this point in the git history - can just go straight to `def and_I_can_submit_the_step_by_step_for_2i_review`.

spec/features/create_step_by_step_spec.rb
```diff
@@ -18,6 +18,7 @@
     and_I_fill_in_the_form
     and_I_see_a_page_created_success_notice
     and_I_see_I_saved_it_last
+    and_I_can_submit_the_step_by_step_for_2i_review
```
ChrisBAshton: Out of scope for this PR maybe, but this seems an odd user journey (being able to submit a step by step for 2i as soon as it's been created). I wonder if there's a more suitable place to put this assertion... for example, after fixing issues identified in the inset-text component?

spec/support/common_feature_steps.rb
```diff
@@ -16,4 +16,8 @@ def given_I_am_not_a_GDS_editor
   def and_I_submit_the_form
     find('input[type=submit]').click
   end
+
+  def given_I_can_access_unreleased_features
+    stub_user.permissions << "Unreleased feature"
+  end
```
ChrisBAshton: Nice 👍

app/views/step_by_step_pages/show/_required_actions.html.erb
```diff
@@ -1,3 +1,9 @@
+<% if @step_by_step_page.status.submitted_for_2i? %>
+  <%= render "govuk_publishing_components/components/inset_text" do %>
+    <p class="govuk-body">Not yet claimed for 2i</p>
+  <% end %>
+<% end %>
```
ChrisBAshton: This will lead to having two inset prompts in the page, if for example it is submitted for 2i and then a step is tweaked. I can see why you've kept it separate, as we don't really want to rewrite the `should_show_required_prepublish_actions` wrapper on line 7.

Not sure if that's an issue from a user perspective - but it probably is a problem from the tests' perspective, which are only expecting one inset_text component at the moment and will lead to an "ambiguous match" if both are rendered. Other components let you pass an `id` to remove the ambiguity (see the `summary-list` usage in `show.html`) but I don't think we can do that out of the box here 🤔

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -50,6 +54,16 @@ def then_I_see_a_submitted_for_2i_success_notice
     expect(page).to have_content("Step by step page was successfully submitted for 2i")
   end
 
+  def and_I_see_a_not_yet_claimed_for_2i_prompt
+    expect(page).to have_css('.govuk-inset-text', text: "Not yet claimed for 2i")
+  end
+
+  def and_I_cannot_see_a_submit_for_2i_button
+    within(".app-side__actions") do
+      expect(page).to_not have_css("a", text: "Submit for 2i review")
```
ChrisBAshton: This can use the predicate matcher:

```ruby
      expect(page).to_not have_link("Submit for 2i review")
```

app/views/step_by_step_pages/show/_actions.erb
```diff
@@ -32,10 +30,13 @@
         step_by_step_page_schedule_path(@step_by_step_page),
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
-    <% elsif !@step_by_step_page.steps_have_content? %>
-      <p class="govuk-body">Step by steps cannot be published until all steps have content.</p>
-      <%= preview_link %>
     <% elsif @step_by_step_page.has_draft? %>
+      <% if !@step_by_step_page.status.submitted_for_2i? && (current_user.has_permission? "Unreleased feature") %>
```
AgaDufrat: Maybe it would be better to have this wrapped in a method and add unit tests as Leena suggested here: https://github.com/alphagov/collections-publisher/pull/678#discussion_r312082528

spec/support/step_nav_steps.rb
alex-ju: Removed this.

spec/features/create_step_by_step_spec.rb
alex-ju: I get your point, but the tests are quite messy and inconsistent at the moment (they're starting to look much better in the past few months though) since we check the actions separately based on document status I wouldn't stress to much on a common scenario at this point, but rather on sanity checks based on states.

app/views/step_by_step_pages/show/_required_actions.html.erb
alex-ju: As discussed, we'll keep two separated inset text (as per design and because of a lack in new copy to work with) but re-placed the 2i one after the required actions to make sure tests are failing.

spec/features/reviewing_step_by_steps_spec.rb
alex-ju: Updated as suggested.

app/views/step_by_step_pages/show/_actions.erb
alex-ju: Sounds like a good suggestion, but the `current_user` permissions will make these helper methods messy and will introduce more complexity in both code and tests. I tried to extract them on separate lines in the view files so we can remove them with the feature flag.

alex-ju:
@ChrisBAshton I merged #764 into this PR, but because of the missing `_spec` suffix the tests weren't picked up by the rake task so I couldn't see the tests were actually failing even in positive scenarios and that a few functions were missing. I tried to add to that logic and make them work.

This should be good for a second review.

spec/factories.rb
```diff
@@ -66,6 +66,11 @@
     end
   end
 
+  factory :step_by_step_page_submitted_for_2i, parent: :step_by_step_page_with_steps do
+    status { "submitted_for_2i" }
+    # must pass `review_requester_id` when invoking
+  end
```
ChrisBAshton: This can be removed as it's not used in this PR. My apologies 😅

app/views/step_by_step_pages/show/_actions.erb
```diff
@@ -14,7 +14,9 @@
         method: :post,
         class: %w(govuk-link govuk-link--no-visited-state app-link--right)
       %>
-    <% elsif @step_by_step_page.can_be_published? %>
+    <% elsif
+      @step_by_step_page.can_be_published? && !(current_user.has_permission? "Unreleased feature") || # remove this line when the feature flag is removed
+      @step_by_step_page.status.approved_2i? && (current_user.has_permission? "Unreleased feature") %>
```
ChrisBAshton: Can probably remove the `"Unreleased feature"` requirement from this line if we're capturing that on line 18 🤔

spec/features/step_by_step_actions_spec.rb
```diff
@@ -21,11 +21,34 @@
     end
   end
 
+  context "Step by step has been submitted for 2i" do
+    before do
+      given_there_is_a_step_by_step_with_status "submitted_for_2i"
+    end
+
+    scenario "show the relevant actions to the step by step author" do
```
ChrisBAshton: I'd add a step here to match the scenario, like:

`and_I_am_the_step_by_step_author`

...

```ruby
def and_I_am_the_step_by_step_author
  @current_user = @review_requester
end
```

spec/features/step_by_step_actions_spec.rb
```diff
@@ -21,11 +21,34 @@
     end
   end
 
+  context "Step by step has been submitted for 2i" do
+    before do
+      given_there_is_a_step_by_step_with_status "submitted_for_2i"
```
ChrisBAshton: I think while we only have one scenario to test, we can put this step directly in the scenario and not in a `before` block.

spec/features/step_by_step_actions_spec.rb
```diff
@@ -21,11 +21,34 @@
...
+      when_I_visit_the_step_by_step_page
+      then_there_should_be_no_primary_action
+      and_the_secondary_action_should_be "Preview"
+      and_there_should_be_tertiary_actions_to %w(Delete)
+    end
+  end
+
+  def given_there_is_a_step_by_step_with_status(status)
+    @review_requester = create(:user, name: "Original Author")
+    @reviewer = create(:user, name: "2i reviewer")
```
ChrisBAshton: This line isn't used yet (but agree it will be useful when we continue adding 2i features). I'd be tempted to leave it out for now.

spec/support/step_nav_steps.rb
```diff
@@ -106,6 +106,12 @@ def given_there_is_a_step_by_step_page_with_broken_links_and_unpublished_changes
     @step_by_step_page = create(:step_by_step_page, steps: [step], draft_updated_at: Time.zone.now, slug: "step-by-step-with-broken-links-and-unpublished-changes")
   end
 
+  def given_there_is_a_step_by_step_page_submitted_for_2i
+    step = create(:step)
+    stub_link_checker_report_success(step)
+    @step_by_step_page = create(:step_by_step_page_submitted_for_2i, steps: [step], slug: "step-by-step-with-link-report")
+  end
```
ChrisBAshton: This doesn't appear to be being used anywhere (and also doesn't appear to create a step by step page with `submitted_for_2i` status). Can it be removed?

spec/features/step_by_step_actions_spec.rb
alex-ju: I tried without it and I get
`ActiveRecord::RecordInvalid:
  Validation failed: Status submitted_for_2i, requires a draft and review_requester_id to be present`

ChrisBAshton: It needs `@review_requester` but it shouldn't need `@reviewer` 🤔 can I just confirm you tried deleting line 39?

app/views/step_by_step_pages/show/_actions.erb
alex-ju: I tried to split the conditional statement making sure we're showing the 'Publish' button for users without "Unreleased feature" permission when a page can be published and we're only showing it when the status is `approved_2i` for users without "Unreleased feature" permission. This way it's easier to remove the first line with the feature flag without re-thinking all the logic.

spec/factories.rb
alex-ju: Removed

spec/features/step_by_step_actions_spec.rb
alex-ju: Updated as suggested.

spec/features/step_by_step_actions_spec.rb
alex-ju: Updated as suggested

spec/support/step_nav_steps.rb
alex-ju: Yes, this is not used anymore as we can now create a `:step_by_step_page_submitted_for_2i`. Removed.

spec/features/step_by_step_actions_spec.rb
alex-ju: Thanks for helping with this @ChrisBAshton, as it seems along with removing `@reviewer` I remove the `review_requester_id` too. Now it's all good, tests are passing.

app/views/step_by_step_pages/show/_actions.erb
ChrisBAshton: Ahh I didn't spot that one was `!` and one wasn't. That makes sense now, thanks

spec/factories.rb
ChrisBAshton: Have spoken IRL 👍 I didn't mean the comment, I meant this entire factory, but I can see it's actually being used in the next commit, so we will need to keep it. Can it be moved to the commit where it is used?

spec/features/step_by_step_actions_spec.rb
```diff
@@ -21,11 +21,31 @@
     end
   end
 
+  context "Step by step has been submitted for 2i" do
+    scenario "show the relevant actions to the step by step author" do
+      given_there_is_a_step_by_step_with_status "submitted_for_2i"
```
ChrisBAshton: I know I wrote the original pseudocode for this, but am having second thoughts about the need to pass a parameter. I think we should just use `given_there_is_a_step_by_step_that_has_been_submitted_for_2i`, adding it to `step_nav_steps.rb` in the same style as the others (i.e. taking lines 35-38, renaming it and moving it).

If it starts to become hard to understand exactly what the status is from the method name alone, we can move to this parameterised approach in future.

alex-ju: Sure, update. It makes sense to live next to the other helpers which provide us with a step-by-step in a certain state.

PR #771, opened 2019-09-25
Fix ordering of steps

ChrisBAshton:
Adding new steps would not set the `position` property - this was only being set when explicitly reordering steps from the reorder screen - which led to strange bugs such as new steps being added to the top of the step by step.

We now set `position` to the "number of steps in the step by step + 1" so that it is added to the end. We also reset the positions if a step is deleted, to prevent step numbers becoming out of sync with how many steps there are.

I had hoped to implement this at the schema level with an `auto_increment` on the `positions` field, however this is not allowed on non-primary key fields, and in any case would not work properly as there are multiple step by steps (so incrementing would apply based on the number of steps in the database, irrespective of which step by step the steps are a part of).

Trello: https://trello.com/c/GkAjQcLt/26-ordering-of-new-steps-is-inconsistent-after-x-amount-of-steps

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -3,6 +3,7 @@ class Step < ApplicationRecord
   validates_presence_of :step_by_step_page
   validates :title, :logic, presence: true
   has_many :link_reports, :dependent => :destroy
+  after_create :set_step_position
 
   def broken_links?
     broken_links.present? && broken_links.any?
@@ -48,5 +49,9 @@ private
 
   def most_recent_batch
     LinkReport.where(step_id: self.id).last
+  end
+
+  def set_step_position
+    update!(position: step_by_step_page.steps.count)
   end
 end
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -4,6 +4,7 @@ class Step < ApplicationRecord
   validates :title, :logic, presence: true
   has_many :link_reports, :dependent => :destroy
   after_create :set_step_position
+  after_destroy :set_parent_step_positions
 
   def broken_links?
     broken_links.present? && broken_links.any?
@@ -53,5 +54,9 @@ private
 
   def set_step_position
     update!(position: step_by_step_page.steps.count)
+  end
+
+  def set_parent_step_positions
+    step_by_step_page.make_step_positions_sequential
   end
 end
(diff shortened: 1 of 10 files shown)


spec/support/publishing_api_helpers.rb
```diff
@@ -25,6 +25,80 @@ def publishing_api_has_no_linked_items
...
+
+  def publishing_api_receives_get_content_id_request(content_items:)
+    content_items.each do |content_item|
+      expect(Services.publishing_api).to(
+        receive(:get_content).with(content_item[:content_id]),
+      ).and_return(content_item)
+    end
+  end
+
+  def step_link_fixtures_return_data
```
brucebolt: This (and `step_link_fixtures_content_items`) feel like they should remain in `StepLinksForRules`, since they are specifically related to step links, rather than anything generic to Publishing API.

app/models/step.rb
```diff
@@ -49,4 +50,8 @@ def batch_link_report_id
   def most_recent_batch
     LinkReport.where(step_id: self.id).last
   end
+
+  def set_step_position
+    update!(position: step_by_step_page.steps.count)
```
brucebolt: Would it be better to use `step_by_step_page.steps.maximum(:position) + 1` to be absolutely sure you are incrementing the highest step by one?

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -548,6 +555,34 @@ def then_I_am_told_that_it_is_published
...
+      content_items: step_link_fixtures_content_items,
+    )
+    visit step_by_step_page_reorder_path(@step_by_step_page)
+    find(".js-reorder:first-child .js-down").click
+    click_button "Save"
+    expect(page).to have_content "Steps were successfully reordered."
+    and_I_should_be_on_the_step_by_step_page
+  end
+
+  def when_I_add_another_step
```
brucebolt: Could this be renamed `when_I_add_a_step` since it isn't dependent on there already being steps?

app/models/step.rb
ChrisBAshton: Hmm I've tried updating as suggested, but this leads to a _lot_ of test failures. I think it's to do with how Rails factories create things.

For example, an assertion that the first step has position 1 works, but an assertion that the next step has position 2 fails (it's actually 3!). Haven't figured out why it's doing that yet.
I also had to modify the code to get the 'position 1' example to work, as `step_by_step_page.steps.maximum(:position)` returns `nil` if there are no existing steps.

I'm inclined to leave this and rely on the existing logic to keep things in order.

PR #774, opened 2019-09-26
Update change note title and entry

brucebolt:
Currently, we are displaying `Internal note` as the headline for all history entries.  A smaller description text explains the actual change (e.g. `Published`).

There are a number of issues with this:
- It is more difficult to users to scan through for specific status changes
- The author is duplicated (their name is appended to the description field, in addition to appearing after the timestamp)

This update adds a headline field to the internal change note table.  This is populated for each history entry with either a status change or where in the workflow it is.  All existing/historic history entries will retain their current style (i.e. labelled as `Internal note`).

Additionally, there are a number of copy changes detailed in a [Google document](https://docs.google.com/document/d/1pvx2HGX2y71voArlqvRt8lPc6hIRkFxcPlRIYnHfFlg/edit) that have been implemented as part of this change.

Finally, a number of missing change note entries have been added, as detailed in the document showing copy changes: when a draft is discarded, and when a step by step is unpublished.

Status entries before making these changes:
Screen Shot 2019-09-26 at 10 03 53

Status changes having performed the same actions after this change:
Screen Shot 2019-09-26 at 09 54 55

Trello card: https://trello.com/c/xdIoXx8s/81-update-change-note-titles-and-entry-stretch

PR #776, opened 2019-09-26
Only approved_2i step by steps can be published or scheduled

AgaDufrat:
**Step by step in review, not approved (reviewer): **
<kbd>
reviewer
</kbd>

**Step by step in review, not approved (review requestor): **
<kbd>
requestor
</kbd>

**Step by step approved:**
<kbd>approved_2i</kbd>

**TO DO** in a separate PR after feature flag is removed: 
*(WIP in improve-2i-publish-schedule branch)*
- Add `approved_2i?` condition to `can_be_published` method in the model and add unit tests
- Publish and schedule routes should not be usable unless SBS is in an approved_2i state

[Trello card](https://trello.com/c/yvGd29qp/74-only-approved-step-by-steps-can-be-published-or-scheduled)

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -75,7 +75,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def mark_as_unscheduled
-    update_attribute(:status, "draft")
+    update_attribute(:status, "approved_2i")
   end
 
   def self.validate_redirect(redirect_url)
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -151,7 +151,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def links_checked_since_last_update?
-    links_checked? && links_last_checked_date > draft_updated_at
+    (links_checked? && links_last_checked_date > draft_updated_at) || status.approved_2i?
   end
 
   def links_checked?
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -65,8 +65,8 @@ class StepByStepPage < ApplicationRecord
   def mark_as_unpublished
     update(
       published_at: nil,
-      draft_updated_at: nil,
-      status: "draft",
+      draft_updated_at: Time.zone.now,
+      status: "approved_2i",
     )
   end
 
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -49,6 +49,10 @@ class StepByStepPage < ApplicationRecord
     update_attribute(:draft_updated_at, nil)
   end
 
+  def mark_as_approved_2i
+    update_attribute(:status, "approved_2i")
+  end
+
   def mark_as_published
     now = Time.zone.now
     update(
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -143,7 +143,7 @@ class StepByStepPage < ApplicationRecord
   end
 
   def should_show_required_prepublish_actions?
-    has_draft? && !scheduled_for_publishing? && !can_be_published?
+    (has_draft? && !scheduled_for_publishing? && !can_be_published?) || status.in_review?
   end
 
   def broken_links_found?
--- a/app/validators/status_prerequisite_validator.rb
+++ b/app/validators/status_prerequisite_validator.rb
@@ -3,7 +3,7 @@ class StatusPrerequisiteValidator < ActiveModel::EachValidator
     if value == "approved_2i"
       return true if can_be_2i_approved?(record)
 
-      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review"
+      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review or scheduled"
     end
 
     if value == "in_review"
@@ -43,6 +43,9 @@ private
 
   def can_be_2i_approved?(record)
     record.has_draft? &&
-      record.status_was == "in_review"
+      (
+        record.status_was == "in_review" ||
+        record.status_was == "scheduled"
+      )
   end
 end
--- a/app/validators/status_prerequisite_validator.rb
+++ b/app/validators/status_prerequisite_validator.rb
@@ -3,7 +3,7 @@ class StatusPrerequisiteValidator < ActiveModel::EachValidator
     if value == "approved_2i"
       return true if can_be_2i_approved?(record)
 
-      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review or scheduled"
+      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review, scheduled or published"
     end
 
     if value == "in_review"
@@ -42,10 +42,7 @@ private
   end
 
   def can_be_2i_approved?(record)
-    record.has_draft? &&
-      (
-        record.status_was == "in_review" ||
-        record.status_was == "scheduled"
-      )
+    allowed_previous_statuses = %w(in_review scheduled published)
+    record.has_draft? && allowed_previous_statuses.include?(record.status_was)
   end
 end
--- a/app/validators/status_prerequisite_validator.rb
+++ b/app/validators/status_prerequisite_validator.rb
@@ -3,13 +3,13 @@ class StatusPrerequisiteValidator < ActiveModel::EachValidator
     if value == "approved_2i"
       return true if can_be_2i_approved?(record)
 
-      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review, scheduled or published"
+      record.errors[attribute] << "#{value}, requires a draft, a reviewer and for status to be in_review, scheduled, published or approved_2i"
     end
 
     if value == "in_review"
       return true if can_be_in_review?(record)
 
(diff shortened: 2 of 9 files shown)


app/models/step_by_step_page.rb
```diff
@@ -75,7 +75,7 @@ def mark_as_scheduled
   end
 
   def mark_as_unscheduled
-    update_attribute(:status, "draft")
+    update_attribute(:status, "approved_2i")
```
ChrisBAshton: To future-proof this, I'd be tempted to `update_attribute(:status, status_was)`. I can't think of a situation where we'd want to go to anything other than `approved_2i` at the moment, but if/when we implement fact-check, which may or may not be an optional workflow, `approved_2i` may not always be correct. What we're effectively doing is hitting 'Undo' and we want to go back to the previous state.

app/models/step_by_step_page.rb
```diff
@@ -63,10 +63,11 @@ def mark_as_published
   end
 
   def mark_as_unpublished
+    now = Time.zone.now
     update(
       published_at: nil,
-      draft_updated_at: nil,
-      status: "draft",
+      draft_updated_at: now,
```
ChrisBAshton: I'm not sure why this commit is necessary - could you expand on the commit message a little?

If anything, I'd think the `draft_updated_at` should be left alone, as we have a "Check for broken links" prompt whenever `draft_updated_at` is more recent than the `links_last_checked` date (except for when step by steps have _just_ been scheduled/published).

app/models/step_by_step_page.rb
```diff
@@ -26,6 +26,10 @@ class StepByStepPage < ApplicationRecord
 
   scope :by_title, -> { order(:title) }
 
+  def approved_2i?
+    status.approved_2i?
```
ChrisBAshton: I think elsewhere we're just calling `step_by_step_page.status.approved_2i?` directly - not sure a helper method is necessary. If it is, could you explain why in the commit?

spec/validators/status_prerequisite_validator_spec.rb
```diff
@@ -72,7 +72,7 @@
       expect(step_by_step_page).to be_valid
     end
 
-    it "allows status to be approved_2i if the current status is published" do
+    it "allows status to be approved_2i if the current status is published and there is a draft" do
```
ChrisBAshton: Could you fix this up to the commit when you defined this `it` block?

app/validators/status_prerequisite_validator.rb
```diff
@@ -46,7 +49,8 @@ def can_be_2i_approved?(record)
       (
        record.status_was == "in_review" ||
        record.status_was == "scheduled" ||
-       record.status_was == "published"
+       record.status_was == "published" ||
+       record.status_was == "approved_2i"
```
ChrisBAshton: This could now be refactored to something like:

```ruby
allowed_previous_statuses = %w(in_review scheduled published approved_2i)
allowed_previous_statuses.include? record.status_was
```

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -96,6 +96,19 @@
     end
   end
 
+  describe "#unpublish" do
```
brucebolt: You might need to rebase this against master, since I've also created tests for unpublishing in https://github.com/alphagov/collections-publisher/pull/774 😁

app/views/step_by_step_pages/show/_required_actions.html.erb
```diff
@@ -34,4 +34,8 @@
   <%= render "govuk_publishing_components/components/inset_text" do %>
     <p class="govuk-body">Not yet claimed for 2i</p>
   <% end %>
+<% elsif @step_by_step_page.status.in_review? %>
+    <%= render "govuk_publishing_components/components/inset_text" do %>
+      <p class="govuk-body">Step by steps cannot be published until it is 2i approved.</p>
```
brucebolt: Could I propose:
```
Step by step cannot be published until it is 2i approved.
```

spec/features/managing_step_by_step_pages_spec.rb
```diff
@@ -320,14 +337,53 @@
...
-    given_there_is_a_step_that_has_no_broken_links
+    given_there_is_an_approved_2i_step_that_has_no_broken_links
     when_I_view_the_step_by_step_page
     then_there_should_be_no_inset_prompt
     and_when_I_click_button "Publish"
     then_I_am_told_that_it_is_published
     and_there_should_continue_to_be_no_inset_prompt
   end
 
+  context "No Unreleased feature permissions" do
```
brucebolt: I like having tests for both scenarios 👍

spec/validators/status_prerequisite_validator_spec.rb
```diff
@@ -27,7 +27,56 @@
       expect(step_by_step_page.errors.messages[:status]).to eq([error_message])
     end
 
-    it "does not allow status to be in_review if the current status is not submitted_for_2i" do
+    it "does not allow status to be in_review if the current status is approved_2i" do
```
brucebolt: There's a lot of scenarios here that are using the same code, but just with a different status value.

Could it be iterated over, like we've done with the user permission tests, e.g. https://github.com/alphagov/collections-publisher/blob/a167d3233451f326689c72d65a110f737f1e1b40/spec/controllers/review_controller_spec.rb#L98-L106

app/models/step_by_step_page.rb
AgaDufrat: I talked about this with Bruce and we weren't sure how long/well `status_was` persists. We were worried that if the app is restarted this information may be lost.
I see what you mean as I imagined we will have a general `approve` status that would require step by step to be both `approved_2i` and `fact_checked` so we would need to change this to `approved` in the future. 
I will read about this and if it's safe enough we should do it.

app/models/step_by_step_page.rb
AgaDufrat: If draft_updated_at is set to nil when we unpublish, it will cause has_draft? to return false but approved_2i requires a step by step to have a draft. Consequently, it will not be possible to unpublish step by step. 

It currently prompts to check for broken links once unpublished. What are your thoughts on adding `|| status.approved_2i?` to `links_checked_since_last_update?` ? 
Would it be ok, not to prompt to check for broken links if the SBS is `approved_2i`? 🤔 

We are showing the prompt when it's a `draft` and `submitted_for_2i` so given that the reviewers check the links manually as well, I would be tempted to assume that the links are correct at that point.

app/views/step_by_step_pages/show/_required_actions.html.erb
AgaDufrat: I changed it in line with Alex and Chris' suggestion (https://gds.slack.com/archives/CLJCUGVD2/p1569511783011000) and updated the screenshots.

app/models/step_by_step_page.rb
ChrisBAshton: Agreed 👍 and thanks for updating the commit message.

Now your first two commit messages have the same headline though 🤔

AgaDufrat: First is Un**scheduled**, second Un**published** 😉

ChrisBAshton: D'oh! 🤦‍♂

app/models/step_by_step_page.rb
```diff
@@ -63,10 +63,11 @@ def mark_as_published
   end
 
   def mark_as_unpublished
+    now = Time.zone.now
```
brucebolt: Since `now` is only being used once, I'd suggest just using `Time.zone.now` directly, e.g.

`draft_updated_at: Time.zone.now,`

app/models/step_by_step_page.rb
```diff
@@ -144,7 +144,7 @@ def links_last_checked_date
   end
 
   def should_show_required_prepublish_actions?
-    has_draft? && !scheduled_for_publishing? && !can_be_published?
+    has_draft? && !scheduled_for_publishing? && !can_be_published? || status.in_review?
```
brucebolt: I'd recommend adding some parentheses here to make it very clear which conditions are associated with the AND and which are associated with the OR.

spec/models/step_by_step_page_spec.rb
```diff
@@ -257,6 +257,13 @@
 
       expect(step_by_step_with_step.should_show_required_prepublish_actions?).to be true
     end
+
+    it "should return true if step by step is in review" do
+      step_by_step_with_step = create(:step_by_step_page)
+      allow(step_by_step_page).to receive(:status). and_return("in_review")
```
brucebolt: There's a space before the `and_return` which probably shouldn't be there.

spec/validators/status_prerequisite_validator_spec.rb
```diff
@@ -79,6 +93,14 @@
 
       expect(step_by_step_page).to be_valid
     end
+
+    it "allows status to be approved_2i if the current status is approved_2i and there is a draft" do
+      allow(step_by_step_page).to receive(:has_draft?).and_return(true)
+      allow(step_by_step_page).to receive(:status_was). and_return("approved_2i")
+      step_by_step_page.status = "approved_2i"
```
brucebolt: There's a space before the `and_return` which probably shouldn't be there.

PR #779, opened 2019-09-27
Allow access to the reviewer and review requester details

brucebolt:
This enables access to details about the reviewer and review requester, e.g. to get the reviewer name:

```
step_by_step_page.reviewer.name
```

Trello card: https://trello.com/c/Fwo562zl/72-2i-reviewer-can-claim-sbs-for-review

### DIFF
--- a/spec/models/step_by_step_page_spec.rb
+++ b/spec/models/step_by_step_page_spec.rb
@@ -20,6 +20,29 @@ RSpec.describe StepByStepPage do
     end
   end
 
+  describe "associations" do
+    let(:stub_user) { create(:user, name: "Author Name") }
+
+    before do
+      step_by_step_page.review_requester_id = stub_user.uid
+    end
+
+    it "should allow return of review requester name" do
+      step_by_step_page.save
+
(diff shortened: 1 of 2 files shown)


spec/models/step_by_step_page_spec.rb
```diff
@@ -20,6 +20,29 @@
...
+      expect(step_by_step_page.review_requester.name).to eq stub_user.name
+    end
+
+    it "should allow return of reviewer name" do
+      reviewer_user = create(:user)
+      step_by_step_page.reviewer_id = reviewer_user.uid
+
+      step_by_step_page.save
+
+      expect(step_by_step_page.review_requester.name).to eq reviewer_user.name
```
ChrisBAshton: Shouldn't this be?

```suggestion
      expect(step_by_step_page. reviewer.name).to eq reviewer_user.name
```

brucebolt: Yes, it should be.

I didn't catch this because both users had a name of `nil`.

PR #780, opened 2019-09-27
2i reviewer can approve and request changes on step by steps

ChrisBAshton:
Trello: https://trello.com/c/2hmErFyt/75-2i-reviewer-can-approve-and-request-changes-on-step-by-steps

- [x] Allow "Approve" action to take changenotes
- [x] 2i reviewers can see the Approve/Request Changes buttons
- [x] 2i reviewers can approve
- [x] 2i reviewers can request changes
~Write `step_by_step_actions_spec.rb` tests for `2i_approved` state.~ Will do this [later](https://trello.com/c/Qle5c1QQ/43-write-tests-display-the-appropriate-buttons-depending-on-the-state-of-the-sbs-and-who-is-logged-in).

NB, uses copy found in the sketch attached to the ["2i reviewer can claim SBS for review" card](https://trello.com/c/Fwo562zl/72-2i-reviewer-can-claim-sbs-for-review).

### DIFF
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -1,6 +1,7 @@
 class ReviewController < ApplicationController
   layout "admin_layout"
 
+  before_action :set_step_by_step_page
   before_action :require_gds_editor_permissions!
   before_action :require_unreleased_feature_permissions!
   before_action :require_2i_reviewer_permissions!, only: %i(
@@ -10,7 +11,12 @@ class ReviewController < ApplicationController
     show_approve_2i_review_form
     show_request_change_2i_review_form
   )
-  before_action :set_step_by_step_page
+  before_action :require_user_to_be_the_2i_reviewer!, only: %i(
+    approve_2i_review
+    request_change_2i_review
+    show_approve_2i_review_form
+    show_request_change_2i_review_form
+  )
 
   def show_approve_2i_review_form
     render :submit_2i_verdict, locals: { approved: true }
@@ -108,5 +114,9 @@ private
 
   def set_step_by_step_page
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
+  end
+
+  def require_user_to_be_the_2i_reviewer!
+    render "shared/forbidden", status: :forbidden unless current_user.uid == @step_by_step_page.reviewer_id
   end
 end
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -16,7 +16,7 @@ class ReviewController < ApplicationController
     )
       generate_change_note("2i approved")
 
-      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully approved_2i."
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved."
     end
   end
 
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -14,7 +14,7 @@ class ReviewController < ApplicationController
       reviewer_id: nil,
       status: status,
     )
-      generate_change_note("2i approved")
+      generate_change_note("2i approved", params[:additional_comment])
 
       redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved."
     end
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -3,8 +3,18 @@ class ReviewController < ApplicationController
 
   before_action :require_gds_editor_permissions!
   before_action :require_unreleased_feature_permissions!
-  before_action :require_2i_reviewer_permissions!, only: %i(approve_2i_review claim_2i_review request_change_2i_review)
+  before_action :require_2i_reviewer_permissions!, only: %i(
+    approve_2i_review
+    claim_2i_review
+    request_change_2i_review
+    show_approve_2i_review_form
+    show_request_change_2i_review_form
+  )
   before_action :set_step_by_step_page
+
+  def show_approve_2i_review_form
+    render :submit_2i_verdict, locals: { approved: true }
+  end
 
   def approve_2i_review
     status = "approved_2i"
@@ -17,6 +27,8 @@ class ReviewController < ApplicationController
       generate_change_note("2i approved", params[:additional_comment])
 
       redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved."
+    else
+      render :submit_2i_verdict, locals: { approved: true }, status: :unprocessable_entity
     end
   end
 
@@ -33,6 +45,10 @@ class ReviewController < ApplicationController
     end
   end
 
+  def show_request_change_2i_review_form
+    render :submit_2i_verdict, locals: { approved: false }
+  end
+
   def request_change_2i_review
     status = "draft"
 
@@ -44,6 +60,8 @@ class ReviewController < ApplicationController
       generate_change_note("2i changes requested", params[:requested_change])
 
       redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested."
+    else
+      render :submit_2i_verdict, locals: { approved: false }, status: :unprocessable_entity
     end
   end
 
--- a/app/controllers/review_controller.rb
+++ b/app/controllers/review_controller.rb
@@ -32,7 +32,7 @@ class ReviewController < ApplicationController
     )
       generate_change_note("2i approved", params[:additional_comment])
 
-      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved."
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved. Please let the author know."
     else
       render :submit_2i_verdict, locals: { approved: true }, status: :unprocessable_entity
     end
@@ -65,7 +65,7 @@ class ReviewController < ApplicationController
     )
       generate_change_note("2i changes requested", params[:requested_change])
 
-      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested."
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested. Please let the author know."
     else
       render :submit_2i_verdict, locals: { approved: false }, status: :unprocessable_entity
     end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,6 +4,7 @@ Rails.application.routes.draw do
   root to: redirect("/step-by-step-pages", status: 302)
 
   resources :step_by_step_pages, path: "step-by-step-pages" do
+    get "approve-2i-review", to: "review#show_approve_2i_review_form"
     post "approve-2i-review", to: "review#approve_2i_review"
     post :check_links
     post "claim-2i-review", to: "review#claim_2i_review"
@@ -15,6 +16,7 @@ Rails.application.routes.draw do
     post :publish
     get :reorder
     post :reorder
+    get "request-change-2i-review", to: "review#show_request_change_2i_review_form"
     post "request-change-2i-review", to: "review#request_change_2i_review"
     post :revert
     post "revert-to-draft", to: "review#revert_to_draft"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,7 +7,7 @@ Rails.application.routes.draw do
     post "approve-2i-review", to: "review#approve_2i_review"
     post :check_links
     post "claim-2i-review", to: "review#claim_2i_review"
-    get "internal-change-notes", to: "interal_change_notes"
+    get "internal-change-notes"
     post "internal-change-notes", to: "internal_change_notes#create"
     get "navigation-rules", to: "navigation_rules#edit"
     put "navigation-rules", to: "navigation_rules#update"
@@ -20,7 +20,7 @@ Rails.application.routes.draw do
     post "revert-to-draft", to: "review#revert_to_draft"
     get :schedule
     post :schedule
-    post "schedule-datetime", to: "schedule_datetime"
+    post "schedule-datetime"
     get "submit-for-2i", to: "review#submit_for_2i"
     post "submit-for-2i", to: "review#submit_for_2i"
     get :unpublish
--- a/spec/controllers/review_controller_spec.rb
+++ b/spec/controllers/review_controller_spec.rb
@@ -84,6 +84,7 @@ RSpec.describe ReviewController do
 
       before do
         step_by_step_page.update_attributes(:status => "in_review", :reviewer_id => reviewer_user.uid)
+        stub_user.uid = reviewer_user.uid
       end
 
       required_permissions = ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"]
@@ -104,6 +105,14 @@ RSpec.describe ReviewController do
 
             expect(response.status).to eq(403)
           end
+        end
+
+        it "cannot be accessed by users other than the reviewer, even with the necessary permissions" do
+          stub_user.permissions = required_permissions
+          stub_user.uid = SecureRandom.uuid
+          get :approve_2i_review, params: { step_by_step_page_id: step_by_step_page.id }
+
+          expect(response.status).to eq(403)
         end
       end
 
--- a/spec/controllers/review_controller_spec.rb
+++ b/spec/controllers/review_controller_spec.rb
@@ -88,6 +88,44 @@ RSpec.describe ReviewController do
 
       required_permissions = ["signin", "GDS Editor", "Unreleased feature", "2i reviewer"]
 
+      describe "GET approve 2i review" do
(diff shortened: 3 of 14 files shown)


config/routes.rb
```diff
@@ -7,7 +7,7 @@
     post "approve-2i-review", to: "review#approve_2i_review"
     post :check_links
     post "claim-2i-review", to: "review#claim_2i_review"
-    get "internal-change-notes", to: "interal_change_notes"
+    get "internal-change-notes", to: "internal_change_notes"
```
ChrisBAshton: Not sure how tests passed - we have some that call `when_I_visit_the_change_notes_tab` that I would expect to crash the service. One to look at later?

app/views/review/_submit_2i_verdict.erb
```diff
@@ -0,0 +1,33 @@
+<%
+  title = approved ? 'Approve step by step' : 'Request changes to step by step'
```
ChrisBAshton: Move this to the commit where Request Changes is implemented.

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -34,6 +40,16 @@
...
   end
 
+  scenario "2i reviewer approves step by step" do
+    given_there_is_a_step_by_step_that_has_been_claimed_for_2i
+    and_I_am_the_reviewer
+    when_I_visit_the_step_by_step_page
+    when_I_click_button "Approve"
+    then_the_page_should_have_title_caption "Approve step by step"
+    and_when_I_click_button "Yes, approve 2i"
+    then_I_can_see_a_success_message "Step by step page was successfully 2i approved."
```
ChrisBAshton: Need to assert the status has changed too.

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -50,6 +50,16 @@
...
   end
 
+  scenario "2i reviewer requests changes to a step by step" do
+    given_there_is_a_step_by_step_that_has_been_claimed_for_2i
+    and_I_am_the_reviewer
+    when_I_visit_the_step_by_step_page
+    when_I_click_button "Request changes"
+    then_the_page_should_have_title_caption "Request changes to step by step"
+    and_when_I_click_button "Send change request"
+    then_I_can_see_a_success_message "Changes to the step by step page were requested."
```
ChrisBAshton: And here.

spec/controllers/review_controller_spec.rb
```diff
@@ -88,6 +88,44 @@
...
 
+      describe "GET approve 2i review" do
+        it "can be accessed by the reviewer, if they have GDS editor, Unreleased feature and 2i reviewer permissions" do
+          stub_user.permissions = required_permissions
+          get :approve_2i_review, params: { step_by_step_page_id: step_by_step_page.id }
+
+          expect(response.status).to eq(200)
+        end
+
+        (required_permissions - %w(signin)).each do |required_permission|
```
kevindew: Generally (I did try find a good source but Google failed me, so you'll have to just trust me :) ) it's not considered a best practice to dynamically create expectations at runtime with loops. This would probably be more commonly handled via a shared expectation.

Though I do wonder here what we're actually testing here, I wouldn't have thought we'd need to test every permission returns a 403 as that would be us using the test to confirm that the configuration of permissions in our controller matches the one in the test spec. Whereas I imagine we just need one test that confirms that we can return a 403?

app/controllers/review_controller.rb
```diff
@@ -95,4 +96,8 @@ def generate_change_note(status, change_note = nil)
   def set_step_by_step_page
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
   end
+
+  def require_user_to_be_the_2i_reviewer!
+    head :forbidden unless current_user.uid == @step_by_step_page.reviewer_id
```
kevindew: I assume this means a user gets a blank response? can we can show the users an error page.

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -55,6 +69,24 @@ def when_I_view_the_step_by_step_page
     visit step_by_step_page_path(@step_by_step_page)
   end
 
+  def then_the_page_should_have_title_caption(caption)
```
kevindew: I'd suggest that you minimise the arguments that are passed through to the test steps. I'm wondering if this was to be done to help with reusability? 

With this style of test I'd suggest that steps should not be optimised for re-usability. I think the steps should be very tailored to the situation for maximum readability and if needs be a step can call a helper method. 

Apologies if my opinions here conflict with conventions deeply rooted in this app, but thought I'd bring it up as it looks like this app is subtly drifting from the more common pattern. Happy to chat about this.

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -43,6 +44,32 @@
...
+    given_there_is_a_step_by_step_that_has_been_claimed_for_2i
+    and_I_am_the_reviewer
+    when_I_visit_the_step_by_step_page
+    when_I_click_button "Approve"
+    then_the_page_should_have_title_caption "Approve step by step"
+    and_I_fill_in_an_optional_note
+    and_when_I_click_button "Yes, approve 2i"
+    then_I_can_see_a_success_message "Step by step page was successfully 2i approved."
+    and_the_step_by_step_status_should_be "Approved 2i"
+    and_I_should_see_my_optional_note_in_the_changenotes
```
kevindew: should this be change_notes?

config/routes.rb
ChrisBAshton: I did, and it didn't!

spec/features/step_by_step_actions_spec.rb
```diff
@@ -55,6 +54,17 @@
...
+  context "Step by step has been claimed for 2i" do
+    scenario "show the relevant actions to the step by step author" do
+      given_there_is_a_step_by_step_that_has_been_claimed_for_2i
+      and_I_am_the_step_by_step_author
+      when_I_visit_the_step_by_step_page
+      then_there_should_be_no_primary_action
+      and_the_secondary_action_should_be "Preview"
+      and_there_should_be_tertiary_actions_to %w(Delete)
+    end
+  end
```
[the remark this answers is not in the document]
ChrisBAshton: Hmmm, looking at it with fresh eyes I agree it's confusing. It makes more sense when we have multiple scenarios inside the context, such as in "Step by step has been submitted for 2i" - this is the context, and the scenarios are the author and the reviewer respectively looking at the step by step. Where there is only one scenario it looks a little odd, but I do add an additional scenario in a later commit.

spec/controllers/review_controller_spec.rb
```diff
@@ -88,231 +88,187 @@
...
+          expected_change_note = "Reverted to draft"
 
-        post :revert_to_draft, params: { step_by_step_page_id: step_by_step_page.id }
-        step_by_step_page.reload
+          post :revert_to_draft, params: { step_by_step_page_id: step_by_step_page.id }
+          step_by_step_page.reload
 
-        expect(step_by_step_page.internal_change_notes.first.headline).to eq expected_change_note
+          expect(step_by_step_page.internal_change_notes.first.headline).to eq expected_change_note
+        end
```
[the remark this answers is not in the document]
ChrisBAshton: Probably worth viewing this commit with `?w=1` at the end (https://github.com/alphagov/collections-publisher/pull/780/commits/d11716fad9dc264306011a60ee704266f0949f6a?w=1), so that whitespace diffs are ignored.

There were a number of tests which set up the step by step to get it into an "in_review" state, which became noisy and difficult to read. Refactored so that the setup only happens once and all of the tests are now in a "this is already in_review" context.

config/routes.rb
```diff
@@ -4,6 +4,7 @@
   root to: redirect("/step-by-step-pages", status: 302)
 
   resources :step_by_step_pages, path: "step-by-step-pages" do
+    get "approve-2i-review", to: "review#approve_2i_review"
     post "approve-2i-review", to: "review#approve_2i_review"
```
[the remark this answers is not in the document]
ChrisBAshton: Have discussed IRL. Think it's best to encapsulate all the 2i review methods in one controller, rather than several small controllers - but happy to rename the methods.

spec/controllers/review_controller_spec.rb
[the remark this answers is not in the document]
ChrisBAshton: Have spoken in IRL and you've confirmed this makes sense now. We've agreed this commit leaves the code in a better place than it was in, and that the file in its entirety could do with a refactor in a future PR.

app/controllers/review_controller.rb
ChrisBAshton: Good spot. I'll [reuse the Content Publisher implementation](https://github.com/alphagov/content-publisher/search?q=forbidden&unscoped_q=forbidden) as best I can.

spec/features/reviewing_step_by_steps_spec.rb
ChrisBAshton: We've spoken IRL - I think this particular method is actually more readable with the argument passing, but agree the scenario itself is now quite long and that some steps should be combined 👍

config/routes.rb
ChrisBAshton: You've looked into this and discovered some interesting Rails behaviour! I've amended my commit.

app/controllers/review_controller.rb
```diff
@@ -1,10 +1,26 @@
...
-  before_action :set_step_by_step_page
+  before_action :require_2i_reviewer_permissions!, only: %i(
+    approve_2i_review
+    claim_2i_review
+    request_change_2i_review
+    show_approve_2i_review_form
+    show_request_change_2i_review_form
+  )
+  before_action :require_user_to_be_the_2i_reviewer!, only: %i(
+    approve_2i_review request_change_2i_review
```
kevindew: looks like you accidentally put 2 on the same line here.

app/controllers/review_controller.rb
```diff
@@ -14,9 +30,9 @@ def approve_2i_review
       reviewer_id: nil,
       status: status,
     )
-      generate_change_note("2i approved")
+      generate_change_note("2i approved", params[:additional_comment])
 
-      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully approved_2i."
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully 2i approved. Please let the author know."
     end
```
kevindew: Did you want this to show the submit_2i_verdict view if the update fails?

app/controllers/review_controller.rb
```diff
@@ -91,4 +111,8 @@ def generate_change_note(status, change_note = nil)
   def set_step_by_step_page
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
   end
+
+  def require_user_to_be_the_2i_reviewer!
+    render "documents/forbidden", status: :forbidden unless current_user.uid == @step_by_step_page.reviewer_id
```
kevindew: I wonder if this would make more sense in the shared views directory. I think `documents` is a Content Publisher concept.

app/views/documents/forbidden.html.erb
```diff
@@ -0,0 +1,5 @@
+<% content_for :title, "Forbidden" %>
+
+<div class="govuk-body" data-gtm="access-limit-forbidden" data-gtm-visibility-tracking="true">
```
kevindew: Do these data attributes have any effect in Collections Publisher?

config/routes.rb
```diff
@@ -4,23 +4,25 @@
   root to: redirect("/step-by-step-pages", status: 302)
 
   resources :step_by_step_pages, path: "step-by-step-pages" do
+    get "approve-2i-review", to: "review#show_approve_2i_review_form"
     post "approve-2i-review", to: "review#approve_2i_review"
     post :check_links
     post "claim-2i-review", to: "review#claim_2i_review"
-    get "internal-change-notes", to: "interal_change_notes"
+    get "internal-change-notes"
```
kevindew: 😄

config/routes.rb
```diff
@@ -4,23 +4,25 @@
...
     get :reorder
     post :reorder
+    get "request-change-2i-review", to: "review#show_request_change_2i_review_form"
     post "request-change-2i-review", to: "review#request_change_2i_review"
     post :revert
     post "revert-to-draft", to: "review#revert_to_draft"
     get :schedule
     post :schedule
-    post "schedule-datetime", to: "schedule_datetime"
+    post "schedule-datetime"
```
kevindew: 👍

spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -55,6 +76,48 @@ def when_I_view_the_step_by_step_page
     visit step_by_step_page_path(@step_by_step_page)
   end
 
+  def and_I_approve_the_step_by_step_with_an_optional_note
+    when_I_click_button "Approve"
+    then_the_page_should_have_title_caption "Approve step by step"
+    and_I_fill_in_an_optional_note
+    and_when_I_click_button "Yes, approve 2i"
```
kevindew: I'd go with making these be the code to actually do the thing rather than descriptive code.

For example:

```
click_on "Approve"
expect(page).to have_text("Approve step by step")
fill_in "additional_comment", with: "Please fix typo before publishing"
click_on "Yes, approve 2i"
```

which ends being more succinct and removes a layer of abstraction.

app/controllers/review_controller.rb
ChrisBAshton: I wondered that too - `shared/` seems to be mostly template partials that can be used to construct pages, rather than containing pages themselves. But agree it would be nice to avoid creating another folder. I'll move it.

app/controllers/review_controller.rb
ChrisBAshton: Good spot. I need an `else` in there.

app/views/documents/forbidden.html.erb
ChrisBAshton: Removed.

spec/support/common_feature_steps.rb
```diff
@@ -28,4 +28,31 @@ def given_I_am_a_2i_reviewer
...
+  def when_I_click_button(button_text)
+    begin
+      click_button button_text, exact: true
+    rescue Capybara::ElementNotFound
+      # some button-like things are actually marked up as link
+      click_link button_text, exact: true
+    end
+  end
+
+  alias_method :and_when_I_click_button, :when_I_click_button
```
kevindew: I don't think this method now needs to be moved, I'd suggest we should aim to keep it out of common steps as the usage of this is more complex than the capybara equivalents.

For example:

```
  ...
  and_i_click_the_remove_button
end

def and_i_click_the_remove_button
  click_button "Remove"
end
```

mental processing for a reader or debugger is just scroll down the file recognise the capybara DSL, then read up on that if needed.

Whereas with this helper approach:

```
  ...
  and_when_I_click_button "Remove"
  ...
```

mental processing is then: scroll through file looking for method fail to find it so either search code base for it or check each helper file for method. Then discover method is alias so then find actual method. Then process the exception based logic handling and go back to your HTML file to work out which one is in use, and finally question whether exact plays a role.

spec/support/common_feature_steps.rb
```diff
@@ -28,4 +28,31 @@ def given_I_am_a_2i_reviewer
...
     ["signin", "GDS Editor", "2i reviewer", "Unreleased feature"]
   end
+
+  def and_I_am_the_step_by_step_author
+    login_as_user @review_requester
+  end
+
+  def and_I_am_the_reviewer
+    login_as_user @reviewer
+  end
```
kevindew: Should these be in step_nav_steps given that is where those variables are defined? I was a bit baffled where the variable came from.

app/controllers/review_controller.rb
```diff
@@ -43,7 +65,9 @@ def request_change_2i_review
     )
       generate_change_note("2i changes requested", params[:requested_change])
 
-      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested."
+      redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Changes to the step by step page were requested. Please let the author know."
+    else
+      show_request_change_2i_review_form
```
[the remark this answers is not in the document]
ChrisBAshton: We do it in a few other places, e.g. in step_by_step_pages_controller. Without it, I don't think we'd keep the user on the same page if it failed to update (which seemed to be [what you were alluding to](https://github.com/alphagov/collections-publisher/pull/780#discussion_r330173119), but I may have misunderstood). What would you suggest instead?

spec/support/common_feature_steps.rb
ChrisBAshton: 👍 I'll move it back to where it was so that it can be revisited in that file later. You make a good point.

app/controllers/review_controller.rb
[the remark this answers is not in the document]
kevindew: Yup - that's what I'd suggest. This is how it was the same in Content Publisher, where we also include a status code to indicate that users input was unprocessable: https://github.com/alphagov/content-publisher/blob/master/app/controllers/documents_controller.rb#L52-L54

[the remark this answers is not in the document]
ChrisBAshton: Thanks, this has been useful!

PR #782, opened 2019-09-30
Remove spaces from start and end of step paths

brucebolt:
Sometimes content designers will copy paths that have a space at the start or the end.

This PR makes two changes:
- Strips spaces from the beginning and end of secondary link paths.  These previously returned a 500 error.
- Strips spaces from the beginning and end of the step by step slug.  This would have given a `Slug is invalid` error.

Trello card: https://trello.com/c/kVsnP1b7

### DIFF
--- a/app/models/step_by_step_page.rb
+++ b/app/models/step_by_step_page.rb
@@ -24,6 +24,7 @@ class StepByStepPage < ApplicationRecord
   validates :status, status_prerequisite: true
   validate :reviewer_is_not_same_as_review_requester
 
+  before_validation :strip_slug_spaces
   before_validation :generate_content_id, on: :create
   before_destroy :discard_notes
 
@@ -175,6 +176,10 @@ private
 
   def generate_content_id
     self.content_id ||= SecureRandom.uuid
+  end
+
+  def strip_slug_spaces
+    self.slug.gsub!(/^\s+|\s+$/, "")
   end
 
   def reviewer_is_not_same_as_review_requester
--- a/spec/controllers/secondary_content_links_controller_spec.rb
+++ b/spec/controllers/secondary_content_links_controller_spec.rb
@@ -22,37 +22,38 @@ RSpec.describe SecondaryContentLinksController do
   end
 
   describe "#create" do
-    it "adds a new secondary content link" do
-      stub_user.permissions << "GDS Editor"
+    context "The content item exists" do
+      before do
+        stub_user.permissions << "GDS Editor"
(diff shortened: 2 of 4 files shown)


spec/controllers/secondary_content_links_controller_spec.rb
```diff
@@ -39,6 +39,22 @@
...
 
+    it "adds a new secondary content link if the base_path has a leading or trailing space" do
+      stub_user.permissions << "GDS Editor"
+
+      allow(Services.publishing_api).to receive(:lookup_content_id).and_return("a-content-id")
+      allow(Services.publishing_api).to receive(:get_content)
+        .with("a-content-id")
+        .and_return(content_item)
+      allow(Services.publishing_api).to receive(:put_content)
+      allow(Services.publishing_api).to receive(:lookup_content_ids).and_return([])
```
ChrisBAshton: Could this block be moved up to a context that is shared by the `it "adds a new secondary content link" do` test?

```ruby
describe "#create" do
  context "The content item exists" do
    before do
      allow(Services.publishing_api).to receive(:lookup_content_id).and_return("a-content-id")
      allow(Services.publishing_api).to receive(:get_content)
        .with("a-content-id")
        .and_return(content_item)
      allow(Services.publishing_api).to receive(:put_content)
      allow(Services.publishing_api).to receive(:lookup_content_ids).and_return([])
    end

    it "adds a new secondary content link" do
      # etc
    end

    it "adds a new secondary content link if the base_path has a leading or trailing space" do
      # etc
    end
```

I think some of the tests below could benefit from this too.

app/models/step_by_step_page.rb
```diff
@@ -177,6 +178,10 @@ def generate_content_id
     self.content_id ||= SecureRandom.uuid
   end
 
+  def strip_slug_spaces
+    self.slug.gsub!(/^\s+|\s+$/, "")
```
ChrisBAshton: As this is the same logic used by the secondary links, could we move this method to a central place to be used by both? Perhaps in `ApplicationHelper`?

brucebolt: I'm not too keen on doing this, since whilst the regex is the same, what's actually happening is different.  One place is updating a value on the model, whilst the other is modifying a local variable.  It means I'd still have a `strip_slug_spaces` method in the model, which would then be referencing another method elsewhere, which feels like extra complexity.

spec/controllers/secondary_content_links_controller_spec.rb
ChrisBAshton: The `it "accepts a full url as the base_path"` can be in this context too 👍

PR #785, opened 2019-09-30
Move existing step by step guidance to a YAML file

AgaDufrat:
### User story

**As a** content designer
**I want** to be able to update contextual guidance in human readable files
**So that** I can do this without knowing Ruby
 
---

-  Decided to store the guidance in a single file as it will be easier for content designers to find and amend without knowing the structure of the folders and files. _This can be revised when we add more text._
- YAML file structure follows the folder structure of the views so it can be easily followed and the text can be reused in the future. It's the preferred way and it also allows to use a 'lazy' lookup without providing full path to the key when referencing it in the `t` method eg. `t(".slug_guidance")`

_Note:_ Not moved as it contains Markdown:
- Guidance on markdown `shared/steps/markdown_help.html.erb` 
- Guidance on secondary links `secondary_content_links/index.html.erb`

[Trello card](https://trello.com/c/lBPFRJQY/100-create-yaml-file-to-store-the-guidance-and-add-current-guidance-to-that-yaml-file)

### DIFF
--- a/app/views/secondary_content_links/index.html.erb
+++ b/app/views/secondary_content_links/index.html.erb
@@ -50,13 +50,9 @@
 
             <%= render "govuk_publishing_components/components/input", {
               label: {
-                text: "Base Path"
+                text: t("secondary_content_links.index.base_path.label")
               },
-              hint: capture do %>
-                For example:
-                <a href="https://www.gov.uk/pay-vat" target="_blank" class="govuk-link">https://www.gov.uk/pay-vat</a>
-                or <a href="https://www.gov.uk/bank-holidays" class="govuk-link" target="_blank">/bank-holidays</a>
-              <% end,
+              hint: t("secondary_content_links.index.base_path.hint"),
               name: "base_path"
             } %>
           </div>
--- a/app/views/step_by_step_pages/_change_notes.html.erb
+++ b/app/views/step_by_step_pages/_change_notes.html.erb
@@ -1,18 +1,18 @@
 <% textarea = capture do %>
     <%= render "govuk_publishing_components/components/textarea", {
       label: {
-        text: "Public change note",
+        text: t("step_by_step_pages.change_notes.label"),
       },
       name: "change_note",
       rows: 4,
-      hint: "Describe the change for users. This change note will be emailed to subscribers."
+      hint: t("step_by_step_pages.change_notes.hint"),
     } %>
 <% end %>
 
 <div class="govuk-grid-row">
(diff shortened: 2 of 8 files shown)


app/views/secondary_content_links/index.html.erb
```diff
@@ -23,13 +23,13 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-  <h2 class="govuk-heading-m">What is secondary content?</h2>
+  <h2 class="govuk-heading-m"><%= t("secondary_content_links.index.header") %></h2>
```
alex-ju: This whole section (guidance on secondary links) could be converted to multi-line govspeak too giving a bit more flexibility in how the content is structured.

app/views/secondary_content_links/index.html.erb
```diff
@@ -52,11 +52,7 @@
               label: {
                 text: "Base Path"
               },
-              hint: capture do %>
-                For example:
-                <a href="https://www.gov.uk/pay-vat" target="_blank" class="govuk-link">https://www.gov.uk/pay-vat</a>
-                or <a href="https://www.gov.uk/bank-holidays" class="govuk-link" target="_blank">/bank-holidays</a>
-              <% end,
+              hint: t("secondary_content_links.index.example"),
```
alex-ju: Could we add the labels for each input to the yaml file too and I suggest we structure them based on their name instead of what their content so that, for example, if we decide to swap the hint for this field to say something else instead of showing an example we won't need to rename the yaml declaration.

```
secondary_content_links:
    index:
        base_path:
            label: Base Path
            hint: For example: https://www.gov.uk/pay-vat or /bank-holidays
```

app/views/step_by_step_pages/_change_notes.html.erb
```diff
@@ -1,18 +1,18 @@
 <% textarea = capture do %>
     <%= render "govuk_publishing_components/components/textarea", {
       label: {
-        text: "Public change note",
+        text: t("step_by_step_pages.change_notes.guidance_title"),
```
alex-ju: As mentioned above, I suggest we name this `label` and the one below `hint`

app/views/step_by_step_pages/_form.html.erb
```diff
@@ -29,7 +29,7 @@
           label: {
             text: "Slug"
```
alex-ju: As mentioned above, could we add the label to the yaml too?

PR #791, opened 2019-10-02
Parse markdown in in-app guidance

brucebolt:
Content designers may wish to update in-app guidance by raising PRs on YAML files.  At the moment they would have to include HTML for any formatting.  The changes in this PR allow markdown to be used in guidance, and updates the existing guidance to use valid markdown.

Trello card: https://trello.com/c/zNqQjbMo

### DIFF
--- a/app/views/secondary_content_links/index.html.erb
+++ b/app/views/secondary_content_links/index.html.erb
@@ -23,15 +23,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-  <h2 class="govuk-heading-m">What is secondary content?</h2>
-    <div class="govuk-body">
-      <p class="govuk-body-s">Secondary content is content which is not linked to in the step by step, but has the step by step sidebar showing on it.</p>
-      <p class="govuk-body-s">We use it for content which:</p>
-      <ul class="govuk-list govuk-list--bullet">
-        <li class="govuk-body-s">helps you to complete a task in the journey, but is not the primary piece of content you need to visit to complete that task</li>
-        <li class="govuk-body-s">is optional and not useful enough to be included in the step by step</li>
-      </ul>
-    </div>
+    <%= render_markdown t("secondary_content_links.guidance_markdown") %>
   </div>
 </div>
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,4 +20,14 @@ module ApplicationHelper
   def published_url(slug)
     "#{Plek.new.website_root}/#{slug}"
   end
+
+  def markdown_to_html(markdown)
+    raw(Kramdown::Document.new(markdown).to_html)
+  end
+
+  def render_markdown(content)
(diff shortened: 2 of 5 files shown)


kevindew:
I think it's fine to downgrade the gems 👍 

One thing that I've been thinking about recently though is whether we want to be giving content designers markdown or govspeak, this came up recently where [govuk_publishing_components couldn't have a dependency on govspeak](https://github.com/alphagov/govuk_publishing_components/pull/827) (as govspeak has a dependency on govuk_publishing_components).

My theory is that when it's an app for guidance, we don't actually use any of the govspeak additions and users are fine with markdown, which can be simpler and faster to render. It doesn't matter whether we render markdown or govspeak for the HTML that goes into the govspeak component.

This all leads into other thoughts of mine I've had where I think we're too lenient on the input we allow users to input as govspeak (you can enter all sorts of HTML and stuff) whereas we should be very flexible for input that's checked into a repo. So actually splitting them where govspeak is restricted more but app embedded markdown is more freeform makes sense.

This is all a very long way of saying: why don't we just not install govspeak and just allow content designers to input markdown and see if that causes any problems?

app/views/secondary_content_links/index.html.erb
```diff
@@ -22,16 +22,8 @@
...
-    <div class="govuk-body">
-      <p class="govuk-body-s">Secondary content is content which is not linked to in the step by step, but has the step by step sidebar showing on it.</p>
-      <p class="govuk-body-s">We use it for content which:</p>
-      <ul class="govuk-list govuk-list--bullet">
-        <li class="govuk-body-s">helps you to complete a task in the journey, but is not the primary piece of content you need to visit to complete that task</li>
-        <li class="govuk-body-s">is optional and not useful enough to be included in the step by step</li>
-      </ul>
-    </div>
+  <div class="govuk-grid-column-full govuk-body-s">
+    <%= govspeak_to_html t("secondary_content_links.guidance") %>
```
kevindew: This should be inside a govspeak component by the way to render the markdown correctly.

brucebolt:
That makes complete sense 👍.  I'll update the PR to only parse markdown.

app/views/secondary_content_links/index.html.erb
brucebolt: I don't think this will be needed now we're not using Govspeak.  All the markdown should have been parsed into valid HTML.

kevindew: We'll still need it to add the styles to the HTML as our naked HTML is pretty much styleless with govuk-frontend.

brucebolt: The relevant classes are already there, since we had plain HTML in the places where this text was previously appearing.

brucebolt: After chatting in person earlier, I've added a `render_markdown` method which wraps the HTML in a govspeak component.

I've also reduced the number of fields in which we allow markdown, since input labels shouldn't need to contain markdown.

PR #797, opened 2019-10-04
[DO NOT MERGE] Only 2i approved step by steps can be published or scheduled

brucebolt:
Requires a step by step to have been 2i approved before it can be scheduled or published.

This protects the relevant routes to ensure the approved status is set, and updates the `can_be_published?` logic so the relevant actions are only shown after approval.

Tests have been updated to match this change.  Feature tests have been changed to assume the SBS have been 2i approved in all cases, except where the test is specifically testing the 2i workflow.

Not to be merged until the 2i workflow is to be released and "Unreleased feature" flag is removed.

Trello card: https://trello.com/c/8o8PMB6n/88-only-approved-step-by-steps-can-be-published-or-scheduled-after-2i-is-released

### DIFF
--- a/app/controllers/step_by_step_pages_controller.rb
+++ b/app/controllers/step_by_step_pages_controller.rb
@@ -5,6 +5,7 @@ class StepByStepPagesController < ApplicationController
 
   before_action :require_gds_editor_permissions!
   before_action :set_step_by_step_page, only: %i[show edit update destroy]
+  before_action :require_to_be_2i_approved, only: %i[publish schedule schedule_datetime]
 
   def index
     @step_by_step_pages = StepByStepFilter::Results.new(filter_params).call
@@ -245,5 +246,12 @@ private
 
   def set_current_page_as_step_by_step
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
+  end
+
+  def require_to_be_2i_approved
+    set_current_page_as_step_by_step
+    unless @step_by_step_page.status.approved_2i?
+      redirect_to @step_by_step_page, notice: "Step by step must be 2i approved before you can #{action_name} this step by step."
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/step_by_step_pages_controller.rb
```diff
@@ -246,4 +247,11 @@ def filter_params
   def set_current_page_as_step_by_step
     @step_by_step_page = StepByStepPage.find(params[:step_by_step_page_id])
   end
+
+  def require_to_be_2i_approved
+    set_current_page_as_step_by_step
+    unless @step_by_step_page.status.approved_2i?
+      redirect_to @step_by_step_page, notice: "Step by step must be 2i approved before you can #{action_name} this step by step."
```
ChrisBAshton: Nice use of `action_name` 👍

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -49,6 +51,27 @@
   end
 
   describe "#publish" do
+    context "2i approval" do
+      it "cannot be published when step by step is not approved" do
```
ChrisBAshton: Current wording could be misinterpreted as "cannot be published if changes requested" (the antithesis to "approved"), but what we really mean is "cannot be published if in the wrong state" (`draft`, `submitted_for_2i`, etc).

Might I suggest...?

```ruby
      it "cannot be published when step by step is not in `approved_2i` state" do
```

spec/controllers/step_by_step_pages_controller_spec.rb
```diff
@@ -204,6 +227,21 @@ def schedule_for_future
...
+        step_by_step_page.mark_draft_updated
+        schedule_for_future
+
+        expect(step_by_step_page.status).not_to be_scheduled
+      end
+
+      it "can be scheduled when step by step is approved" do
+        schedule_for_future
+
+        expect(step_by_step_page.status).to be_scheduled
```
ChrisBAshton: Love these Railsy predicate matchers too 👍

brucebolt:
Thanks Chris, I have updated the test name 👍

PR #801, opened 2019-10-04
Configurable status names and updated ordering

kevindew:
This updates the ordering of statuses to change the filtering dropdown. This also moves the presentation of status names to be a translation file concern. It then renames the "In review" status to be "In 2i review". More info in commits.

Before:

Screen Shot 2019-10-04 at 13 30 41

After:

Screen Shot 2019-10-04 at 13 33 33


spec/services/step_by_step_filter/options_spec.rb
```diff
@@ -2,69 +2,17 @@
...
-        },
-        selected: false,
-      }
-
-      expect(subject).to include(status_option)
-    end
+    expect(described_class.statuses)
+      .to include(text: "All",
+                  data_attributes: { show: "all" },
+                  selected: true)
```
ChrisBAshton: This test no longer does quite what it says on line 4 - it only checks if an `All` option exists.

I'd suggest checking that `described_class.statuses.count` is equal to  something like `step_by_step_page.STATUSES.count + 1`

spec/services/step_by_step_filter/options_spec.rb
```diff
@@ -2,69 +2,17 @@
...
-        selected: false,
-      }
-
-      expect(subject).to include(status_option)
-    end
+    expect(described_class.statuses("scheduled"))
+      .to include(text: "Scheduled",
+                  value: "scheduled",
+                  data_attributes: { show: "scheduled" },
+                  selected: true)
```
ChrisBAshton: Same here - otherwise test will pass even if we only have "Scheduled" in the filter list.

config/locales/en.yml
```diff
@@ -3,7 +3,7 @@ en:
     statuses:
       draft: Draft
       submitted_for_2i: Submitted for 2i
-      in_review: In review
+      in_review: In 2i review
```
ChrisBAshton: I'd tweak the commit message (`shouldn't be really clear on the value of that before proceeding` should be _should_, I think?)

app/controllers/review_controller.rb
```diff
@@ -79,9 +79,9 @@ def submit_for_2i
         review_requester_id: current_user.uid,
         status: status,
       )
-        generate_change_note(status, params[:additional_comments])
+        generate_change_note(I18n.t!("step_by_step_page.statuses.#{status}"), params[:additional_comments])
 
-        redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully #{status.humanize.downcase}."
+        redirect_to step_by_step_page_path(@step_by_step_page.id), notice: "Step by step page was successfully submitted for 2i review."
```
brucebolt: Probably out of scope for this PR, but these success notices feel like they should be held as translations too.  Maybe something we can do if there's time at the end of the mission.

config/locales/en.yml
kevindew: Thanks 👍 good catch.

spec/services/step_by_step_filter/options_spec.rb
kevindew: Ah ha did I go too far? I've added an extra test in.

app/controllers/review_controller.rb
kevindew: Yup would be a good step - it's always rather painful when you're transitioning and there's not clear logic of what is in or isn't.

>  if there's time at the end of the mission
😆

spec/services/step_by_step_filter/options_spec.rb
```diff
@@ -1,70 +1,25 @@
...
-    other_options.each do |status|
-      status_option = {
-        text: status.humanize,
-        value: status,
-        data_attributes: {
-          show: status,
-        },
-        selected: false,
-      }
+    it "defaults to all option selected" do
```
ChrisBAshton: Suggestion to avoid ambiguity:

```suggestion
    it "defaults to 'All' option selected" do
```

(Otherwise kind of reads like "all options should be selected by default")

spec/services/step_by_step_filter/options_spec.rb
```diff
@@ -1,70 +1,25 @@
...
+                    selected: true)
+    end
 
-      expect(subject).to include(status_option)
+    it "returns filter options with the selected status" do
+      expect(described_class.statuses("scheduled"))
+        .to include(text: "Scheduled",
+                    value: "scheduled",
+                    data_attributes: { show: "scheduled" },
+                    selected: true)
```
ChrisBAshton: In a similar vein, currently all options might have `selected: true` and would currently pass the test suite. Is it worth validating something like:

```ruby
unselected_options = described_class.statuses.reject(&:selected)
expect(unselected_options.count).to eq(described_class.statuses.count - 1)
```

kevindew: Sure, I can see how that's missing. This all seems to be going progressively over-the-top though given this is essentially a view concern and the logic is rather trivial. The previous tests essentially re-implemented most of the method.

Given we already have this covered by a feature test I'm going to just move this logic into a view and kill the class.

kevindew: Ok I've removed it now.

spec/services/step_by_step_filter/options_spec.rb
kevindew: Makes sense, but test no longer exists.

PR #806, opened 2019-10-07
Ignore GOV.UK prefix to show absolute links on Sidebar Settings

ChrisBAshton:
This will allow absolute URL links to be picked up by StepLinksForRules
for inclusion in the Sidebar Settings page.

This PR also refactors `StepContentParser` a little (see first commit).

Trello: https://trello.com/c/4JeWifGe/86-strip-govuk-from-link-markdown-and-show-them-on-sidebar-settings

### DIFF
--- a/app/services/step_content_parser.rb
+++ b/app/services/step_content_parser.rb
@@ -12,6 +12,8 @@ class StepContentParser
   # matches [Link text](url)
   LINK_REGEX = /\[[^\[\]]+\]\(([^)]+)/.freeze
 
+  GOVUK_DOMAIN_REGEX = /^(https?:\/\/)?(www\.)?gov\.uk\//.freeze
+
   def parse(step_text)
     sections = step_text.rstrip.delete("\r").split("\n\n").map do |section|
       section.lines.map(&:chomp)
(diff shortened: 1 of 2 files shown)


app/services/step_content_parser.rb
```diff
@@ -58,7 +58,9 @@ def safely_parse_path(path)
   end
 
   def relative_paths(content)
-    all_links_in_content(content).select { |href| href.match(/^\/[a-z0-9]+.*/i) }
+    all_links_in_content(content)
+      .map { |url| strip_govuk_prefix(url) }
```
brucebolt: We have a lot of regular expressions defined as constants (e.g. `LINK_REGEX`).  It might be a good idea to define `GOVUK_PREFIX_REGEX` as a constant, then use it here?

ChrisBAshton: Thanks for the suggestion. Now done 👍 

Was going to do the same for the line below it, but got into a muddle (called it `ABSOLUTE_URL_REGEX`, which is effectively what we're trying to do here, but the regex isn't _really_ an absolute URL regex, so was a bit of a rabbit hole. Would end up with something like `NOT_A_SLUG_REGEX`, which just doesn't read easily!)

PR #807, opened 2019-10-07
Do not revert step by step to draft under certain circumstances

brucebolt:
Sometimes a reviewer or review requester will want to make minor changes without getting stuck in a 2i review loop.  This change prevents the status from changing to 'draft' when a step by step that is currently submitted for 2i or in 2i review is edited.  If a step by step has been approved, published or scheduled, then the status is returned to 'draft' and another 2i review is required.

Trello card: https://trello.com/c/4FN9RaqT

### DIFF
--- a/spec/services/step_nav_publisher_spec.rb
+++ b/spec/services/step_nav_publisher_spec.rb
@@ -5,6 +5,8 @@ RSpec.describe StepNavPublisher do
   include GdsApi::TestHelpers::PublishingApi
 
   let(:step_nav) { create(:step_by_step_page_with_steps) }
+  let(:review_requester_user) { create(:user) }
+  let(:reviewer_user) { create(:user) }
 
   before do
     stub_any_publishing_api_call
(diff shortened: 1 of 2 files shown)


spec/services/step_nav_publisher_spec.rb
```diff
@@ -29,6 +31,40 @@
         expect(Services.publishing_api).to have_received(:put_content).with(content_id, payload)
       end
     end
+
+    context "marks draft updated" do
```
ChrisBAshton: Kevin has picked up on this in one of my PRs, but our use of `context` is quite confusing. I'd expect this to be a `describe "whether a step by step should be reverted to draft when edited"`, _or_ a `context "there is a step by step page with steps"`.

But it looks like `context` is used in this way throughout the file, so probably one to tackle in a future cleanup PR 🤔

spec/services/step_nav_publisher_spec.rb
```diff
@@ -29,6 +31,40 @@
...
+      let(:step_nav) { create(:step_by_step_page_with_steps) }
+
+      it "does not mark submitted for 2i review as draft" do
```
ChrisBAshton: I'd rephrase all of these to be less dev-speak:

`it "does not revert to draft if a step by step in `submitted_for_2i` state is edited"`

spec/services/step_nav_publisher_spec.rb
```diff
@@ -29,6 +31,40 @@
...
+        step_nav.update_attributes(:status => "submitted_for_2i", :review_requester_id => review_requester_user.uid, :reviewer_id => reviewer_user.uid)
```
ChrisBAshton: Any reason why we're not using `status: "submitted_for_2i"` syntax rather than the rocket?

spec/services/step_nav_publisher_spec.rb
```diff
@@ -29,6 +31,40 @@
...
+        expect(step_nav.status).to eq "submitted_for_2i"
+      end
+
+      it "does not mark in 2i review as draft" do
+        step_nav.update_attributes(:status => "in_review", :review_requester_id => review_requester_user.uid, :reviewer_id => reviewer_user.uid)
+        StepNavPublisher.update(step_nav)
+        expect(step_nav.status).to eq "in_review"
+      end
+
+      it "marks 2i approved as draft" do
```
ChrisBAshton: Again, think this would be clearer as `it "should revert to draft if a step by step in 'approved_2i' state is edited"`.

spec/services/step_nav_publisher_spec.rb
brucebolt: No particular reason.  The usage appears inconsistent throughout the Collections Publisher tests (e.g. `spec/controllers/review_controller_spec.rb` uses rockets for `update_attributes` whereas `spec/factories.rb` uses the colon).

ChrisBAshton: Hmm and there's a mixture in this file too. I think generally we prefer the colon syntax unless there's a reason to use the rocket syntax (such as the name of the symbol being incompatible with colon syntax).
I'd be tempted to use the colon syntax here to try to nudge the codebase in the right direction (a cleanup PR later will probably still be needed on the repo as a whole!). It's also a bit less overhead to read.

brucebolt: I've updated these lines to use the colon syntax

PR #810, opened 2019-10-07
Remove compatiblity with govuk-template

alex-ju:
This application doesn't rely on `govuk-template` so we're removing this flags to use the new colour scheme. This aligns the styles for Collections Publisher with Content Publisher and Signon.

### DIFF
--- a/app/assets/stylesheets/objects/_side.scss
+++ b/app/assets/stylesheets/objects/_side.scss
@@ -3,7 +3,7 @@
   @include govuk-clearfix;
 
   padding: govuk-spacing(3);
-  background: govuk-colour("light-grey", $legacy: "grey-4");
+  background: govuk-colour("light-grey");
 }
 
 .app-side__actions {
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/objects/_side.scss
```diff
@@ -41,7 +41,7 @@
 
     &:hover,
     &:active {
-      color: govuk-colour("bright-red");
+      color: govuk-colour("red");
```
ChrisBAshton: All the other changes in this PR are to do with removing the legacy stuff, but this change looks like it's just fixing a color that shouldn't have been used in the first place. Do you mind splitting this one out into a separate commit?

alex-ju: This should've been `govuk-colour("red", $legacy: "bright-red")`, but because the legacy colour wasn't declared instead of only removing the legacy declaration it looks like a colour swap. More details about [how the update path for colours on govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#update-colours).

ChrisBAshton: Ah, makes sense 👍

PR #813, opened 2019-10-08
Test actions thoroughly

ChrisBAshton:
Our step by steps can now be in any of a number of different 'states', and accessed by a number of different 'actors', all of which affects the actions available on the step by step.

We've started a dedicated feature file to test that the correct actions apply at the correct times, but this PR fills in some of the gaps:

- [x] Tests that there is the right quantity of Primary, Secondary and Tertiary actions
  (won't silently fail if another action creeps in, for example)
- [x] Test actions for when step by step is 2i approved
- [x] Test actions for when step by step is scheduled
- [x] Test actions for when step by step is published
- [x] Test actions for when step by step is edited after publish

Trello: https://trello.com/c/Qle5c1QQ/43-write-tests-display-the-appropriate-buttons-depending-on-the-state-of-the-sbs-and-who-is-logged-in

### DIFF
--- a/spec/features/step_by_step_actions_spec.rb
+++ b/spec/features/step_by_step_actions_spec.rb
@@ -67,12 +67,12 @@ RSpec.feature "Contextual action buttons for step by step pages" do
   end
 
   def then_the_primary_action_should_be(action_text)
-    custom_error = "Couldn't find '#{action_text}' as a primary action in: \n #{action_html}"
-    expect(page).to have_css(".app-side__actions .gem-c-button:not(.gem-c-button--secondary)", text: action_text), custom_error
+    expect(page).to have_css(primary_action_selector, count: 1)
+    expect(page).to have_css(primary_action_selector, text: action_text), "Couldn't find '#{action_text}' as a primary action in: \n #{action_html}"
   end
 
   def then_there_should_be_no_primary_action
-    expect(page).not_to have_css(".app-side__actions .gem-c-button:not(.gem-c-button--secondary)")
+    expect(page).not_to have_css(primary_action_selector)
   end
 
   def and_the_secondary_action_should_be(action_text)
@@ -85,6 +85,10 @@ RSpec.feature "Contextual action buttons for step by step pages" do
       custom_error = "Couldn't find '#{action_text}' as a tertiary action in: \n #{action_html}"
       expect(page).to have_css(".app-side__actions .govuk-link", text: action_text), custom_error
     end
+  end
+
+  def primary_action_selector
+    ".app-side__actions .gem-c-button:not(.gem-c-button--secondary)"
   end
 
   def action_html
--- a/spec/features/step_by_step_actions_spec.rb
+++ b/spec/features/step_by_step_actions_spec.rb
@@ -16,7 +16,7 @@ RSpec.feature "Contextual action buttons for step by step pages" do
       given_there_is_a_step_by_step_page_with_a_link_report
       when_I_visit_the_step_by_step_page
       then_the_primary_action_should_be "Submit for 2i review"
-      and_the_secondary_action_should_be "Preview"
+      and_there_should_be_secondary_actions_to %w(Preview)
       and_there_should_be_tertiary_actions_to %w(Delete)
     end
   end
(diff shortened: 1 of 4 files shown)


spec/features/step_by_step_actions_spec.rb
```diff
@@ -67,12 +67,12 @@
   end
 
   def then_the_primary_action_should_be(action_text)
-    custom_error = "Couldn't find '#{action_text}' as a primary action in: \n #{action_html}"
-    expect(page).to have_css(".app-side__actions .gem-c-button:not(.gem-c-button--secondary)", text: action_text), custom_error
+    expect(page).to have_css(primary_action_selector, count: 1)
```
brucebolt: It might be useful to have custom error text here, since the second expectation has this and it may not be clear what is required without reading the commit message

spec/features/step_by_step_actions_spec.rb
```diff
@@ -76,8 +75,19 @@ def then_there_should_be_no_primary_action
   end
 
   def and_the_secondary_action_should_be(action_text)
-    custom_error = "Couldn't find '#{action_text}' as a secondary action in: \n #{action_html}"
-    expect(page).to have_css(".app-side__actions .gem-c-button--secondary", text: action_text), custom_error
+    expect(page).to have_css(secondary_action_selector, count: 1)
+    assert_secondary_action_exists(action_text)
+  end
```
brucebolt: It feels like there is duplication between the `and_the_secondary_action_should_be` and `and_the_secondary_actions_should_be` methods since they both test the same thing (just one takes a string, the other an array).  Maybe there could be one method which accepts an array like you've done for `and_there_should_be_tertiary_actions_to`, then have single item arrays when this method is invoked?  That would also mean `assert_secondary_action_exists` wouldn't be needed.

spec/features/step_by_step_actions_spec.rb
```diff
@@ -85,6 +85,16 @@
...
+  context "Step by step has been published" do
+    scenario "show the relevant actions to the step by step author" do
+      given_there_is_a_published_step_by_step_page
+      when_I_visit_the_step_by_step_page
+      then_there_should_be_no_primary_action
+      and_there_should_be_no_secondary_action
+      and_there_should_be_tertiary_actions_to ["Unpublish", "View on GOV.UK"]
+    end
+  end
+
```
brucebolt: It looks like this is also being tested in https://github.com/alphagov/collections-publisher/blob/master/spec/features/managing_step_by_step_pages_spec.rb#L369-L379.

Is there any way the duplication could be removed?

ChrisBAshton: Thanks, had meant to do that 👍 done.

spec/features/step_by_step_actions_spec.rb
ChrisBAshton: Have now made the approach consistent with how we do tertiary actions.

spec/features/step_by_step_actions_spec.rb
ChrisBAshton: The default error message is better, I think - it says something along the lines of:

> "Expected to find '.app-side__actions .gem-c-button:not(.gem-c-button--secondary)' 1 time. Found 3 times".

If I wrote a custom error text I'd be able to inject `action_html`, which is nice, but I'd also have to grab the actual vs expected count, which starts to get convoluted IMO.

kevindew:
Sorry to be a pain and don't let my comments stop this being merged as I can see this is really close and builds on existing work. More just adding this comment as food for thought (hopefully).

I think we're at risk here of unit testing the view through feature tests here. Which risks creating precedence/confusion in the purpose of feature tests - using classic testing paradigms this feels like inverting the [test pyramid concept](https://martinfowler.com/bliki/TestPyramid.html) towards the [test ice cream cone](http://saeedgatson.com/the-software-testing-ice-cream-cone/).

Where I think these stray from being feature tests is that we just check that a button is there, that itself is not a feature. It could be considered a feature that an approved step by step can be published for example, so there could be arguments that rather than testing presence of buttons these tests are scenarios of a feature - but it could become superfluous given they'll take you into existing feature journeys (I imagine the maths quickly adds up).

So I'm kind of wondering have we hit a point where we have a view that is sufficiently complex we should write a unit test for it (which would probably be more concise and much faster to run), if that's a smell that we have logic that should be tested elsewhere or if maybe we're double testing as a lot of these aspects are part of existing features?

Again sorry to be a pain, and interested in your thoughts - maybe I'm misinterpreting our approach.

ChrisBAshton:
@kevindew agreed - sounds like the view logic should ideally be in a presenter, and this test suite would be moved to a presenter unit test suite. 👍 Perhaps in a future PR?

kevindew:
> @kevindew agreed - sounds like the view logic should ideally be in a presenter, and this test suite would be moved to a presenter unit test suite. 👍 Perhaps in a future PR?

Yup sounds good 👍

PR #815, opened 2019-10-08
Create auth_bypass_tokens that expire and have a draft assets flag

ChrisBAshton:
Stores extra information in the JWT token created for previewing step by step pages, including:

* When it was created
* The content ID of the step by step
* The user ID of the user whose edit created the token
* A `draft_asset_manager_access` flag (set to `true`). We will later do work to detect this flag and allow access to draft assets.

In addition, we set a 1 month expiry time on the JWT token, after which the token becomes invalid. This is desired as we'll be making changes to asset manager to allow access to all assets on the draft stack (which have not been marked as "access limited") when this token is present, and it is unnecessary to grant unrestricted access beyond a reasonable time period required for fact checking.

https://trello.com/c/rjBygcv0/131-create-collections-publisher-authbypasstokens-that-expire-and-have-a-draft-assets-flag


app/views/step_by_step_pages/show/_actions.erb
```diff
@@ -1,6 +1,11 @@
 <% preview_link = render "govuk_publishing_components/components/button", {
   text: "Preview",
-  href: preview_url(@step_by_step_page.slug, auth_bypass_id: @step_by_step_page.auth_bypass_id),
+  href: preview_url(
+    @step_by_step_page.slug,
+    auth_bypass_id: @step_by_step_page.auth_bypass_id,
+    step_by_step_id: @step_by_step_page.content_id,
+    user_id: current_user.uid,
+  ),
```
kevindew: I'd suggest that we have a particular helper function for `step_by_step` preview urls to make this more concise, since we're passing in a few arguments from a step by step page - do we have a view helper for step_by_step pages?

Could be something like:

```
def step_by_step_page_preview_url(step_by_step_page, user = nil)
  user ||= current_user
  token_payload = {
    "sub" => step_by_step_page.auth_bypass_id,
    "iss" => user.uid,
    "iat" => Time.zone.now.to_i,
    "exp" => 1.month.from_now.to_i,
    "draft_asset_manager_access" => true,
    "content_id" => step_by_step_page.content_id,
  }
  token = JWT.encode(token_payload, ENV["JWT_SECRET"], "HS256")
  
  "#{Plek.new.external_url_for('draft-origin')}/#{step_by_step_page.slug}?token=#{token}"
end
```

I'm not sure we need the whole cascade of the existing methods - do other things use them? I say this from a slight prejudice that the JwtHelper class is not a real class so smells a bit too me 🙂

spec/lib/jwt_helper_spec.rb
```diff
@@ -5,10 +5,13 @@
...
-      expected_url = "#{url}?token=#{expected_token}"
-
-      expect(described_class.access_limited_preview_url(url, 123)).to eq(expected_url)
+      # The generated token value now changes every second due to the time-related `iat` and `exp` fields,
+      # so we can only consistently reproduce the token value by setting the system time.
+      Timecop.travel Time.zone.local(2020, 1, 1, 10, 0, 0) do
+        expected_token = "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEyMywiaXNzIjo5LCJpYXQiOjE1Nzc4NzI4MDAsImV4cCI6MTU4MDU1MTIwMCwiZHJhZnRfYXNzZXRfbWFuYWdlcl9hY2Nlc3MiOnRydWUsImNvbnRlbnRfaWQiOjd9.QDj7svewjQcLhikv_o37cy7_dBPH-ME2e-0ec5IcIn4"
+        expected_url = "#{url}?token=#{expected_token}"
+        expect(described_class.access_limited_preview_url(url, 123, 7, 9)).to eq(expected_url)
+      end
```
kevindew: Talking of the JwtHelper smell it feels a bit strange that we'd need essentially the same test here and in application helper.

app/views/step_by_step_pages/show/_actions.erb
ChrisBAshton: Agreed 👍 thanks for confirming I'm on the right track!

spec/lib/jwt_helper_spec.rb
ChrisBAshton: Have removed JwtHelper in my latest version of the branch.

app/helpers/application_helper.rb
```diff
@@ -12,10 +12,8 @@ def authenticated_preview_url(slug)
   end
 
   def anonymous_preview_url(step_by_step_page)
-    JwtHelper.access_limited_preview_url(
-      authenticated_preview_url(step_by_step_page.slug),
-      step_by_step_page.auth_bypass_id
-    )
+    token = JWT.encode({ "sub" => step_by_step_page.auth_bypass_id}, ENV["JWT_AUTH_SECRET"], "HS256")
```
ChrisBAshton: Note to self: fix up linting on this line.

app/helpers/application_helper.rb
```diff
@@ -11,8 +11,12 @@ def authenticated_preview_url(slug)
     "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
   end
 
-  def anonymous_preview_url(step_by_step_page)
-    token = JWT.encode({ "sub" => step_by_step_page.auth_bypass_id}, ENV["JWT_AUTH_SECRET"], "HS256")
+  def anonymous_preview_url(step_by_step_page, user)
+    payload = {
+      "sub" => step_by_step_page.auth_bypass_id,
+      "iss" => user.uid,
```
ChrisBAshton: Note to self: there's trailing whitespace on this line that needs removing

app/helpers/application_helper.rb
```diff
@@ -15,6 +15,7 @@ def anonymous_preview_url(step_by_step_page, user)
     payload = {
       "sub" => step_by_step_page.auth_bypass_id,
       "iss" => user.uid,
+      "draft_asset_manager_access" => true,
```
ChrisBAshton: Note to self: can probably merge this and previous commit.

app/helpers/application_helper.rb
```diff
@@ -7,14 +7,21 @@ def content_tagger_url
...
   end
 
-  def preview_url(slug, auth_bypass_id: nil)
-    url = "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
-
-    if auth_bypass_id.present?
-      url = JwtHelper.access_limited_preview_url(url, auth_bypass_id)
-    end
+  def authenticated_preview_url(slug)
+    "#{Plek.new.external_url_for('draft-origin')}/#{slug.sub(/^\//, '')}"
```
kevindew: Do you know why we strip this first slash off the slug?

Is it new that we're having authenticated and anonymous preview urls? Most of our apps seem to consistently just give out the link with a token avoiding the need for two?

app/helpers/application_helper.rb
```diff
@@ -7,14 +7,21 @@ def content_tagger_url
...
+  end
 
-    url
+  def anonymous_preview_url(step_by_step_page, user)
```
kevindew: As this is now coupled to a step by step should it have step_by_step in the function name?

app/helpers/application_helper.rb
```diff
@@ -7,14 +7,21 @@ def content_tagger_url
...
+    payload = {
+      "sub" => step_by_step_page.auth_bypass_id,
+      "iss" => user.uid,
+      "iat" => Time.zone.now.to_i,
+      "exp" => 1.month.from_now.to_i,
+      "draft_asset_manager_access" => true,
+      "content_id" => step_by_step_page.content_id,
+    }
+    token = JWT.encode(payload, ENV["JWT_AUTH_SECRET"], "HS256")
```
kevindew: 👍

app/helpers/application_helper.rb
kevindew: Actually reading more I now understand. I see we're also using this function to link to other GOV.UK pages. That also explains the slash removal.

I think it might be clearer if this function was just `draft_govuk_url` and accepted a path rather than a slug then could be just `Plek.find("draft-origin", external: true) + path`

Arguably you could probably just have a helper for `draft_govuk` and then just append to paths in view code.

I do wonder if we might end up adding the token to some of the other urls at some point - but we can wait and see.

ChrisBAshton: > I think it might be clearer if this function was just draft_govuk_url and accepted a path rather than a slug then could be just Plek.find("draft-origin", external: true) + path

Thanks for the suggestion, I'll do this 👍

db/schema.rb
```diff
@@ -125,6 +125,7 @@
     t.bigint "step_by_step_page_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["position"], name: "index_steps_on_position"
```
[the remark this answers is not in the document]
ChrisBAshton: 🤦‍♂ 

Thanks! This happens occasionally, and I usually catch it before committing. I wonder if it trips up anyone else?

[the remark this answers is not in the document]
kevindew: It looks like you have an extra index. you might want to do db:reset or remove the index manually so there's less difference.

Hopefully we don't have the situation where our prod database has this index and is out of sync with db/schema.rb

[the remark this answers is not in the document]
ChrisBAshton: Have now done a reset 👍 I was working on step positions db schema a while ago, that could be it.

PR #822, opened 2019-10-09
Generate breadcrumbs and headings consistently

alex-ju:
This PR updates a series of views to consistently use the admin_layout file for placing breadcrumbs, title and context. This improves the generated markup and makes the copy consistent across pages: the copy for each section we show on the summary page is played back in the heading on the edit pages.


app/views/step_by_step_pages/new.html.erb
```diff
@@ -1,22 +1,17 @@
 <%
   links = [
     {
-      text: 'Step by step publisher',
+      text: "Step by step publisher",
       href: step_by_step_pages_path
     },
     {
-      text: 'Create new'
+      text: "Create a new step by step navigation"
```
ChrisBAshton: I think we can drop the word "navigation". What do you think?

app/views/step_by_step_pages/new.html.erb
```diff
@@ -1,22 +1,17 @@
 <%
   links = [
     {
-      text: 'Step by step publisher',
+      text: "Step by step publisher",
```
ChrisBAshton: This is unchanged apart from double quotes - we use single quotes in most places in the other edited files, and this file was mostly single quotes before the change too. Could you look at that improving that consistency?

app/views/step_by_step_pages/new.html.erb
alex-ju: Sounds sensible and makes it consistent with the primary action on the summary page, will update

alex-ju:
Thanks for the review, @ChrisBAshton I updated with your suggestions.

PR #824, opened 2019-10-10
Add contextual guidance to edit pages

alex-ju:
This adds contextual guidance to the add/edit step-by-step and add/edit step pages.

[Trello card - Add Contextual guidance to Add/Edit Content page](https://trello.com/c/nNcyHxaw)
[Trello card - Add Contextual guidance to Add/Edit Step page](https://trello.com/c/d9XCwVa2)

PR #825, opened 2019-10-10
Add contextual guidance page

JonathanHallam:
This adds the contextual guidance page, links to it from SBS's that are in review and tests for same. This will provide a place where 2i reviewers can go to see guidance on what the intended use of attributes are.

Trello - https://trello.com/c/K4Uyg9m6/127-add-a-view-giving-some-contextual-guidance-to-2i-reviewers

### DIFF
--- a/app/views/step_by_step_pages/guidance.html.erb
+++ b/app/views/step_by_step_pages/guidance.html.erb
@@ -1,0 +1,48 @@
+<%
+  links = [
+    {
+      text: 'Step by step publisher',
+      href: step_by_step_pages_path
+    },
+    {
+      text: @step_by_step_page.title,
+      href: @step_by_step_page
+    },
+    {
+      text: 'Guidance'
+    }
+  ]
+%>
+<%= render 'shared/steps/step_breadcrumb', links: links %>
+
+<% guidance = t("guidance") %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h3 class="govuk-heading-m"><%= guidance.keys[0].to_s.titleize %></h3>
+    <% guidance[:new_step_by_step][:shared].each do |title, content| %>
+      <h2 class="govuk-heading-s"><%= title.to_s.titleize %></h2>
+      <p class="govuk-body"><%= content %></p>
+    <% end %>
+
+    <h3 class="govuk-heading-m"><%= guidance.keys[1].to_s.titleize %></h3>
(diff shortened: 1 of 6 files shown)


app/views/step_by_step_pages/show/_required_actions.html.erb
```diff
@@ -45,9 +45,15 @@
 <% if @step_by_step_page.status.submitted_for_2i? %>
   <%= render "govuk_publishing_components/components/inset_text" do %>
     <p class="govuk-body">Not yet claimed for 2i</p>
+    <% if current_user.permissions.include?("2i reviewer") %>
+      <a href="/step-by-step-pages/<%=@step_by_step_page.id%>/guidance">Guidance</a>
```
brucebolt: There's a `link_to` helper method that will create the relevant HTML for you and also include the correct path, e.g.
```
<%= link_to "Guidance", step_by_step_page_guidance_path(@step_by_step_page) %>
```

config/routes.rb
```diff
@@ -28,6 +28,7 @@
     get :unpublish
     post :unpublish
     post :unschedule
+    get :guidance
```
brucebolt: Could this be included in the right place in this block to keep everything alphabetical?

brucebolt:
You may also want to coordinate with Alex, since he's implemented the guidance in a slightly different format in the YAML file: https://github.com/alphagov/collections-publisher/pull/824/files#diff-37c9b835927e3c363588f172de9bb6fa

app/views/step_by_step_pages/guidance.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    }
+  ]
+%>
+<%= render 'shared/steps/step_breadcrumb', links: links %>
+
+<% guidance = t("guidance") %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h3 class="govuk-heading-m"><%= guidance.keys[0].to_s.titleize %></h3>
+    <% guidance[:new_step_by_step][:shared].each do |title, content| %>
```
leenagupte: What happens if another item is added betwixt `title` and `content` in the YAML file? Perhaps it would be safer to reference them directly?

e.g. `guidance[:new_step_by_step][:shared][:title].to_s.titleize`

leenagupte: Also, there isn't an entry for `content` in the YAML file

PR #826, opened 2019-10-11
Use update_column for backfilling auth_bypass_id values

brucebolt:
This migration was written to use `update`.  This changes the `updated_at` field during the migration (making it look like all step by steps have been modified by a content designer).  Changing to `update_column` so this time is not changed.

This migration has not yet been deployed to production.

Trello card: https://trello.com/c/jKRvjyAe


kevindew:
Woa good catch!

PR #827, opened 2019-10-11
Add link to Content Tagger for step by steps

brucebolt:
Adds a link to the step by step in Content Tagger under a new "Tags" section.

The label will always say "Edit", since the tags can be changed regardless of whether the step by step is draft or published.
Screen Shot 2019-10-11 at 15 20 33

Trello card: https://trello.com/c/uoBZfPNn

### DIFF
--- a/app/presenters/step_by_step_page_presenter.rb
+++ b/app/presenters/step_by_step_page_presenter.rb
@@ -103,6 +103,21 @@ class StepByStepPagePresenter
     }
   end
 
+  def tags_section
+    {
+      borderless: true,
+      id: "tags",
+      title: "Tags",
+      edit: {
+        link_text: "Edit",
+        href: "#{Plek.find('content-tagger', external: true)}/taggings/#{step_by_step_page.content_id}",
+        data_attributes: {
+          gtm: "edit-topics",
+        },
+      },
+    }
+  end
+
   def last_saved
     last_saved_time = format_full_date_and_time(step_by_step_page.updated_at)
     return "#{last_saved_time} by #{step_by_step_page.assigned_to}" if assigned?
--- a/app/views/step_by_step_pages/show.html.erb
+++ b/app/views/step_by_step_pages/show.html.erb
@@ -76,6 +76,9 @@
         @step_by_step_page_presenter.secondary_links_settings
       ]
     }  %>
+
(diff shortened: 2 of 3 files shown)


app/views/step_by_step_pages/show.html.erb
```diff
@@ -76,6 +76,16 @@
         @step_by_step_page_presenter.secondary_links_settings
       ]
     }  %>
+
+    <%= render "govuk_publishing_components/components/summary_list", {
```
leenagupte: I think you can call this with just `topic_settings` in the same way that `summary_list_params` is used further up the file. 

e.g.
`<%= render "govuk_publishing_components/components/summary_list", @step_by_step_page_presenter.topic_settings  %>`

app/presenters/step_by_step_page_presenter.rb
```diff
@@ -103,6 +103,26 @@ def secondary_links_settings
...
+          edit: {
+            link_text: "Edit",
+            href: "#{Plek.find('content-tagger', external: true)}/taggings/#{step_by_step_page.content_id}",
+            data_attributes: {
+              gtm: "edit-topics",
+            },
+          },
+        },
+      ],
+    }
```
leenagupte: I think this can just be: 

```
    {
      borderless: true,
      id: "tags",
      title: "Tags",
      edit: {
        href: "#{Plek.find('content-tagger', external: true)}/taggings/#{step_by_step_page.content_id}",
          data_attributes: {
            gtm: "edit-topics",
           },
         },
      ],
    }
```

brucebolt: This will give a layout like this:
Screen Shot 2019-10-11 at 15 20 33

Are we happy just having the word "Tags", or should it be "Topics"?

leenagupte: It does say "Tags" in the card, so I would leave it as is.

leenagupte: Do you really need to nest "edit" inside "items"?

brucebolt: No, I haven't yet pushed the code that rendered the above screenshot

brucebolt: Pushed the updated code

leenagupte: Oops! Sorry.

PR #829, opened 2019-10-11
Add contextual guidance for notes

alex-ju:
Add Contextual guidance to change notes:
- internal change notes (on history page)
- notes when submitting for 2i
- notes when requesting changes or approving
- notes when publishing or scheduling

[Trello card](https://trello.com/c/EOm5X8Mp)

### DIFF
--- a/app/views/review/submit_2i_verdict.erb
+++ b/app/views/review/submit_2i_verdict.erb
@@ -23,26 +23,31 @@
   <%= render "shared/steps/form_errors", resource: @step_by_step_page %>
 <% end %>
 
+<%
+  review_phrase = "2i_review_#{approved ? 'approved' : 'request_changes'}"
+  textarea_name = approved ? "additional_comment" : "requested_change"
+  button_text = approved ? "Yes, approve 2i" : "Send change request"
+%>
+
 <%= form_tag do %>
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
+  <%= render "govuk_publishing_components/components/contextual_guidance", {
+    html_for: "additional-comments",
+    title: t("step_by_step_pages.#{review_phrase}.title"),
+    content: t("step_by_step_pages.#{review_phrase}.guidance")
+  } do %>
       <%= render "govuk_publishing_components/components/textarea", {
         label: {
-          text: approved ? "Add a note (optional)" : "Describe changes required"
+          text: t("step_by_step_pages.#{review_phrase}.label")
         },
-        name: approved ? "additional_comment" : "requested_change",
+        id: "additional-comments",
+        name: textarea_name,
       } %>
-    </div>
-  </div>
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
       <%= render "govuk_publishing_components/components/button", {
-        text: approved ? "Yes, approve 2i" : "Send change request"
+        text: button_text,
+        margin_bottom: true
       } %>
-    </div>
-    <div class="govuk-grid-column-full govuk-!-margin-top-3">
-      <%= link_to "Cancel", @step_by_step_page, class: "govuk-link" %>
-    </div>
-  </div>
+
+      <%= tag.p link_to("Cancel", @step_by_step_page, class: "govuk-link"), class: "govuk-body" %>
+    <% end %>
 <% end %>
--- a/app/views/review/submit_for_2i.html.erb
+++ b/app/views/review/submit_for_2i.html.erb
@@ -19,25 +19,24 @@
 <% content_for :context, 'Submit for 2i' %>
 
 <%= form_tag do %>
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
-      <%= render "govuk_publishing_components/components/textarea", {
-        label: {
-          text: "Additional comments (optional)"
-        },
-        name: "additional_comments"
-      } %>
-    </div>
-  </div>
+  <%= render "govuk_publishing_components/components/contextual_guidance", {
+    html_for: "additional-comments",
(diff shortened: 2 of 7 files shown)


app/views/review/submit_2i_verdict.erb
```diff
@@ -23,26 +23,25 @@
...
-      <%= link_to "Cancel", @step_by_step_page, class: "govuk-link" %>
-    </div>
-  </div>
+    <%= render "govuk_publishing_components/components/button", {
+      text: approved ? "Yes, approve 2i" : "Send change request",
+      margin_bottom: true
+    } %>
+
+    <%= tag.p link_to("Cancel", @step_by_step_page, class: "govuk-link"), class: "govuk-body" %>
+  <% end %>
```
kevindew: There's a lot of small conditionals here that make this somewhat hard to read.

I wonder if the whole thing should be in a conditional.

For example:

```
<% if approved %>
  <%= render "govuk_publishing_components/components/contextual_guidance" .. %>
<% else %>
     <%= render "govuk_publishing_components/components/contextual_guidance" .. %>
<% end %>
```

or set some variables first,

for example:

```
phrase_prefix = "step_by_step_pages.2i_review_#{approved ? 'approved' : 'request_changes'}"
textarea_name = approved ? "additional_comment" : "requested_change"
button_text = approved ? "Yes, approve 2i" : "Send change request"

```

app/views/step_by_step_pages/internal_change_notes.html.erb
```diff
@@ -20,25 +20,30 @@
 
 <%= render 'nav', step_by_step_page: @step_by_step_page, active: "internal-change-notes" %>
 
+<%= render "govuk_publishing_components/components/contextual_guidance", {
+  html_for: "internal-change-note-description",
+  title: t("step_by_step_pages.internal_change_notes.label"),
+  content: t("step_by_step_pages.internal_change_notes.guidance")
+} do %>
```
kevindew: Would this normally be inside the form and wrap just the textarea?

config/locales/en.yml
```diff
@@ -62,10 +62,25 @@ en:
     reorder:
       instructions_markdown: Use the up/down buttons on the right to reorder steps, or click and hold on a step to reorder using drag and drop.
       automatically_updated_markdown: Step numbers will be automatically updated on save.
+    internal_change_notes:
+      label: Internal note
+      guidance: "Add a summary of your changes and why. For example: what’s been added, removed or replaced. Provide any context that helps understand what you’ve done. For example: a link to zendesk and trello ticket if there’s one."
```
kevindew: No biggie as there is already precedent but for long lines like this you might want to use the `>` folded scalar property from yaml so these long lines can span multiple lines: https://yaml.org/spec/1.2/spec.html#id2760844

app/views/review/submit_2i_verdict.erb
alex-ju: Good suggestion @kevindew! updated

config/locales/en.yml
alex-ju: Updated to use multi-line markup so it's easier to read and edit.

app/views/step_by_step_pages/internal_change_notes.html.erb
alex-ju: Yes, it would normally, but since we only have one input I thought I can leave it close to the textarea and submit to make it clear there are not other inputs in the page. Let me know if it's a blocker though, I can go through them and update the nesting order.

config/locales/en.yml
ChrisBAshton: It doesn't appear to have been updated... maybe it was updated in the wrong commit?

app/views/step_by_step_pages/internal_change_notes.html.erb
ChrisBAshton: Hmmm I'm with Kevin on this one I think. If someone wants to add another input to this page in the future, they'd have to re-nest everything appropriately so that this contextual guidance doesn't apply to their input. I don't think there's much value in making it clear there's currently only one input on the page.

config/locales/en.yml
ChrisBAshton: Yes, looks like you updated it in d9e7a571476336772a4ef9e25ba29887edc5a48d, ideally you would do it here rather than change it halfway through the PR.

app/views/review/submit_for_2i.html.erb
```diff
@@ -18,26 +18,24 @@
...
-    <div class="govuk-grid-column-full">
-      <%= render "govuk_publishing_components/components/button", {
-        text: "Submit for 2i"
-      } %>
-    </div>
-    <div class="govuk-grid-column-full govuk-!-margin-top-3">
-      <%= link_to 'Cancel', @step_by_step_page, class: "govuk-link" %>
-    </div>
-  </div>
+<%= render "govuk_publishing_components/components/contextual_guidance", {
```
ChrisBAshton: Same suggestion on form nesting here.

config/locales/en.yml
```diff
@@ -69,6 +69,10 @@ en:
       title: Notify users about this change?
       label: Public change note
       hint: Describe the change for users. This change note will be emailed to subscribers.
+    submit_for_2i:
+      label: Additional comments (optional)
+      title: Submit for 2i
+      guidance: "Add a summary of your changes and why. For example: what’s been added, removed or replaced. Provide any context that helps understand what you’ve done. For example: a link to zendesk and trello ticket if there’s one."
```
ChrisBAshton: Same suggestion on multi-line approach here.

alex-ju:
Thank you for the review @kevindew and @ChrisBAshton.
I addressed all the comments as suggested.

PR #831, opened 2019-10-14
Add guidance to sidebar settings and secondary links pages

alex-ju:
This PR
 - adds guidance to sidebar settings page and fixes broken markup
 - adds guidance to secondary links page and fixes broken markup

[Trello card](https://trello.com/c/ij0U9BPn)

### DIFF
--- a/app/views/navigation_rules/edit.html.erb
+++ b/app/views/navigation_rules/edit.html.erb
@@ -21,67 +21,64 @@
 <%= render 'shared/steps/form_notice', message: notice, status: "success" %>
 
 <div class="govuk-grid-row">
-    <%= form_for(step_by_step_page_navigation_rules_path(@step_by_step_page), method: :put) do |form| %>
-      <div class="govuk-grid-column-full">
-        <table class="govuk-table table-stepnav-rules" data-module="filterable-table">
-          <thead class="govuk-table__head">
-            <tr>
-              <th class="govuk-table__header" scope="col">Page title</th>
-              <th class="govuk-table__header" scope="col">Path</th>
-              <th class="govuk-table__header" scope="col">Sidebar content of page</th>
-            </tr>
-          </thead>
+  <%= form_for(step_by_step_page_navigation_rules_path(@step_by_step_page), method: :put) do |form| %>
+    <div class="govuk-grid-column-full">
+      <table class="govuk-table table-stepnav-rules">
+        <thead class="govuk-table__head">
+          <tr>
+            <th class="govuk-table__header" scope="col">Page title</th>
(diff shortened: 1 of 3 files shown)


app/views/secondary_content_links/index.html.erb
```diff
@@ -90,7 +81,7 @@
         <tr class="govuk-table__row">
           <td class="govuk-table__cell" colspan="3">
             <p class="govuk-hint">No secondary links have been added yet.</p>
-          </tr>
+          </td>
```
kevindew: 🕵 👍

app/views/navigation_rules/edit.html.erb
```diff
@@ -20,6 +20,12 @@
 
 <%= render 'shared/steps/form_notice', message: notice, status: "success" %>
 
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-6">
+    <%= render_markdown t("sidebar_settings.guidance_markdown") %>
+  </div>
+</div>
```
kevindew: Not sure if I should comment to you on this as the implementer but this copy feels a bit out of context when you browse this page.

Screen Shot 2019-10-14 at 15 52 09

It feels like this is missing something that introduces the options - or maybe the explanations should be after the table? I get though that we're in a bit of a pickle as there isn't a spot for contextual guidance.

If it's all product ok'd happy to tick it though as code looks good.

config/locales/en.yml
```diff
@@ -90,14 +90,13 @@ en:
 
   secondary_content_links:
     guidance_markdown: |
-      ##What is secondary content?
```
kevindew: This looks loads better without this question

config/locales/en.yml
```diff
@@ -90,14 +90,13 @@ en:
...
-
       Secondary content is content which is not linked to in the step by step, but has the step by step sidebar showing on it.
 
       We use it for content which:
 
-      - could help you complete a task but is not the main piece of content included in the step by step
+      - could help you complete a task but is not the main piece of content included in the step by step.
```
kevindew: Do you mean for the list item below to have a full stop too? Our styleguide tends to not have the periods at the end of list items: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps

alex-ju: Good spot. I've just pasted from the document without looking a the diff. Removed the full stop.

app/views/navigation_rules/edit.html.erb
alex-ju: Agree with you that it feels a bit out of the blue, can raise it with product to see if they can enhance it.

PR #834, opened 2019-10-16
Show Revert to draft link when step by step is submitted for 2i and in review

AgaDufrat:
This allows the author to revert to draft when the Step by step is in `submitted_for_2i` state or in `in_review` state.

**Before**
before

**After**
after

[Trello card](https://trello.com/c/FoRZjA0I/164-when-a-sbs-is-submitted-for-2i-theres-no-revert-to-draft-link-only-an-unpublish-option)

_Note:_ Functionality that allows the reviewer to abandon review will be added as a separate PR.

### DIFF
--- a/app/helpers/step_nav_actions_helper.rb
+++ b/app/helpers/step_nav_actions_helper.rb
@@ -16,6 +16,13 @@ module StepNavActionsHelper
       !step_by_step_page.broken_links_found?
   end
 
+  def can_revert_to_draft?(step_by_step_page, user)
+    return true if step_by_step_page.status.approved_2i?
+
+    %w[submitted_for_2i in_review].include?(step_by_step_page.status) &&
+      step_by_step_page.review_requester_id == user.uid
+  end
+
 private
 
   def can_claim_first_review?(step_by_step_page, user)
--- a/spec/helpers/step_nav_actions_helper_spec.rb
+++ b/spec/helpers/step_nav_actions_helper_spec.rb
@@ -118,4 +118,42 @@ RSpec.describe StepNavActionsHelper do
       end
     end
   end
+
+  describe "#can_revert_to_draft?" do
+    let(:step_by_step_page) { create(:step_by_step_page) }
+
+    allowed_statuses = %w[submitted_for_2i in_review approved_2i]
+    user_restricted_statuses = allowed_statuses - %w[approved_2i]
+    disallowed_statuses = StepByStepPage::STATUSES - allowed_statuses
+
+    it "can revert to draft if it is approved 2i" do
+      step_by_step_page.status = "approved_2i"
+
+      expect(can_revert_to_draft?(step_by_step_page, user)).to be true
+    end
+
+    user_restricted_statuses.each do |user_restricted_status|
+      it "can revert to draft if it is #{user_restricted_status.humanize.downcase} and the user is a review requester" do
+        step_by_step_page.status = user_restricted_status
+        step_by_step_page.review_requester_id = user.uid
+
+        expect(can_revert_to_draft?(step_by_step_page, user)).to be true
+      end
+
+      it "cannot revert to draft if it is #{user_restricted_status.humanize.downcase} and the user is not a review requester" do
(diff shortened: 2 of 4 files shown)


spec/helpers/step_nav_actions_helper_spec.rb
```diff
@@ -72,4 +72,35 @@
       expect(helper.can_submit_2i_review?(step_by_step_page, reviewer_user)).to be true
     end
   end
+
+  describe "#can_revert_to_draft?" do
+    let(:step_by_step_page) { create(:step_by_step_page) }
+
+    it "can reverted to draft if it is approved 2i" do
```
ChrisBAshton:
```suggestion
    it "can be reverted to draft if it is approved 2i" do
```

spec/helpers/step_nav_actions_helper_spec.rb
```diff
@@ -72,4 +72,35 @@
...
+    permitting_statuses.each do |permitting_status|
+      it "can be reverted to draft if it is #{permitting_status.humanize.downcase} and the user is a review requester" do
+        step_by_step_page.status = permitting_status
+        step_by_step_page.review_requester_id = user.uid
+
+        expect(can_revert_to_draft?(step_by_step_page, user)).to be true
+      end
+    end
+
+    not_permitting_statuses = %w[draft scheduled published]
```
ChrisBAshton:
```suggestion
    not_permitting_statuses = StepByStepPage::STATUSES - permitting_statuses
```

app/helpers/step_nav_actions_helper.rb
```diff
@@ -9,6 +9,13 @@ def can_submit_2i_review?(step_by_step_page, user)
       step_by_step_page.reviewer_id == user.uid
   end
 
+  def can_revert_to_draft?(step_by_step_page, user)
+    return true if step_by_step_page.status.approved_2i?
+
+    %w[submitted_for_2i in_review].include?(step_by_step_page.status) &&
+      step_by_step_page.review_requester_id == user.uid
```
leenagupte: Is there are reason why only the user who submitted for 2i can revert to draft?

AgaDufrat: As per Craig's comment:
1) the author should be able to Revert to draft
2) the reviewer should be able to Abandon review, changing it to submitted_for_2i. This is different functionality which is NOT part of this card. We have a separate card for this.
3) other people with review permissions should be able to Claim review, leaving status unchanged but changing the reviewer field.

spec/helpers/step_nav_actions_helper_spec.rb
AgaDufrat: I don't think this will work because `approved_2i` in not in `permitting_statuses`

spec/helpers/step_nav_actions_helper_spec.rb
AgaDufrat: This has already been done. I didn't change any logic about showing _Revert to draft_ when the status is `approved_2i`. I only moved it into the method, so I didn't think there is a need to explain that here. Maybe I should split it into 2 commits so it's clearer what has been added?

app/helpers/step_nav_actions_helper.rb
leenagupte: Could you add this to the commit message? and add a test for when the user is not the review requester?

spec/helpers/step_nav_actions_helper_spec.rb
ChrisBAshton: Then maybe this?

```suggestion
    not_permitting_statuses = StepByStepPage::STATUSES - permitting_statuses - "approved_2i"
```

leenagupte: Perhaps:
```
allowed_statuses = %w[submitted_for_2i in_review approved_2i]
user_restricted_statuses = allowed_statuses - "approved_2i"
disallowed_statuses = StepByStepPage::STATUSES - allowed_statuses
```

I'm not sure if that makes anything clearer of not.

PR #835, opened 2019-10-16
Fix bug that caused box to appear with no suggestions

JonathanHallam:
Adds in a check for the status of a SBS and provides guidance accordingly. This should stop a bug where the "To publish this SBS you need to" box appeared without giving any suggestions.

Trello - https://trello.com/c/7IzRnut3/163-to-publish-this-sbs-you-need-to-doesnt-list-any-things-the-user-needs-to-do

### DIFF
--- a/app/views/step_by_step_pages/show/_required_actions.html.erb
+++ b/app/views/step_by_step_pages/show/_required_actions.html.erb
@@ -38,6 +38,12 @@
           </li>
         <% end %>
       <% end %>
+
+      <% if @step_by_step_page.status.draft? || @step_by_step_page.status.submitted_for_2i? %>
+        <li>
+          <p class="govuk-body">Get 2i approval</p>
+        </li>
+      <% end %>
     </ul>
   <% end %>
 <% end %>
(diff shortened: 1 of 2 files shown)


app/views/step_by_step_pages/show/_required_actions.html.erb
```diff
@@ -38,6 +38,12 @@
           </li>
         <% end %>
       <% end %>
+
+      <% if %w(draft submitted_for_2i).include?(@step_by_step_page.status) %>
```
leenagupte: I wonder if this should be split in to different actions? 
e.g. if status is `draft`, the action should be "Submit for 2i" 
if status is `submitted for 2i` and user is not the drafter, the action should be "2i changes"
if status is `submitted for 2i` and user is the drafter, the action should be "2i review required"
if status is `in_review`, the action is ....

Ideally if there was more time, these could be encapsulated in methods, one for each action, e.g. `fix_broken_links?`.

Disclaimer: the words need more thought.

PR #837, opened 2019-10-17
Broken links bug

JonathanHallam:
Trello - https://trello.com/c/XM016j2s/149-bug-check-for-broken-links-when-there-are-no-links

## What's changed and why?
Stops a message from appearing at the top of the page asking the user to "Check for broken links" if the step-by-step doesn't contain any links.

### DIFF
--- a/app/helpers/step_nav_actions_helper.rb
+++ b/app/helpers/step_nav_actions_helper.rb
@@ -9,6 +9,12 @@ module StepNavActionsHelper
       step_by_step_page.reviewer_id == user.uid
   end
 
+  def must_check_for_broken_links?(step_by_step_page)
+    step_by_step_page.steps.any? &&
+      !step_by_step_page.links_checked_since_last_update? &&
+      !step_by_step_page.broken_links_found?
+  end
+
 private
 
   def can_claim_first_review?(step_by_step_page, user)
--- a/app/helpers/step_nav_actions_helper.rb
+++ b/app/helpers/step_nav_actions_helper.rb
@@ -11,6 +11,7 @@ module StepNavActionsHelper
 
   def must_check_for_broken_links?(step_by_step_page)
     step_by_step_page.steps.any? &&
+      contains_links?(step_by_step_page) &&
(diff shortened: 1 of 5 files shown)


spec/factories.rb
```diff
@@ -67,6 +67,15 @@
     end
   end
 
+  factory :step_by_step_page_with_steps_and_links, parent: :step_by_step_page_with_steps do
```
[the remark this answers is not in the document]
JonathanHallam: So I've played around with this a bit and adding the navigation rules to the `:step` creates a lot of errors. This is because I can't really just add a link but have to add the navigation rules. I think I'd be able to make it work by changing all of those tests but that would be a much bigger diff. Do you think it's worth doing for the sake of cleaner code?

spec/helpers/step_nav_actions_helper_spec.rb
```diff
@@ -72,4 +73,48 @@
...
+        create(:step, step_by_step_page: step_by_step_page, contents: "- [Good stuff](http://foo.co.uk/good/stuff)")
+
+        expect(helper.must_check_for_broken_links?(step_by_step_page)).to be true
+      end
+
+      it "returns true if there are internal links and link checker hasn't been run" do
+        step_by_step_page = create(:draft_step_by_step_page)
+
+        expect(helper.must_check_for_broken_links?(step_by_step_page)).to be true
+      end
```
ChrisBAshton: This is repeating the block on line 85.

app/helpers/step_nav_actions_helper.rb
```diff
@@ -29,4 +30,12 @@ def can_take_over_review?(step_by_step_page, user)
       step_by_step_page.reviewer_id != user.uid &&
       user.has_permission?("2i reviewer")
   end
+
+  def links?(step_by_step_page)
```
ChrisBAshton: This is fine, but I think `contains_links?` or `has_links?` would be a little clearer.

leenagupte:
@ChrisBAshton Comments have been addressed and fixed-up.

PR #838, opened 2019-10-17
Update guidance on using number

Tobiogunsina:
Update guidance on using number


alex-ju:
This change has already been implemented

PR #839, opened 2019-10-18
Remove fix broken links requirement

leenagupte:
Trello: https://trello.com/c/t2MylOor
Reverts changes made in PR #754 

## What's changed?

Remove the requirement that forced users to fix all of the broken links before they can publish step-by-steps.

## Why?

In many cases step-by-steps contain links to draft content or other draft step-by-steps. It's also possible to for draft step-by-steps to be linked to each other. As Link Checker API doesn't handle draft content, and can only check links that are publicly accessible, publishers could end up in a chicken or egg situation where they can't publish any step-by-steps.

### Before
Screen Shot 2019-10-18 at 15 42 34

### After
Screen Shot 2019-10-18 at 15 42 44

### DIFF
--- a/spec/helpers/step_nav_actions_helper_spec.rb
+++ b/spec/helpers/step_nav_actions_helper_spec.rb
@@ -110,9 +110,9 @@ RSpec.describe StepNavActionsHelper do
         expect(helper.must_check_for_broken_links?(step_by_step_page)).to be true
       end
 
-      it "returns false if link checker has been run and there are no broken links" do
+      it "returns false if link checker has been run since the last update" do
         step_by_step_page = create(:draft_step_by_step_page)
-        stub_link_checker_report_success(step_by_step_page.steps.first)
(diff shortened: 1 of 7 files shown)


spec/helpers/step_nav_actions_helper_spec.rb
```diff
@@ -116,9 +116,9 @@
         expect(helper.must_check_for_broken_links?(step_by_step_page)).to be true
       end
 
-      it "returns false if link checker has been run and there are no broken links" do
+      it "returns false if link checker has been run and there are broken links" do
```
ChrisBAshton: I'd maybe rephrase this to:

> `it "returns false even if the latest link report found broken links"`

Current wording sounds like it should return false _because_ it found broken links, but really the presence of broken links is inconsequential.

leenagupte:
@ChrisBAshton Your comments have been addressed and fixed-up.

PR #844, opened 2019-10-21
Update status on page refresh after update

1pretz1:
This commit makes sure the status of the step by step is in the correct
state after an edit has been made. Previously, the code to update the
state was inside a sidekiq worker which may not have completed upon the
new page being loaded. This led to the wrong links being displayed until
the worker has processed the job.

Trello:
https://trello.com/c/L4IbSFuv/111-status-doesnt-update-to-draft-until-page-is-refreshed

### DIFF
--- a/app/controllers/secondary_content_links_controller.rb
+++ b/app/controllers/secondary_content_links_controller.rb
@@ -6,7 +6,7 @@ class SecondaryContentLinksController < ApplicationController
     @secondary_content_link = step_by_step_page.secondary_content_links.new(content_item)
 
     if @secondary_content_link.save
-      update_downstream
+      StepByStepUpdater.call(step_by_step_page, current_user)
 
       redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id), notice: "Secondary content was successfully linked."
     else
@@ -17,7 +17,7 @@ class SecondaryContentLinksController < ApplicationController
 
   def destroy
     if secondary_content_link.destroy
-      update_downstream
+      StepByStepUpdater.call(step_by_step_page, current_user)
 
       redirect_to step_by_step_page_secondary_content_links_path(step_by_step_page.id), notice: "Secondary content link was successfully deleted."
     else
@@ -38,10 +38,6 @@ private
 
   def secondary_content_link
     @secondary_content_link ||= step_by_step_page.secondary_content_links.find(params[:id])
-  end
-
-  def update_downstream
-    StepByStepDraftUpdateWorker.perform_async(step_by_step_page.id, current_user.name)
   end
 
   def content_item
--- a/spec/services/step_nav_publisher_spec.rb
+++ b/spec/services/step_nav_publisher_spec.rb
@@ -25,59 +25,26 @@ RSpec.describe StepNavPublisher do
       let(:step_nav) { create(:published_step_by_step_page) }
 
       it "defines :auth_bypass_ids in the publishing-api payload" do
+        step_nav.mark_draft_updated
         StepNavPublisher.update(step_nav)
-        content_id = /[0-9a-f]{5,40}/ # changes on each test run, as does Array value below
         payload = hash_including(:auth_bypass_ids => instance_of(Array))
-        expect(Services.publishing_api).to have_received(:put_content).with(content_id, payload)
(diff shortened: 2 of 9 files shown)


spec/services/step_nav_publisher_spec.rb
```diff
@@ -24,57 +24,23 @@
     context "published step by step having a step edited" do
       let(:step_nav) { create(:published_step_by_step_page) }
 
-      it "defines :auth_bypass_ids in the publishing-api payload" do
+      it "defines :access_limited in the publishing-api payload" do
```
ChrisBAshton: This doesn't match with what's being tested (`auth_bypass_ids` on line 30)

spec/services/update_step_by_step_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    allow(Services.publishing_api).to receive(:put_content)
+    allow(Services.publishing_api)
+      .to receive(:lookup_content_ids)
+      .and_return({})
+
+    UpdateStepByStep.call(step_by_step, current_user)
+  end
+
+  after do
+    expect(StepByStepDraftUpdateWorker.jobs.size).to eq(1)
```
ChrisBAshton: This is a nice sanity-check 👍

spec/services/step_nav_publisher_spec.rb
```diff
@@ -24,57 +24,23 @@
...
-        expect(step_nav.status).to eq "draft"
-      end
-
       it "does not change the auth_bypass_id when step by step in draft state is edited" do
         StepNavPublisher.update(step_nav)
         expect(step_nav.auth_bypass_id).to eq "33ac75d8-4adf-48a7-acb2-bbf4e7f644a3"
       end
 
       it "changes the auth_bypass_id when step by step in published state is edited" do
+        allow(SecureRandom).to receive(:uuid).and_return("bd814775-304c-46d0-919f-f9e0f6cba4e9")
```
ChrisBAshton: Could you move this content ID to its own variable, so we don't have to repeat the string on lines 42 and 46?

app/controllers/secondary_content_links_controller.rb
```diff
@@ -40,10 +40,6 @@ def secondary_content_link
     @secondary_content_link ||= step_by_step_page.secondary_content_links.find(params[:id])
   end
 
-  def update_downstream
-    StepByStepDraftUpdateWorker.perform_async(step_by_step_page.id, current_user.name)
```
ChrisBAshton: Instead of making a new class which just does one thing (a bit of a smell), I'd suggest exposing a new method we can call on `StepByStepDraftUpdateWorker` and just update the definition inside `update_downstream`.

Something like:

```ruby
def update_downstream
  StepByStepDraftUpdateWorker.
    update_status_sync(step_by_step_page).
    perform_async(step_by_step_page.id, current_user.name)
end
```

Would mean the only code that needs changing in each file is the `update_downstream` method.

spec/services/step_nav_publisher_spec.rb
1pretz1: Good spot, don't know why this changed but have changed back to `auth_bypass_ids`.

spec/services/step_nav_publisher_spec.rb
1pretz1: Sure!

app/services/update_step_by_step.rb
```diff
@@ -0,0 +1,6 @@
+class UpdateStepByStep
```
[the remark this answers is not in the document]
1pretz1: Hmm, which others are you referring to? I've had a look in the repo and can't find any classes that have a suffix of `Updater` or the presence of an updater directory. I've updated the class name anyway to `StepByStepPageUpdater` to make you happy! 😄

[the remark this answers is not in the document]
leenagupte: 🤗 I was thinking about `StepNavPublisher` and `StepContentParser` which follow the pattern of "Thing" + "What it's doing". Thanks for updating!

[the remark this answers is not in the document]
1pretz1: Ohh okay, good point!

app/controllers/secondary_content_links_controller.rb
ChrisBAshton: Fair! 👍

PR #846, opened 2019-10-22
Restructure locales YAML file

AgaDufrat:
## User story
**As a** Content Designer
**I want** to easily modify the contextual guidance and make sure it's consistent across the app
**So that** the various places the guidance appears remain synchronised

## What
- Shared contextual guidance is stored under `guidance` key, which removes repetition
- Rewrite guidance.html.erb so it doesn't rely on the structure of the YAML file
- Change formatting of the guidance page as shown below

| Before  | After |
| ------------- | ------------- |
| guidance_before  | new  |

[Trello card](https://trello.com/c/r0ycsFmz/177-change-the-structure-of-the-contextual-guidance-yaml-file-so-each-piece-of-guidance-only-appears-once-in-the-file)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -64,21 +64,7 @@ en:
       time:
         label: Time
   sidebar_settings:
-    guidance_markdown: |
-      ## Showing or hiding the sidebar
-
-      ### Always show navigation
-
-      Sidebar will always show on the piece of content. Choose this option if the content is about a task that is always completed as part of the step by step journey. For example, booking a theory driving test is always part of the wider journey of learning to drive.
-
-      ### Show navigation if users comes from step by step
-
-      Sidebar will only show if you clicked on a link in the step by step to get there For example: If the page is part of multiple user journeys and showing step by steps risks limiting the way the page is used'.
-
-      ### Never show navigation
-
-      Chose this option when you're linking to a page that it would never be appropriate for the step by step side bar to appear on.
-
+    header: Showing or hiding the sidebar
   secondary_content_links:
     guidance_markdown: |
       Secondary content is content which is not linked to in the step by step, but has the step by step sidebar showing on it.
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -122,17 +118,17 @@ en:
...
+        Add step by step after your title unless it will be inappropriate or make the title excessively long. Keep it under 85 characters
+      slug:
+        "Slug is the link for your step by step. It should be lowercase and you should use hyphens to separate words. For example: get-driving-license"
+      introduction:
+        The introduction explains what the step by step helps the user do, or what it is. Keep it short. Tell users if it is not relevant to them. Do not include any important information in here that doesn’t exist elsewhere.
+      meta_description:
+        Meta description is shown in the GOV.UK search results. It should explain what the step by step helps the user do, or what it is. Keep it under 160 characters.
     new_step:
-      shared:
+        shared:
```
leenagupte: Is the `shared` key still needed?

config/locales/en.yml
```diff
@@ -133,26 +119,24 @@ en:
...
+      header: Order of steps
+      order:
         Steps and tasks should be listed in the order in which users need to complete them.
-      origin:
-        Use the up/down buttons on the right to reorder steps, or click and hold on a step to reorder using drag and drop.
-        Step numbers will be automatically updated on save.
     sidebar_settings:
-      shared:
-        Showing or hiding the sidebar:
+      header: Sidebar settings
```
leenagupte: It looks like this is nested under "secondary_content_links". Should this be under sidebar_settings on line 67?

config/locales/en.yml
AgaDufrat: I'm removing it in the next commit, when I'm changing the `new_step` part. It got picked up by diff because I accidentally add a space - will fix that.

config/locales/en.yml
leenagupte: Sorry. Now that I'm looking at the whole file, I can see how you've separated the contextual and full page guidance headings.

config/locales/en.yml
```diff
@@ -138,23 +138,7 @@ en:
 
           - could help you complete a task but is not the main piece of content included in the step by step.
           - is optional and not useful enough to be included in the step by step
-    change_notes:
```
leenagupte: Is the change notes guidance no longer being used?

AgaDufrat: The ones I deleted are not used anywhere. 
The views use those ones and they are only used once:
https://github.com/alphagov/collections-publisher/blob/839e95f12a5d00f584f0556615ef5a4ae4d70d2f/config/locales/en.yml#L37-L59

leenagupte: Can you add that to the commit message?

PR #898, opened 2020-01-07
Switch to rubocop-govuk

benthorner:
https://trello.com/c/EcitEsfK/1297-audit-dependabot-for-collections-publisher

This switches to rubocop-govuk, expands the set of directories to
be linted, removes a bunch of custom overrides for this app, and
fixes all the associated linting errors. Please see the commits for
more details.


benthorner:
@AlanGabbianelli FYI

app/helpers/application_helper.rb
```diff
@@ -29,7 +29,7 @@ def published_url(slug)
   end
 
   def markdown_to_html(markdown)
-    raw(Kramdown::Document.new(markdown).to_html)
+    raw(Kramdown::Document.new(markdown).to_html) # rubocop:disable Rails/OutputSafety
```
1pretz1: Could we use `sanitize` here instead of `raw`, seemingly we'd mitigate all the threats associated with `raw` through stripping scripts and other potentially harmful things? - https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html.

I ran the tests with sanitize and they all passed which shows at least to the tests the two methods are interchangeable.

I know this isn't really about linting directly, soz.

benthorner: Good question! 

We have to be careful with `sanitize`, since for database content it could produce different behaviour, since we don't know what the content is. I saw that we use the same approach in Content Publisher, but I thought I'd dig a bit deeper: it should be fine to use `sanitize` in all cases, and I've updated the commit to explain.

1pretz1: Woo 🎉

PR #969, opened 2020-03-24
Rake task to publish empty coronavirus content item

hannako:
Rake task to publish coronavirus_landing_page to the Publishing API. So that we can spike a new publishing workflow. See [trello](https://trello.com/c/0uZVAYYt/35-look-at-options-for-publishers-to-update-page)


lib/tasks/publishing_api.rake
```diff
@@ -2,6 +2,32 @@ require_relative "../publish_organisations_api_route"
...
 namespace :publishing_api do
+  desc "Publish coronavirus_landing_page to publishing api"
+  task publish_coronavirus_landing_page: :environment do
+    content_id = "774cee22-d896-44c1-a611-e3109cce8eae"
+    params = {
+      base_path: "/coronavirus",
+      publishing_app: "collections-publisher",
+      rendering_app: "collections",
+      public_updated_at: Time.zone.now.iso8601,
+      update_type: "major",
```
sihugh: Shall we stick with minor updates for now?

hannako:
@sihugh would it be safer in the first instance to change the url to something different eg `/coronavirus-landing-page` and create a redirect in integration on short url manager to `/coronavirus`?

lib/tasks/publishing_api.rake
```diff
@@ -2,6 +2,32 @@ require_relative "../publish_organisations_api_route"
...
+  desc "Publish coronavirus_landing_page to publishing api"
+  task publish_coronavirus_landing_page: :environment do
+    content_id = "774cee22-d896-44c1-a611-e3109cce8eae"
+    params = {
+      base_path: "/coronavirus",
+      publishing_app: "collections-publisher",
+      rendering_app: "collections",
+      public_updated_at: Time.zone.now.iso8601,
+      update_type: "minor",
+      schema_name: "edition",
```
sihugh: I think should be `document_type: "coronavirus_landing_page`?

sihugh: We don't need this line I think

hannako: thanks removed

hannako: @sihugh Getting `["Schema could not be validated as the schema_name was not provided"]` when running the task on integration https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/132401/console I'll stick it back in!

PR #977, opened 2020-04-01
Makes the interface more user friendly

DilwoarH:
Makes it easier for users to follow the process by splitting everything up into steps

### DIFF
--- a/app/views/coronavirus/index.html.erb
+++ b/app/views/coronavirus/index.html.erb
@@ -1,29 +1,48 @@
 <% content_for :title, "Publish gov.uk/coronavirus content" %>
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-three-quarters">
-    <div>
-      <p class="govuk-body">Updating coronavirus landing page content:</p>
-      <ol class="govuk-list govuk-list--number">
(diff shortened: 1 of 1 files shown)


app/views/coronavirus/index.html.erb
```diff
@@ -46,6 +67,13 @@
           text: "Publish",
           margin_bottom: true
       } %>
+    <% end %>
+
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "6. Check live",
+      padding: true
+    } %>
+    <%= form_with(url: coronavirus_publish_path) do %>
```
sihugh, line 76: I think this is a duplicate line (the good one is on line 48) - the form needs to wrap the radio buttons and the publish button

DilwoarH: we can remove the form I think - It's just a link

app/views/coronavirus/index.html.erb
```diff
@@ -1,29 +1,50 @@
...
           text: "Update draft",
           margin_bottom: true
       } %>
+    <% end %>
+
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "4. Preview changes",
+      padding: true
+    } %>
+    <%= form_with(url: coronavirus_update_draft_path) do %>
```
sihugh, line 35: This form isn't needed - it's just a link too

PR #989, opened 2020-04-06
Add a Rake task to toggle the livestream

edwardkerry:
- We have an "extended hours" rota from 1430 to 2200 every weekday whose
  "extra" job consists of editing three lines of YAML, clicking a button
  in Collections Publisher, then reverting the three lines of YAML and
  clicking the same button, to enable and disable the coronavirus press
  conference live stream. It requires two people: one to raise the PR,
  and one to approve the PR, in each case.
- In an effort to save ourselves some work, here's a Rake task that we
  can run through Jenkins (maybe Content Designers can run it - some of
  them can already run some tasks?).
- This publishes the draft content item. We still have to use the UI
  buttons in Collections Publisher to make the draft LIVE.
- Run with
  `'publishing_api:coronavirus_livestream[true, 32nd April 2020]'`
  to enable for a date (it doesn't check that the date is valid), and
  `'publishing_api:coronavirus_livestream[false]'` to disable.

Co-authored-by: Ed Kerry <edward.kerry@digital.cabinet-office.gov.uk>
Co-authored-by: Issy Long <issy.long@digital.cabinet-office.gov.uk>

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -28,6 +28,27 @@ namespace :publishing_api do
     Services.publishing_api.publish(content_id)
   end
 
+  desc "Toggle the daily coronavirus livestream"
+  task :coronavirus_livestream, %i[enabled? date] => :environment do |_task, args|
+    url = "https://raw.githubusercontent.com/alphagov/govuk-coronavirus-content/master/content/coronavirus_landing_page.yml".freeze
+    response = RestClient.get(url)
+    if response.code == 200
+      content_id = "774cee22-d896-44c1-a611-e3109cce8eae"
+      enabled = args[:enabled?] == "true"
+      corona_content = YAML.safe_load(response.body)["content"]
+      payload = CoronavirusPagePresenter.new(corona_content).payload
+      params = {
+          "date" => args[:date],
+          "show_video" => enabled,
+        }
+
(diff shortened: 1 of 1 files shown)


sihugh:
This is great, thanks.  We'll add a card to bring this into the publishing app so that content people can do the updates.  One risk is that people may change the status of the live stream by publishing unrelated content changes, so we'll have to think about how that works

issyl0:
@sihugh I'm confused. So this isn't mergeable? Or it is, but in a different app? Which publishing app are you talking about, because this is Collections Publisher right here. It's also a little weird that the content is in a separate repo...

sihugh:
> @sihugh I'm confused. So this isn't mergeable? Or it is, but in a different app? Which publishing app are you talking about, because this is Collections Publisher right here. It's also a little weird that the content is in a separate repo...

It's not exactly a mature publishing experience.  We deliberately put the content somewhere else so we don;t accrue a deploy backlog that makes zero difference to the app.  I think that'd be more confusing.  

With this approach, we're at risk of overwriting the livestream config if someone publishes content via the standard route. I think the answer is to persist the livestream config in the app db and remove it entirely from the yaml.  I'll do something in that vein.

issyl0:
Looks like this is superseded by #996.

PR #994, opened 2020-04-09
Merge landing and business page workflows

sihugh:
These do exactly the same thing apart from a slight difference in the validation, so let's merge them.  This makes it easier to reuse for future pages too as we'll just have to create a new section in the all_page_configuration section.

- The landing and business actions have been merged into "show" 
- The update_landing and update_business actions have been merged into "update"
- The publish_landing and publish_business actions have been merged into "publish"

I've swapped the routes to use standard rails resource routes now.  It made sense to start with, but as we're adding functionality, it's good to stay with convention as much as possible.

https://trello.com/c/jH5CnWb5/182-generalise-hub-page-work


spec/support/coronavirus_feature_steps.rb
```diff
@@ -45,25 +45,37 @@ def then_the_business_content_is_sent_to_publishing_api
...
-  click_on("Business page")
+  click_on("Business support page")
 end
 
 def when_i_visit_the_publish_coronavirus_page
   visit "/coronavirus"
 end
 
+def when_i_visit_a_non_existent_page
+  visit "/coronavirus/flimflam"
```
hannako, line 68: 😂

PR #996, opened 2020-04-09
Turn on coronavirus live streams from collections publisher

hannako:
## What
- first attempt at adding a button to the coronavirus pages tab in collections publisher so that content designers/others can activate the livestream without dev input. 
## How
- Set the livestream state (on or off) in collections publisher instead of in the [yaml](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml) file in the govuk-coronavirus-content repo
- Persist live stream state in database
- Read from database when creating payload for publishing api

[Trello card](https://trello.com/c/tsal0J9i/158-create-a-way-for-content-designers-to-switch-the-livestream-on-and-off)

### DIFF
--- a/app/controllers/coronavirus_controller.rb
+++ b/app/controllers/coronavirus_controller.rb
@@ -1,3 +1,5 @@
+require_relative "../services/live_stream_updater.rb"
+
 class CoronavirusController < ApplicationController
   before_action :require_coronavirus_editor_permissions!
   layout "admin_layout"
@@ -11,6 +13,27 @@ class CoronavirusController < ApplicationController
     end
 
     render :index, locals: { links: links }
+  end
+
+  def live_stream
+    live_stream = LiveStream.first_or_create
+    @live_stream = LiveStreamUpdater.new(live_stream).resync
+  end
+
+  def publish_live_stream
+    @live_stream = LiveStream.last
+    updater = LiveStreamUpdater.new(@live_stream, params[:on])
+    if updater.updated?
+      if updater.published?
+        flash[:notice] = "Live stream turned #{@live_stream.state ? 'on' : 'off'}"
+        redirect_to coronavirus_live_stream_path
+      else
+        flash["alert"] = "Live stream has not been updated - please try again"
+      end
+    else
+      flash["alert"] = "Content item has not been updated - please try again"
+      redirect_to coronavirus_live_stream_path
+    end
   end
 
   def show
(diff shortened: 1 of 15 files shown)


spec/services/live_stream_updater_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(LiveStream.last.state).to be true
+    end
+  end
+
+  context "Unsuccesful interaction with publishing api" do
+
+    let(:updater) { described_class.new( live_stream, state_params_on ) }
+
+    before :each do
+      stub_any_publishing_api_call_to_return_not_found
```
sihugh, line 39: This should raise a `GdsApi::HTTPNotFound` error which I don't think we're catching yet. The full list is here: https://github.com/alphagov/gds-api-adapters/blob/bd031098cbfcee4934ad0648a9b9e29793e6f776/lib/gds_api/exceptions.rb#L62-L93

app/services/live_stream_updater.rb
```diff
@@ -0,0 +1,95 @@
...
+    presenter.payload.merge(
+      {
+        "title" => "Coronavirus (COVID-19): what you need to do",
+        "description" => content_item["description"],
+      },
+    )
+  end
+
+  def landing_page_id
+    "774cee22-d896-44c1-a611-e3109cce8eae"
```
sihugh, line 82: This feels like a constant that we could then maybe reuse in the tests.  (I think it's defined somewhere else already too)

app/controllers/coronavirus_controller.rb
```diff
@@ -1,3 +1,5 @@
+require_relative "../services/live_stream_updater.rb"
```
sihugh, line 1: Do we need this?

app/services/live_stream_updater.rb
```diff
@@ -0,0 +1,95 @@
...
+      update_content_item
+    end
+  end
+
+  def update_content_item
+    with_longer_timeout do
+      begin
+        response =
+          Services.publishing_api.put_content(landing_page_id, live_stream_payload)
+        response.code == 200 ? response : object.toggle(:state)
```
sihugh, line 44: I think this line might confuse me, so I might have a think about how I might unconfuse myself

app/services/live_stream_updater.rb
```diff
@@ -0,0 +1,95 @@
...
+      rescue GdsApi::HTTPClientError
```
sihugh, line 45: If we get a HTTPServerError we'll need to toggle back too.

sihugh: HTTPErrorResponse might be the right one

app/services/live_stream_updater.rb
hannako: Yes it's the big has at the bottom of the controller. Maybe instead of a hash there should be a module that contains the data.

PR #1000, opened 2020-04-15
Uniqueness validator explicitly uses case sensitivity

hannako:
Fixes Rail 6.1 deprecation warnings

```
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :slug attribute in StepByStepPage model, pass `case_sensitive: true` option explicitly to the uniqueness validator

DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :slug attribute in Tag model, pass `case_sensitive: true` option explicitly to the uniqueness validator

DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :base_path attribute in SecondaryContentLink model, pass `case_sensitive: true` option explicitly to the uniqueness validator

```

### DIFF
--- a/app/models/secondary_content_link.rb
+++ b/app/models/secondary_content_link.rb
@@ -3,7 +3,7 @@ class SecondaryContentLink < ApplicationRecord
   validates_presence_of :step_by_step_page
 
   validates :title, :base_path, :content_id, :publishing_app, :schema_name, :step_by_step_page_id, presence: true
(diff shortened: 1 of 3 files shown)


hannako:
@benthorner Thanks Ben, will set as false and merge.

PR #1005, opened 2020-04-20
Fix a couple of coronavirus issues 

sihugh:
Fixes: 
- [The livestream toggle page](https://collections-publisher.publishing.service.gov.uk/coronavirus/live_stream) "Check live" link that is not currently environment specific
- [The business support page](https://collections-publisher.publishing.service.gov.uk/coronavirus/business) has a "View live version" link that points to the /coronavirus page


sihugh:
@1pretz1 thanks for the comment - I've updated accordingly 👍

PR #1016, opened 2020-04-27
Live stream url updater

hannako:
[trello](https://trello.com/c/PdKUXgfr/216-update-the-landing-page-publishing-tool-to-allow-content-designers-to-add-a-specific-youtube-url)

Screenshot 2020-04-28 at 13 43 12

Once this is merged we can remove some keys from the [coronavirus landing page yaml](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml)

https://github.com/alphagov/govuk-coronavirus-content/blob/5142f9c856d4e798a68fbcab2bfa40d94245218a/content/coronavirus_landing_page.yml#L236-L237

https://github.com/alphagov/govuk-coronavirus-content/blob/5142f9c856d4e798a68fbcab2bfa40d94245218a/content/coronavirus_landing_page.yml#L219-L220 we should probably leave these in but set the values to nil

### DIFF
--- a/app/services/live_stream_updater.rb
+++ b/app/services/live_stream_updater.rb
@@ -1,49 +1,41 @@
 class LiveStreamUpdater
-  def initialize(object, state = nil)
-    @object = object
-    @state = state
+  attr_reader :object
+
+  def initialize
     @content_item = fetch_live_content_item
+    @url = live_url
+    @object = live_stream_object
   end
 
-  def updated?
-    update_object_and_content_item.try(:code) == 200
+  def update?
+    update_content_item.try(:code) == 200
   end
 
-  def published?
+  def publish?
     publish_content_item.try(:code) == 200
-  end
-
-  def resync
-    states_in_sync? ? object : object.toggle(:state)
   end
 
 private
 
-  attr_reader :state, :live_stream, :content_item
-  attr_accessor :object
+  attr_reader :live_stream, :content_item, :url
+  attr_writer :object
 
-  def states_in_sync?
-    live_state == object.state
+  def live_stream_object
+    LiveStream.first_or_create(url: url)
   end
 
-  def live_state
-    content_item["details"]["live_stream_enabled"]
-  end
-
-  def update_object_and_content_item
-    if object.update(state: state)
-      update_content_item
+  def live_url
+    if content_item.has_key?("details")
+      content_item["details"]["live_stream"]["video_url"]
     end
   end
 
   def update_content_item
     with_longer_timeout do
       begin
-        response =
-          Services.publishing_api.put_content(landing_page_id, live_stream_payload)
-        response.code == 200 ? response : object.toggle(:state)
+        Services.publishing_api.put_content(landing_page_id, live_stream_payload)
       rescue GdsApi::HTTPErrorResponse
-        object.toggle(:state)
+        object.update(url: url)
       end
     end
   end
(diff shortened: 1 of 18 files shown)


app/views/coronavirus/live_stream.html.erb
```diff
@@ -1,25 +1,49 @@
...
   <div class="govuk-grid-column-three-quarters">
-    <!-- turn livestream on/off buttons -->
-    <div>
-      <% if @live_stream.state %>
-        <%= link_to "Turn it off", coronavirus_publish_live_stream_path(on: false), method: :post, class: "govuk-button govuk-button--warning" %>
-      <% else %>
-        <%= link_to "Turn it on", coronavirus_publish_live_stream_path(on: true), method: :post, class: "govuk-button" %>
+    <div class="govuk-form-group">
+      <%= form_with(url: "update_live_stream", method: "post", local: true) do %>
+        <%= label_tag(:url, "URL", class: "govuk-label" ) %>
```
issyl0, line 8:
```suggestion
        <%= label_tag(:url, "YouTube livestream link", class: "govuk-label" ) %>
```

issyl0: I don't know where on the page we should put this, but thinking for if I were an uninitiated content designer, I'd like some more explanations of what this is and where to find the livestream link for today. A floating Google doc isn't necessarily the easiest thing to find!

config/routes.rb
```diff
@@ -2,8 +2,10 @@
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 
   root to: redirect("/step-by-step-pages", status: 302)
-  post "/coronavirus/publish_live_stream", to: "coronavirus#publish_live_stream"
   get "/coronavirus/live_stream", to: "coronavirus#live_stream"
+  post "/coronavirus/update_live_stream", to: "coronavirus#update_live_stream"
```
sihugh, line 6: Convention might be for this to be a `put`? (See the `rake routes` output for the coronavirus resources below)

app/services/live_stream_updater.rb
```diff
@@ -1,68 +1,62 @@
 class LiveStreamUpdater
-  def initialize(object, state = nil)
-    @object = object
-    @state = state
+  attr_reader :object
+
+  def initialize
     @content_item = fetch_live_content_item
```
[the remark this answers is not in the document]
hannako: Thanks, will do!

app/views/coronavirus/live_stream.html.erb
hannako: Yes totally agree with comments. 
- We discussed whether there should be checks on the validity of the youtube link, ie is it actually live, is it from the right user etc, and the decision was that it's out of scope. Ie the links are being added by humans and they can make sure it's taken from the right place. However there is a preview button to make sure that the link works as expected.

- I've been worrying about the publishing of another person's changes unexpectedly generally. Ie at the moment there seems to be multiple content designers making changes to copy and raising pr's at the same time. I _think_ the reason it's been ok is because the content PR's are only merged once they are happy to be made live, so if someone accidentally publishes the changes it's not an issue. But maybe we should make that a key part of the process, ie explicitly state that you should only merge a change to the yaml files if you are happy for the change to be live. 
- I thought I could add a progressive disclosure to the livestream partial with guidance on where to get the link from. I'll add it and re-request your reviews so you can see what you think.

app/services/live_stream_updater.rb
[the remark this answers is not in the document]
hannako: Thanks!

config/routes.rb
```diff
@@ -2,6 +2,8 @@
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 
   root to: redirect("/step-by-step-pages", status: 302)
+  post "/coronavirus/publish_live_stream", to: "coronavirus#publish_live_stream"
```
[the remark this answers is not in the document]
hannako: You mean why aren't they there anymore? I removed them in the hide button pr

config/routes.rb
hannako: thanks

app/services/live_stream_updater.rb
```diff
@@ -45,29 +44,30 @@ def publish_content_item
       begin
         Services.publishing_api.publish(landing_page_id, "minor")
       rescue GdsApi::HTTPErrorResponse
-        object.update(url: live_url)
```
hannako: There was no point in this line, as the content item has already been updated. ie the object and the content item are in sync.

PR #1027, opened 2020-05-12
Update coronavirus landing page publishing tool

hannako:
We have [changed the structure](https://github.com/alphagov/govuk-coronavirus-content/pull/219) of the content item. The publishing tool needs to know about it.

### DIFF
--- a/spec/components/markdown_editor_spec.rb
+++ b/spec/components/markdown_editor_spec.rb
@@ -8,13 +8,9 @@ RSpec.describe "Markdown editor", type: :view do
   end
 
   it "renders a textarea with a label and toolbar" do
-    render "components/markdown_editor",  label: {
-                                            text: "Body",
-                                          },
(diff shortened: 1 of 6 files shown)


spec/components/markdown_editor_spec.rb
```diff
@@ -9,8 +9,8 @@
 
   it "renders a textarea with a label and toolbar" do
     render "components/markdown_editor",  label: {
-                                            text: "Body",
```
DilwoarH, line 12: was this intended?

hannako: Maybe the person who added it thought it looked better? But rubocop was complaining so I removed it. 🤷

hannako: Ah I see sorry - yeah, have tidied that up! Thanks 👍

PR #1045, opened 2020-06-02
Add the worker support hub page

sihugh:
The paths here should match the file in https://github.com/alphagov/govuk-coronavirus-content/pull/275

https://trello.com/c/Pd19z0rM/326-add-employee-hub-configuration-to-collections-publisher

### DIFF
--- a/app/controllers/coronavirus_controller.rb
+++ b/app/controllers/coronavirus_controller.rb
@@ -202,6 +202,14 @@ private
           base_path: "/coronavirus/education-and-childcare",
           github_url: "https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_education_page.yml",
         },
+      employees:
+        {
+          name: "Worker page",
+          content_id: "5ebf285a-9165-476c-be90-66b9729f50da".freeze,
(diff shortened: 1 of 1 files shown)


app/controllers/coronavirus_controller.rb
```diff
@@ -202,6 +202,14 @@ def all_pages_configuration
...
           github_url: "https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_education_page.yml",
         },
+      employees:
+        {
+          name: "Worker page",
+          content_id: "5ebf285a-9165-476c-be90-66b9729f50da".freeze,
+          raw_content_url: "https://raw.githubusercontent.com/alphagov/govuk-coronavirus-content/master/content/coronavirus_worker_page.yml".freeze,
+          base_path: "/coronavirus/worker-support",
+          github_url: "https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_worker_page.yml",
+        },
```
reggieb, line 212: Rather than freezing the individual strings, would it be simpler just to add `# frozen_string_literal: true` to the start of the file. Also should the hash be frozen?

sihugh: Weird that rubocop doesn't care about that - maybe because it's not assigned to anything 🤔

PR #1046, opened 2020-06-02
Livestream can be updated with new signon permission

hannako:
## What
Allow users with a new signon permission `Livestream editor` access to edit the daily livestream URL.

## Why
So that we can hand over control of updating the livestream URL to Cabinet Office/number 10.

## Before PR
Currently to edit the livestream you need both GDS editor and Coronavirus editor permissions:
- GDS Editor permissions required to access the landing page of Collections Publisher (root is step by step index page). 
- From landing page, navigate to Coronavirus Pages, which requires Coronavirus Editor permissions. 

## After PR
- If user has only Livestream Editor permissions, accessing the app they will now be routed to the livestream edit page. 
- This is the only page they can access.
- Coronavirus editors accessing the app will see no change in their flow.

[trello](https://trello.com/c/sRpPbpqM/323-enable-co-no10-to-update-the-livestream)

### DIFF
--- a/app/controllers/live_stream_controller.rb
+++ b/app/controllers/live_stream_controller.rb
@@ -1,0 +1,45 @@
+class LiveStreamController < ApplicationController
+  before_action :require_coronavirus_editor_permissions!
+  layout "admin_layout"
+
+  def index
+    @live_stream = updater.object
+  end
+
+  def update
+    @live_stream = LiveStream.last
+    if @live_stream.update(url: url_params, formatted_stream_date: formatted_date)
+      if updater.update
+        flash[:notice] = "Draft live stream url updated!"
+      else
+        flash["alert"] = "Live stream url has not been updated - please try again"
+      end
+    else
+      flash[:notice] = @live_stream.errors.full_messages.join(", ")
+    end
+    redirect_to live_stream_index_path
+  end
+
+  def publish
+    if updater.publish
+      flash[:notice] = "New live stream url published!"
+    else
+      flash["alert"] = "Live stream url has not been published - please try again"
+    end
+    redirect_to live_stream_index_path
+  end
+
+private
+
+  def updater
+    LiveStreamUpdater.new
+  end
+
+  def url_params
+    params[:url]
+  end
+
+  def formatted_date
+    DateTime.now.strftime("%-d %B %Y")
+  end
+end
--- a/app/controllers/live_stream_controller.rb
+++ b/app/controllers/live_stream_controller.rb
@@ -12,10 +12,10 @@ class LiveStreamController < ApplicationController
       if updater.update
         flash[:notice] = "Draft live stream url updated!"
       else
-        flash["alert"] = "Live stream url has not been updated - please try again"
+        flash[:alert] = "Live stream url has not been updated - please try again"
       end
     else
-      flash[:notice] = @live_stream.errors.full_messages.join(", ")
+      flash[:alert] = @live_stream.errors.full_messages.join(", ")
     end
     redirect_to live_stream_index_path
   end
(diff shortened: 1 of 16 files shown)


app/views/layouts/application.html.erb
```diff
@@ -13,9 +13,11 @@
     </li>
   <% end %>
 
-  <li class='<%= active_navigation_item == 'topics' ? 'active' : nil %>'>
-    <%= link_to 'Specialist sector pages', topics_path %>
-  </li>
+  <% unless livestream_editor? %>
+    <li class='<%= active_navigation_item == 'topics' ? 'active' : nil %>'>
```
reggieb, line 17: HTML tag attributes should be double quote I think. So:
```
<li class="<%= active_navigation_item == 'topics' ? 'active' : nil %>">
```

spec/support/coronavirus_feature_steps.rb
```diff
@@ -4,26 +4,40 @@ def given_i_am_a_coronavirus_editor
...
+  stub_publishing_api_has_item(JSON.parse(live_coronavirus_content_item))
+end
+
+def live_coronavirus_content_item
+  File.read(Rails.root.join + "spec/fixtures/coronavirus_content_item.json")
+end
+
+def coronavirus_live_stream_hash
+  h = JSON.parse(live_coronavirus_content_item)
+  h["details"]["live_stream"]
```
reggieb, line 35: Would dig be a good option here? It would stop an error being raised if the "details" key is missing.

spec/support/coronavirus_feature_steps.rb
```diff
@@ -4,26 +4,40 @@ def given_i_am_a_coronavirus_editor
...
+end
+
+def coronavirus_content_id
+  h = JSON.parse(live_coronavirus_content_item)
```
reggieb, line 39: That the third repetition of `JSON.parse(live_coronavirus_content_item)`. I'd suggest
```ruby
def coronavirus_content_json
  @coronavirus_content_json ||= JSON.parse(live_coronavirus_content_item)
end
```
And then update the three instances to use that method. So here:
```ruby
def coronavirus_content_id
  coronavirus_content_json["content_id"]
end
```

spec/support/livestream_helpers.rb
```diff
@@ -0,0 +1,21 @@
+def stub_todays_youtube_link
+  stub_request(:get, todays_link)
+end
+
+def stub_yesterdays_youtube_link
+  stub_request(:get, yesterdays_link)
+end
+
+def yesterdays_link
+  h = JSON.parse(live_coronavirus_content_item)
```
reggieb, line 10: Same pattern again - Looks like `coronavirus_content_json` would be useful here too.

app/controllers/live_stream_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    @live_stream = updater.object
+  end
+
+  def update
+    @live_stream = LiveStream.last
+    if @live_stream.update(url: url_params, formatted_stream_date: formatted_date)
+      if updater.update
+        flash[:notice] = "Draft live stream url updated!"
+      else
+        flash["alert"] = "Live stream url has not been updated - please try again"
```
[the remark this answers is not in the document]
hannako: Thanks

app/controllers/live_stream_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+    else
+      flash[:notice] = @live_stream.errors.full_messages.join(", ")
```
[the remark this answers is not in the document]
hannako: yes it should, thanks for spotting that 👍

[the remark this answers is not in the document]
reggieb: Ummmm. It's still showing as a notice.

PR #1053, opened 2020-06-10
Coronavirus Publishing Tool - accordions are persisted and can be created and updated

hannako:
pair_programming effort with @reggieb 

## What
This PR is concerned with fetching and persisting the accordions (aka sections) on the coronavirus landing page and hub pages. 

## How

1. Move the current coronavirus publishing page to `coronavirus/slug/prepare`
2. Wire up the [new show page](https://github.com/alphagov/collections-publisher/pull/1056) with the route `coronavirus/slug` 
3. The coronavirus content items have a key of `sections` with a value that is an array of `sub_sections`. Each `sub_section` has a `title`, and a `list` of links. We've used the same terminology for the database objects:
     - CoronavirusPage has a `section_title`
     -  SubSection has a `title` and `content` (as markdown). The content field stores the list of links and subtitles if present. 
4. The first time either the prepare or show page is visited, we fetch the relevant yaml from govuk-coronavirus-content and create a CoronavirusPage and its associated SubSections. 
5. Thereafter we read from the database. 
6. SubSections can be created and updated. 

Future work is needed to do the following, but is split into separate PR's because this one is already a bit of a beast:
- delete a section
- push to publishing api

[Trello](https://trello.com/c/lrCwu2Qk/356-story-1a-basic-editing-of-coronavirus-accordions-within-collections-publisher)


app/services/yaml_fetcher.rb
```diff
@@ -15,9 +15,7 @@ def success?
     false
   end
 
-  def body
-    response.body
-  end
+  delegate :body, to: :response
```
hannako, line 18: rubocop!

PR #1063, opened 2020-06-18
Coronavirus publishing tool: Reordering subsections

danacotoran:
Pair programming effort @danacotoran and @hannako 

## What 
Implements reordering of subsections (aka accordions)

## How
**Frontend** 
- Implements accordion reordering following the existing [pattern established in step-by-step pages]( https://collections-publisher.integration.publishing.service.gov.uk/step-by-step-pages/87/reorder) 

- On the accordion reordering page [there is a form](https://github.com/alphagov/collections-publisher/pull/1063/files#diff-c9b3ce154ac76d4a02cb44e9c925ca34) with a `<input type="hidden" name="section_order_save" id="js_order_save" />` On page load (and whenever the items under `<ul id="js-reorder-group">` are moved up/down), the value of this hidden input is dynamically updated to reflect the positions of the items. 

- The value will look something like this `value="[{"id":652,"position":1},{"id":653,"position":2},{"id":654,"position":3}]"`

- When the user hits the Save button, this form data is `POST`ed to the form endpoint.

**Backend**
- Adds a position attribute to the SubSection model
- When we fetch the subsection data from github and create the model, we set the position based on its position in the source file.
- When we display and send the subsections back to publishing api, we do so in order of the subsection position.
- I've implemented rolling back of the new positions if the call to publishing api fails. I don't know if this is necessarily a good thing, it might be annoying to lose your work. It is easy to remove. 

Trello: https://trello.com/c/RfnRjvr2/362-story-5-accordions-can-be-reordered-from-accordion-publishing-tool

### DIFF
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -37,23 +37,33 @@ class CoronavirusPagesController < ApplicationController
 
   def reorder
     coronavirus_page
+    @old_positions = positions
     if request.post? && params.key?(:section_order_save)
-      reordered_sections = JSON.parse(params[:section_order_save])
-      reordered_sections.each do |section_data|
-        sub_section = @coronavirus_page.sub_sections.find(section_data["id"])
-        sub_section.update(position: section_data["position"])
+      @new_positions = JSON.parse(params[:section_order_save])
+      set_positions(@new_positions)
+      if draft_updater.send
+        message = { notice: "Sections were successfully reordered." }
+      else
+        set_positions(@old_positions)
+        message = { alert: "Sorry! Sections have not been reordered: " + draft_updater.errors.to_sentence }
       end
-      message = if draft_updater.send
-                  { notice: "Steps were successfully reordered." }
-                else
-                  { alert: draft_updater.errors.to_sentence }
-                  # #undo_order_change
-                end
       redirect_to coronavirus_page_path(slug), message
     end
   end
 
 private
+
+  def positions
+    coronavirus_page.sub_sections.each_with_object([]) do |sub_section, array|
+      array << { "id" => sub_section.id, "position" => sub_section.position }
+    end
+  end
+
+  def set_positions(positions)
+    positions.each do |sub_section|
+      SubSection.find(sub_section["id"]).update(position: sub_section["position"])
+    end
+  end
 
   def initialise_coronavirus_pages
     page_configs.keys.map do |page|
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -37,6 +37,20 @@ class CoronavirusPagesController < ApplicationController
 
   def reorder
     coronavirus_page
+    if request.post? && params.key?(:section_order_save)
+      reordered_sections = JSON.parse(params[:section_order_save])
+      reordered_sections.each do |section_data|
+        sub_section = @coronavirus_page.sub_sections.find(section_data["id"])
+        sub_section.update(position: section_data["position"])
+      end
+      message = if draft_updater.send
+                  { notice: "Steps were successfully reordered." }
+                else
+                  { alert: draft_updater.errors.to_sentence }
+                  # #undo_order_change
+                end
+      redirect_to coronavirus_page_path(slug), message
+    end
   end
 
 private
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -1,6 +1,6 @@
 class CoronavirusPagesController < ApplicationController
   before_action :require_coronavirus_editor_permissions!
-  before_action :require_unreleased_feature_permissions!, only: %w[show reorder]
+  before_action :require_unreleased_feature_permissions!, only: %w[show]
   before_action :redirect_to_index_if_slug_unknown, only: %w[prepare show]
   before_action :initialise_coronavirus_pages, only: %w[index]
   layout "admin_layout"
@@ -35,35 +35,7 @@ class CoronavirusPagesController < ApplicationController
     end
   end
 
-  def reorder
-    coronavirus_page
-    @old_positions = positions
-    if request.post? && params.key?(:section_order_save)
-      @new_positions = JSON.parse(params[:section_order_save])
-      set_positions(@new_positions)
-      if draft_updater.send
-        message = { notice: "Sections were successfully reordered." }
-      else
-        set_positions(@old_positions)
-        message = { alert: "Sorry! Sections have not been reordered: " + draft_updater.errors.to_sentence }
-      end
-      redirect_to coronavirus_page_path(slug), message
-    end
-  end
-
 private
-
-  def positions
-    coronavirus_page.sub_sections.each_with_object([]) do |sub_section, array|
-      array << { "id" => sub_section.id, "position" => sub_section.position }
-    end
-  end
-
-  def set_positions(positions)
-    positions.each do |sub_section|
-      SubSection.find(sub_section["id"]).update(position: sub_section["position"])
-    end
-  end
 
   def initialise_coronavirus_pages
     page_configs.keys.map do |page|
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -1,6 +1,6 @@
 class CoronavirusPagesController < ApplicationController
   before_action :require_coronavirus_editor_permissions!
-  before_action :require_unreleased_feature_permissions!, only: %w[show]
+  before_action :require_unreleased_feature_permissions!, only: %w[show reorder]
   before_action :redirect_to_index_if_slug_unknown, only: %w[prepare show]
   before_action :initialise_coronavirus_pages, only: %w[index]
   layout "admin_layout"
@@ -33,6 +33,10 @@ class CoronavirusPagesController < ApplicationController
     else
       redirect_to coronavirus_page_path(slug)
     end
+  end
+
+  def reorder
+    coronavirus_page
   end
 
 private
--- a/app/views/coronavirus_pages/_sub_sections_summary_list.html.erb
+++ b/app/views/coronavirus_pages/_sub_sections_summary_list.html.erb
@@ -1,8 +1,9 @@
-
+<% reorder ||= false %>
 <% sub_sections =
  @coronavirus_page.sub_sections.map do |sub_section|
    {
-     field: sub_section.title ,
+     id: rand(10000),
+     field: sub_section.title,
      edit: { href: edit_coronavirus_page_sub_section_path(@coronavirus_page.slug, sub_section) },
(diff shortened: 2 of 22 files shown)


app/controllers/coronavirus_pages_controller.rb
```diff
@@ -35,8 +35,36 @@ def publish
...
+  def reorder
+    coronavirus_page
+    @old_positions = positions
+    if request.post? && params.key?(:section_order_save)
+      @new_positions = JSON.parse(params[:section_order_save])
+      set_positions(@new_positions)
+      if draft_updater.send
+        message = { notice: "Sections were successfully reordered." }
+      else
+        set_positions(@old_positions)
```
hannako, line 47: As per my PR comment, not sure if we would want to roll back? You could just show an error asking the user to click save again.

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -35,8 +35,36 @@ def publish
     end
   end
 
+  def reorder
+    coronavirus_page
+    @old_positions = positions
+    if request.post? && params.key?(:section_order_save)
```
reggieb, line 41: Checking if the request is a `POST` and having one action handle both the edit and the update actions is not a standard pattern for rails. I think this could be improved by splitting this into two actions. For a start it would remove the need for the `if` statement completely. I'd even suggest that this could be in a separate reorder controller, which would help in keeping down the size of the `CoronavirusPagesController`.

The `reorder` action (or `CoronavirusPageReordersController` `show` action) could just be:
```ruby
def reorder
  coronavirus
end
```
Then an `update_reorder` action could contain lines 40 + 42 to 51, and the methods called.

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -35,8 +35,36 @@ def publish
...
+      redirect_to coronavirus_page_path(slug), message
+    end
+  end
+
 private
 
+  def positions
+    coronavirus_page.sub_sections.each_with_object([]) do |sub_section, array|
+      array << { "id" => sub_section.id, "position" => sub_section.position }
+    end
```
reggieb, line 59: I think `map` is a better method to use here. Also, if the positions are memoized, the instance variable `@old_positions` becomes redundant. So
```ruby
def positions
  @positions ||= coronavirus_page.sub_sections.map do |sub_section|
    { "id" => sub_section.id, "position" => sub_section.position }
  end
end
```
The method could be named `old_positions` or `current_positions` to prevent confusion with `positions` being used elsewhere for other collections.

app/views/coronavirus_pages/_sub_sections_summary_list.html.erb
```diff
@@ -13,9 +14,28 @@
...
-<div class="covid-manage-page__accordion-summary-list">
-  <%= render "govuk_publishing_components/components/summary_list", {
-    title: @coronavirus_page.sections_title,
-    items: sub_sections
-  } %>
+
+<% if reorder  %>
+  <div class="covid-manage-page__accordion-summary-list">
+    <ul class="govuk-list step-by-step-reorder__list" id="js-reorder-group">
+      <% sub_sections.each_with_index do |accordion, index| %>
```
reggieb, line 21: `accordion` is a display choice and not a description of the object. I think it would be clearer if `sub_section` was used here. As that way there will be consistency in the way an instance of `SubSection` is labelled.

Also each `sub_section` now has a position, so we don't need to create an index to uniquely identify each subsection. So I think this line should become:
```irb
<% sub_sections.each do |sub_section| %>
```
and line 23 would become:
```irb
<div class="govuk-grid-row" id="step-<%= sub_section['position'] %>">
```

app/views/coronavirus_pages/_sub_sections_summary_list.html.erb
```diff
@@ -1,8 +1,9 @@
-
+<% reorder ||= false %>
```
reggieb, line 1: I think there should be a separate partial for the reorder display, rather than reusing this partial.

app/views/coronavirus_pages/reorder.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+%>
+
+<% content_for :title, "Coronavirus (COVID-19)" %>
+<% content_for :context, "Reorder #{@coronavirus_page.sections_title.downcase} accordion" %>
+
+<div class="covid-manage-page govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render_markdown "Use the up/down buttons on the right to reorder the sections, or click and hold on a section to reorder using drag and drop." %>
+    <%= render "coronavirus_pages/sub_sections_summary_list", { reorder: true } %>
```
reggieb, line 20: If there was separate partial for the reorder view, this could become something like:
```irb
<%= render "coronavirus_pages/sub_sections_reorder_list" %>
```

config/routes.rb
```diff
@@ -20,6 +20,8 @@ def matches?(request)
 
   resources :coronavirus_pages, path: "coronavirus", only: %i[index show update], param: :slug do
     get "prepare", on: :member
+    get "reorder", on: :member
+    post "reorder", on: :member
```
reggieb, line 24: As stated previously - `post` and `get` should not point at the same action as that leads to the action having to be far more complicated than it would be if it only had to handle a single action.

spec/controllers/coronavirus_pages_controller_spec.rb
```diff
@@ -82,4 +83,48 @@
...
+      stub_user.permissions << "Unreleased feature"
+      stub_request(:get, coronavirus_page.raw_content_url)
+        .to_return(status: 200, body: raw_content)
+      stub_coronavirus_publishing_api
+      live_stream
+    end
+    let!(:sub_section_0) { create :sub_section, position: 0, coronavirus_page: coronavirus_page }
+    let!(:sub_section_1) { create :sub_section, position: 1, coronavirus_page: coronavirus_page }
+
+    let(:section_params) { "[{\"id\":#{sub_section_0.id}, \"position\":1},{\"id\":#{sub_section_1.id},\"position\":0}]" }
```
reggieb, line 109: I think if `to_json` is used on an array, the construction of this data becomes a little simpler and easier to manipulate. For example, if `section_params` was built like this:
```ruby
let(:sub_section_0_params) { { id:  sub_section_0.id, position: 1 }  }
let(:sub_section_1_params) { { id:  sub_section_0.id, position: 0 }   }
let(:section_params) { [ sub_section_0_params, sub_section_1_params ].to_json }
```
an additional context could be added that clarified that it was the `position` data that was setting the position and that reorder wasn't simply doing a reverse sort.
``` ruby
context "with positions reversed" do
  let(:sub_section_0_params) { { id:  sub_section_0.id, position: 0 }  }
  let(:sub_section_1_params) { { id:  sub_section_0.id, position: 1 }   }

  it "reverses the positions" do
    subject
    expect(sub_section_0.reload.position).to eq 0
    expect(sub_section_1.reload.position).to eq 1
  end
end
```

spec/factories.rb
```diff
@@ -245,7 +245,8 @@
 
   factory :sub_section do
     title { Faker::Lorem.sentence }
-    content { "(#{Faker::Lorem.sentence})[/#{File.join(Faker::Lorem.words)}]" }
+    content { "[#{Faker::Lorem.sentence}](/#{File.join(Faker::Lorem.words)})" }
+    position { rand(4) }
```
reggieb, line 249: Would there be a problem if two sub_sections had the same position. If not, this is fine. If it would be problematic, I wonder if something like this would be better:
```ruby
position {  (SubSection.maximum(:position) || 0) + 1 }
```

spec/services/coronavirus_pages/model_builder_spec.rb
```diff
@@ -46,6 +46,26 @@
...
+
+        # {0=>"How to protect yourself and others",
+        #  1=>"Employment and financial support",
+        #  2=>"School closures, education, and childcare",
+        #  3=>"Businesses and other organisations",
+        #  4=>"Healthcare workers and carers",
+        #  5=>"Travel",
+        #  6=>"How coronavirus is affecting public services",
+        #  7=>"How you can help",
+        #  8=>"Coronavirus (COVID-19) cases in the UK"}
```
reggieb, line 62: I'm not sure these comments are useful (the actual content of the comments isn't important, and if someone wants to look it up, it would be better to look at the source yaml document), and they will be difficult to maintain. I'd suggest removing this comment.

spec/support/coronavirus_feature_steps.rb
```diff
@@ -174,6 +174,79 @@ def when_i_visit_a_coronavirus_page
...
 end
 
+### Reordering sections spec ##
+
+def when_i_visit_the_reorder_page
+  visit "/coronavirus/landing/reorder"
+end
+
+def set_up_basic_sub_sections
+  c = FactoryBot.create(:coronavirus_page, :landing)
```
reggieb, line 184: Use descriptive parameter names on multi-line statements. So:
```ruby
coronavirus_page = FactoryBot.create(:coronavirus_page, :landing)
```

spec/support/coronavirus_feature_steps.rb
```diff
@@ -174,6 +174,79 @@ def when_i_visit_a_coronavirus_page
...
+                    coronavirus_page_id: c.id,
+                    position: 0,
+                    title: "I am first",
+                    content: "###title\n[label](/url)")
+  FactoryBot.create(:sub_section,
+                    coronavirus_page_id: c.id,
+                    position: 1,
+                    title: "I am second",
+                    content: "###title\n[label](/url)")
+  path = Rails.root.join + "spec/fixtures/simple_coronavirus_page.yml"
```
reggieb, line 195: `Rails.root.join` does not require a `+` operator. So:
```ruby
path = Rails.root.join "spec/fixtures/simple_coronavirus_page.yml"
```

spec/support/coronavirus_feature_steps.rb
```diff
@@ -174,6 +174,79 @@ def when_i_visit_a_coronavirus_page
...
+  github_yaml_content = File.read(path)
+  stub_request(:get, c.raw_content_url)
+    .to_return(status: 200, body: github_yaml_content)
+end
+
+def i_see_subsection_one_in_position_one
+  e = find("#step-0").find(".step-by-step-reorder__step-title")
```
reggieb, line 202: Again - a descriptive variable name would be preferable to a single letter one on a multi-line statement.

reggieb:
I decided to request changes before reading the specs. As I wanted to request the changes before someone else approved the PR and it was merged, I submitted my review at that point. I then went through the specs and added my comments on those as single comments. I have now completed my initial review.

app/views/coronavirus_pages/_sub_sections_summary_list.html.erb
reggieb: I tried removing the index and using the sub_section position, but that didn't work. I think the problem is that a sub_section may not have a position initially.

PR #1067, opened 2020-06-22
Add confirmation banner when publishing without 2i

1pretz1:
This adds some guidance around when a publisher should publish without
2i when they hit the "Publish without 2i review" button. The
confirmation pop up follows the same pattern as some of the other
actions like 'delete' and 'revert to draft'.

The bulk of the text was copied from whitehall's 'Force publish' partial
here -> https://github.com/alphagov/whitehall/blob/master/app/views/admin/edition_workflow/_force_publish_modal.html.erb,
with a few words changed as there isn't a 'Force publish' concept in
collections-publisher.

### Screenshot (before content review)
Screenshot 2020-06-22 at 11 53 20

Trello:
https://trello.com/c/NrZVDIlG/2018-3-alert-users-when-a-step-by-step-is-force-published-in-collections-publisher

### DIFF
--- a/app/views/step_by_step_pages/show/_actions.erb
+++ b/app/views/step_by_step_pages/show/_actions.erb
@@ -59,7 +59,8 @@
         <%= render "govuk_publishing_components/components/button", {
           text: "Publish without 2i review",
           href: step_by_step_page_publish_without_2i_review_path(@step_by_step_page),
-          destructive: true
+          destructive: true,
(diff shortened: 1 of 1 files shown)


mattclear:
Hey @1pretz1 - can we tweak this a bit so the wording is closer to what's used in mainstream publisher? Sorry, I should have suggested this earlier.

> You should only publish step by steps without 2i review in exceptional circumstances, for example if you're the only person responding to an emergency out of hours call.

PR #1068, opened 2020-06-22
377 coronavirus publishing tool update draft content item (parsing and JSON build)

reggieb:
Create a presenter `SubSectionJsonPresenter` that takes a SubSection instance and generates a hash (for conversion to JSON) that can be used to publish the data within the sub section.

The presenter uses `PatternMaker` to allow human readable instructions to be converted into Regular Expressions.

Example usage:
```ruby

2.6.6 :006 > s = SubSection.new(title: 'This title', content: "## A Title\n(foo)[path/to/foo]\n(bar)[path/to/bar]")
 => #<SubSection id: nil, title: "This title", content: "## A Title\n(foo)[path/to/foo]\n(bar)[path/to/bar]", coronavirus_page_id: nil, created_at: nil, updated_at: nil> 
2.6.6 :007 > p = SubSectionJsonPresenter.new(s)
 => #<SubSectionJsonPresenter:0x0000563743762a70 @sub_section=#<SubSection id: nil, title: "This title", content: "## A Title\n(foo)[path/to/foo]\n(bar)[path/to/bar]", coronavirus_page_id: nil, created_at: nil, updated_at: nil>> 
2.6.6 :008 > p.output
 => {:details=>{:sections=>[{:title=>"This title", :sub_sections=>[{:title=>"A Title", :list=>[{:label=>"foo", :url=>"path/to/foo"}, {:label=>"bar", :url=>"path/to/bar"}]}]}]}} 
```
https://trello.com/c/zNcOOclL/377-coronavirus-publishing-tool-update-draft-content-item

### DIFF
--- a/app/presenters/sub_section_json_presenter.rb
+++ b/app/presenters/sub_section_json_presenter.rb
@@ -1,8 +1,16 @@
-class SubSectionJsonPresenter
+# frozen_string_literal: true
 
+class SubSectionJsonPresenter
   HEADER_PATTERN = PatternMaker.call(
-    'starts_with hashes then perhaps_spaces then target(words) and nothing_else'.freeze,
-    hashes: '#+'
+    "starts_with hashes then perhaps_spaces then target(title) and nothing_else",
+    hashes: "#+",
+    title: '\w.+',
+  )
+
+  LINK_PATTERN = PatternMaker.call(
+    "starts_with within(brackets,target(label)) then within(sq_brackets,target(url))",
+    label: '\s*\w.+',
+    url: '[\w\/]+',
   )
 
   attr_reader :sub_section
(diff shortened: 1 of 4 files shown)


app/presenters/sub_section_json_presenter.rb
```diff
@@ -0,0 +1,100 @@
...
+            title: title,
+            sub_sections: sub_sections,
+          },
+        ],
+      },
+    }
+  end
+
+  def errors
+    @errors ||= []
```
hannako, line 39: So we basically swallow the error? So if a user enters bad markdown, there's no validation running on that, so it will bomb out when we try to parse it. The result will be that the draft content item will be missing a subsection. That seems ok I think.

hannako: Should we send the errors somewhere? eg Sentry?

reggieb: Yes - that's how it behaves. I think we can add the handling of errors later (on publish perhaps)

PR #1071, opened 2020-06-23
Coronavirus publishing tool: Add some validations to SubSections 

hannako:
### What
Currently we allow invalid subsections to be saved to the database. Adding validations so that all subsections must have title, content, and a parent CoronavirusPage.

### Before
```
s = SubSection.create!
s.valid?
true
```
### After
```
s = SubSection.create!
ActiveRecord::RecordInvalid (Validation failed: Title can't be blank, Content can't be blank, Coronavirus page can't be blank)
```

trello: https://trello.com/c/zNcOOclL/377-coronavirus-publishing-tool-update-draft-content-item

### DIFF
--- a/app/controllers/sub_sections_controller.rb
+++ b/app/controllers/sub_sections_controller.rb
@@ -9,12 +9,11 @@ class SubSectionsController < ApplicationController
 
   def create
     @coronavirus_page = CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
-    sub_section = @coronavirus_page.sub_sections.new(sub_section_params)
-    if sub_section.save
-      redirect_to coronavirus_page_path(@coronavirus_page.slug)
+    @sub_section = @coronavirus_page.sub_sections.new(sub_section_params)
+    if @sub_section.save
+      redirect_to coronavirus_page_path(@coronavirus_page.slug), notice: "Sub-section was successfully created."
     else
-      flash[:alert] = "Not saved!"
-      redirect_to new_coronavirus_page_sub_section_path
+      render :new
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/controllers/sub_sections_controller.rb
```diff
@@ -24,12 +23,12 @@ def edit
   end
 
   def update
-    sub_section = SubSection.find(params[:id])
-    if sub_section.update(sub_section_params)
-      redirect_to coronavirus_page_path(sub_section.coronavirus_page.slug)
+    @sub_section = SubSection.find(params[:id])
+    @coronavirus_page = @sub_section.coronavirus_page
+    if @sub_section.update(sub_section_params)
+      redirect_to coronavirus_page_path(@coronavirus_page.slug), notice: "Step was successfully updated."
```
sihugh, line 29: Should this be `Sub-section` rather than `Step`?

spec/models/coronavirus_page_spec.rb
```diff
@@ -15,4 +15,14 @@
       expect(CoronavirusPage.subtopic_pages).to eq [business, education, employees]
     end
   end
+
+  describe "dependencies" do
+    let!(:employees) { create :coronavirus_page, :employees }
+    let!(:sub_section) { create :sub_section, coronavirus_page: employees }
+
+    it "deletion destroys all child subsections" do
+      employees = CoronavirusPage.find_by(slug: "employees")
```
reggieb, line 24: Is this line needed as `employees` is already available via the `let!(:employees)` declaration on line 29.

spec/models/sub_section_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "rails_helper"
+
+RSpec.describe SubSection do
+  let(:sub_section) { create :sub_section }
+
+  describe "validations" do
+    it "should belong to a coronavirus_page" do
+      should validate_presence_of(:coronavirus_page)
```
reggieb, line 8: Personally, I'm not a great fan of shoulda matchers - they usually shorthand specs for stuff that doesn't really need testing (like joins). However, as they are present, if you want to test `validate :presence` then the whole of this spec could be replaced with:

```ruby
RSpec.describe SubSection, , type: :model do
  it "validates presence" do
    should validate_presence_of(:coronavirus_page)
    should validate_presence_of(:title)
    should validate_presence_of(:content)
  end
end
```

spec/models/coronavirus_page_spec.rb
hannako: Thanks 👍

app/controllers/sub_sections_controller.rb
hannako: Good spot! Anyone would think I was copying existing code!

PR #1076, opened 2020-06-25
Add Notify to collections publisher

1pretz1:
Adds Notify to Collections Publisher and adds an action to send an email when content has been published without 2i review so content teams can keep track of who is publishing without review as this should only happen in exceptional circumstances.

Depends on the below being merged:
- [x] [govuk-secrets](alphagov/govuk-secrets#1000)
- [x] [govuk-puppet](alphagov/govuk-puppet#10457)

### Email sent in integration (before content review)
Screenshot 2020-06-25 at 13 21 36

Trello:
https://trello.com/c/NrZVDIlG/2018-3-alert-users-when-a-step-by-step-is-force-published-in-collections-publisher

### DIFF
--- a/app/mailers/publisher_notifications.rb
+++ b/app/mailers/publisher_notifications.rb
@@ -1,0 +1,15 @@
+class PublisherNotifications < ApplicationMailer
+  include ApplicationHelper
+
+  PUBLISH_WITHOUT_2I_EMAIL = ENV.fetch("PUBLISH_WITHOUT_2I_EMAIL", "test-email-address@gmail.com").freeze
+
+  def publish_without_2i(step_by_step_page, current_user)
+    @current_user = current_user
+    @step_by_step_page = step_by_step_page
+    @live_url = published_url(step_by_step_page.slug)
+
+    view_mail(template_id,
+              to: PUBLISH_WITHOUT_2I_EMAIL,
+              subject: "[PUBLISHER] Step by step published without 2i review for '#{step_by_step_page.title}'")
+  end
+end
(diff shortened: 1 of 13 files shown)


app/mailers/publisher_notifications.rb
```diff
@@ -0,0 +1,16 @@
+class PublisherNotifications < Mail::Notify::Mailer
+  include ApplicationHelper
+
+  TEMPLATE_ID = ENV.fetch("GOVUK_NOTIFY_TEMPLATE_ID", "test-template-id").freeze
```
benjamineskola, line 4: We've implemented this differently in other places that use Notify (as a method on the base class) but I'm not sure it's important enough to change.

app/mailers/publisher_notifications.rb
```diff
@@ -0,0 +1,16 @@
+class PublisherNotifications < Mail::Notify::Mailer
```
benjamineskola, line 1: I think the convention is that `PublisherNotifications` (and all mailers) should inherit from `ApplicationMailer`, and then `ApplicationMailer` can be changed to inherit from `Mail::Notify::Mailer` — but again, not sure if this is important enough to warrant changing.

1pretz1:
I've updated the code to use an ApplicationMailer - I didn't do this at first as it seemed a bit overkill for one mailer action but consistency is king so it has been updated @benjamineskola!

mattclear:
@1pretz1 For this one, I'd suggest:

>Hello,
>
>Peter Hartshorn has published a step by step without 2i review: 'Peters test – publish without 2i review'.
>
>You can view it here:
>https://www.integration.publishing.service.gov.uk/peters-test-test
>
>Sent by the GOV.​UK Collections Publisher application

PR #1077, opened 2020-06-25
Coronavirus publishing tool: Service to update drafts

hannako:
Pair programming effort with @reggieb 

## What
Move the code that updates the content item out of the coronavirus_pages controller and into a new service, so that we can call this service from different controllers without duplication.

## Why 
We are about to implement updates of the coronavirus page content item in the subsections controller, ie after saving a change, or creating a new section we will need to update the whole content item. However we still need to provide a way to update the content item with changes made in github via the `/prepare` action. So it makes sense to create a service to handle this so we can call it in both places. 

trello: https://trello.com/c/zNcOOclL/377-coronavirus-publishing-tool-update-draft-content-item

### DIFF
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -106,50 +106,12 @@ private
     params["update-type"] == "major"
   end
 
-  def valid_content?(content, type)
-    return false if content.nil?
-
-    required_keys =
-      type.to_sym == :landing ? required_landing_page_keys : required_hub_page_keys
-    missing_keys = (required_keys - content.keys)
-    if missing_keys.any?
-      flash["alert"] = "Invalid content - please recheck GitHub and add #{missing_keys.join(', ')}."
-      return false
-    end
-
-    true
-  end
-
   def slug
     params[:slug] || params[:coronavirus_page_slug]
   end
 
   def slug_unknown?
     !page_configs.key?(slug.to_sym)
-  end
-
-  def required_landing_page_keys
-    %w[
-      title
-      meta_description
-      header_section
-      announcements_label
-      announcements
-      nhs_banner
-      sections
-      topic_section
-      notifications
-    ]
-  end
-
-  def required_hub_page_keys
-    %w[
-      title
-      header_section
-      sections
-      topic_section
-      notifications
-    ]
   end
 
   def page_configs
--- a/app/controllers/coronavirus_pages_controller.rb
+++ b/app/controllers/coronavirus_pages_controller.rb
@@ -19,13 +19,11 @@ class CoronavirusPagesController < ApplicationController
   end
 
   def update
-    if slug_unknown?
-      flash["alert"] = "Page could not be updated because the configuration cannot be found."
-    else
-      fetch_content_and_push
-    end
+    return slug_unknown_for_update if slug_unknown?
 
-    redirect_to prepare_coronavirus_page_path
+    message =
+      draft_updater.errors ? { alert: draft_updater.errors } : { notice: "Draft content updated" }
+    redirect_to prepare_coronavirus_page_path(slug), message
   end
 
   def publish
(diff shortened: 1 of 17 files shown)


app/controllers/coronavirus_pages_controller.rb
```diff
@@ -81,10 +77,14 @@ def fetch_content_and_push
         end
       end
     else
-      flash["alert"] = "Error received from GitHub - #{response.code}"
+      flash["alert"] = "Error received from GitHub - #{builder.errors.to_sentence}"
```
sihugh: 👍  pleased about this :-)

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -65,13 +65,9 @@ def publish_page
...
 
   def fetch_content_and_push
-    response = RestClient.get(page_config[:raw_content_url])
-
-    if response.code == 200
-      corona_content = YAML.safe_load(response.body)["content"]
-
-      if valid_content?(corona_content, page_type)
-        presenter = CoronavirusPagePresenter.new(corona_content, page_config[:base_path])
+    if details_builder.data && details_builder.success?
```
sihugh: Is it worth encapsulating this check in `details_builder` in entirety?

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -108,7 +108,7 @@ def valid_content?(content, type)
     return false if content.nil?
 
     required_keys =
-      type == :landing ? required_landing_page_keys : required_hub_page_keys
+      type.to_sym == :landing ? required_landing_page_keys : required_hub_page_keys
```
sihugh: we might be able to get rid of some of these checks if we can move towards defining a govuk-content-schemas schema

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -46,7 +46,7 @@ def coronavirus_page
   end
 
   def updater
-    CoronavirusPages::Updater.new(slug)
+    CoronavirusPages::ModelBuilder.new(slug)
```
sihugh, line 47: I don't think it's used in multiple places yet, but is it worth memoizing to prevent potentially re-instantiating multiple times in future work?

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -42,13 +40,22 @@ def initialise_coronavirus_pages
   end
 
   def coronavirus_page
-    @coronavirus_page ||= updater.page
+    @coronavirus_page ||= model_builder.page
```
sihugh, line 43: Fair enough, I guess `page` is the only public method on `model_builder`. Could be worth combining though :-)

app/services/coronavirus_pages/draft_updater.rb
```diff
@@ -0,0 +1,38 @@
...
+    def payload
+      if content_builder.success?
+        CoronavirusPagePresenter.new(content_builder.data, base_path)
+      else
+        raise DraftUpdaterError, content_builder.errors.to_sentence
+      end
+    end
+
+    def send
+      @send ||= Services.publishing_api.put_content(content_id, payload)
```
sihugh, line 26: Might we want this to run multiple times? Possibly not, but it might be a little surprising to have a one-shot updater? (Arguing with myself now)

reggieb: I think having a long lived object handling multiple requests is bad OOP practice. If you want to avoid mutation and threading issues the best policy is to have an instance of an object handle each request. So each instance of the draft updater handles one update process. If you want to run the update again, you initiate another instance.

app/controllers/coronavirus_pages_controller.rb
reggieb: This was the root of a number of discussions between @hannako and I. In the final form, this code has moved to the `DraftUpdater` where it looks like this:

```ruby
    def payload
      if content_builder.success?
        CoronavirusPagePresenter.new(content_builder.data, base_path)
      else
        raise DraftUpdaterError, content_builder.errors.to_sentence
      end
    end
```
Note that `DetailsBuilder` has been renamed to `ContentBuilder` in the final version.

app/controllers/coronavirus_pages_controller.rb
reggieb: Yes - that is the intention. Just not in this PR :)

app/controllers/coronavirus_pages_controller.rb
reggieb: Yes - good idea

PR #1084, opened 2020-07-01
Move livestream logic

hannako:
## What
Move the logic that combines the livestream data into the payload for publishing api, out of the CoronavirusPagePresenter and into the ContentBuilder and LivestreamUpdater.

## Why
When we make content changes, we fetch the github yaml file, turn it into a hash and combine it with the coronaviruspage data persisted in the database. We should add the livestream data persisted in the database to the content hash as well, rather than doing that elsewhere in the code base. As that is the pattern that can be continued as we persist more and more parts of the github content into the collections publisher database.

### DIFF
--- a/app/services/live_stream_updater.rb
+++ b/app/services/live_stream_updater.rb
@@ -39,7 +39,7 @@ private
 
   def update_content_item
     with_longer_timeout do
-      Services.publishing_api.put_content(landing_page_id, live_stream_payload)
+      Services.publishing_api.put_content(landing_page_id, payload)
     rescue GdsApi::HTTPErrorResponse
       object.update(url: live_url, formatted_stream_date: live_date)
     end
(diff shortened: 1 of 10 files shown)


spec/factories.rb
```diff
@@ -248,4 +248,12 @@
     content { "(#{Faker::Lorem.sentence})[/#{File.join(Faker::Lorem.words)}]" }
     coronavirus_page
   end
+
+  factory :live_stream do
+    url { "https://www.youtube.com/watch?v=NiplUCnwc5A" }
+    formatted_stream_date { "1 April 2020" }
+    trait :without_validations do
```
hannako, line 255: `without_validations`:

There is url validation on the livestream model, which meant adding a stub every time we create a payload for publishing api in a spec. This isn't really representative of how the app works, because there is a livestream object in the database, so we are just fetching it rather than creating it.

app/services/live_stream_updater.rb
```diff
@@ -67,15 +67,25 @@ def presenter
     CoronavirusPagePresenter.new(live_content_item["details"], "/coronavirus")
   end
 
-  def live_stream_payload
+  def payload
     presenter.payload.merge(
       {
         "title" => "Coronavirus (COVID-19): what you need to do",
         "description" => live_content_item["description"],
+        "details" => live_content_item["details"].deep_merge(live_stream_payload),
```
hannako, line 75: Previously `presenter.payload` on line 71 contained the livestream content. Now it's removed from the presenter it needs to be added in here. Added benefit is it will be easier to remove in the future because all the livestream updater stuff is in one class.

spec/controllers/sub_sections_controller_spec.rb
```diff
@@ -31,7 +32,7 @@
       stub_request(:get, raw_content_url)
         .to_return(body: raw_content)
       stub_coronavirus_publishing_api
-      stub_youtube
```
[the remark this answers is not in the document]
hannako: Yes it's used in the feature tests. Because I removed the code that tested the live stream object being initialised (which is when the youtube url is validated), I thought I'd leave the database empty for the feature tests so that bit of the code was still being used.

PR #1085, opened 2020-07-01
Hide markdown editor bullets button from covid accordion editor

danacotoran:
## What

- Extend markdown editor component with an option to hide the bullets button on the toolbar.
- Remove the bullets button from the markdown editor toolbar on the coronavirus accordion section editor form.

Markdown editor component with the bullets button vs without bullets button:

Screenshot 2020-07-01 at 11 56 30

## Why

Users should not need to use this control in this particular instance, we are removing the button from the toolbar as an additional measure.  

https://trello.com/c/thetabg3

### DIFF
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -1,6 +1,7 @@
 <%
   textarea ||= {}
   component_classes = %w[app-c-markdown-editor govuk-form-group]
+  hide_bullets_button ||= false
 %>
 <%= tag.div class: component_classes do %>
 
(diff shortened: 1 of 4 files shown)


app/views/components/_markdown_editor.html.erb
```diff
@@ -15,9 +16,11 @@
           <md-link class="app-c-markdown-editor__toolbar-button" title="Link" aria-label="Link">
             <%= render "components/markdown_editor/link.svg" %>
           </md-link>
-          <md-unordered-list class="app-c-markdown-editor__toolbar-button" title="Bullets" aria-label="Bullets">
-            <%= render "components/markdown_editor/bullets.svg" %>
-          </md-unordered-list>
+          <% if !hide_bullets_button %>
```
DilwoarH, line 19:
```suggestion
          <% unless hide_bullets_button %>
```

danacotoran: Thanks for the suggestion @DilwoarH. I always forget about this useful `unless` statement, for some reason.

PR #1088, opened 2020-07-07
Coronavirus publishing tool: Form landing page content payload correctly

hannako:
## What
When #1084 was merged, a bug was introduced. 
- Updating the livestream url was behaving as normal.
- Updating the page via the prepare action was removing the livestream content still stored in github from the content item

## How
Combine livestream content currently stored in github, with the persisted livestream data, when we publish the landing page.

### DIFF
--- a/app/services/coronavirus_pages/content_builder.rb
+++ b/app/services/coronavirus_pages/content_builder.rb
@@ -11,7 +11,7 @@ class CoronavirusPages::ContentBuilder
       validate_content
       data = github_data
       data["content_sections"] = sub_sections_data # Rename to sections when ready to go live
-      data["live_stream"] = live_stream_data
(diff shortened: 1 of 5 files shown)


app/services/coronavirus_pages/content_builder.rb
```diff
@@ -71,6 +72,10 @@ def github_data
     @github_data ||= github_raw_data["content"]
   end
 
+  def github_live_stream_data
+    github_data["live_stream"] ||= {}
```
sihugh, line 76: Is this supposed to be just `||`?

hannako: Yes, thank you. Have amended the commit to remove it.

PR #1089, opened 2020-07-07
Cache bust yaml fetch from Github

sihugh:
Github caches requests to the raw endpoint for 5 minutes (and it feels longer sometimes).  Adding a cache bust query parameter seems to defeat this, and will hopefully make repeated updates more reliable in the short term.

In the long term, the yaml is going to be less important as we're pulling this into the publishing tool.


spec/controllers/coronavirus_pages_controller_spec.rb
```diff
@@ -7,6 +7,7 @@
   let(:coronavirus_page) { create :coronavirus_page, :of_known_type }
   let(:slug) { coronavirus_page.slug }
   let(:raw_content_url) { CoronavirusPages::Configuration.page(slug)[:raw_content_url] }
+  let(:raw_content_url_regex) { Regexp.new(raw_content_url) }
```
hannako, line 10: I've never seen Regexp.new before. That's so handy!

PR #1096, opened 2020-07-14
Coronavirus publishing tool: Set and display state

hannako:
## What

- Add some documentation
- Toggle state of CoronavirusPage when page is published, updated, or changes are discarded


docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
+## Coronavirus Publishing Tool
+
+There are currently three coronavirus hub pages(/coronavirus/business-support, /coronavirus/education-and-childcare, /coronavirus/worker-support), and a coronavirus landing page. As we approached lockdown, a basic publishing interface was required at short notice. This is how it works:
+
+1. Content is added to the relevant yaml file in the [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content/tree/master/content) repository.
+2. Content editor with Coronavirus editor signon permissions visits [Coronavirus pages](https://collections-publisher.publishing.service.gov.uk/coronavirus) tab of collections publisher.
```
reggieb, line 6: I'd suggest a slight rewording to address two issues
- I don't think it's immediately clear that Content editor is a user
- The permission is defined as a specific string, which I think would helpful to use here. 

So I'd suggest:
```
2. Content editor (a user with "Coronavirus editor" permissions) visits [Coronavirus pages](https://collections-publisher.publishing.service.gov.uk/coronavirus) tab of collections publisher.
```

docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
...
+3. Selects which page to publish.
+4. Clicking update draft and publish buttons, fetches the yaml file, parses it, and sends it to publishing api as JSON.
+5. Collections renders the page by reading from this updated content item.
+
+### First iteration of new tool
+
+In order to reduce the load on content designers and supporting developers, a better publishing tool was required that would remove the need to make content changes via github. The first iteration of this tool was to move the accordion content out of the yaml file in govuk-coronavirus-content, and into the collections publisher database. This means it can be edited from collections publisher directly via a friendlier UI. **NB the content for the rest of the page must still be edited via the steps above.**
```
reggieb, line 13: Perdantic point, but this paragraph starts in the past tense and then shifts to the present tense half way through. As I presume this is a live document, and the behaviour may well change, this should be a record of what happened at the first iteration and therefore should all be in the past tense - even where behaviour introduced in this iteration is still extant. So:
```
This means it could be edited from collections publisher directly via a friendlier UI. 
**NB the content for the rest of the page would still be edited via the steps above.**
```

docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Future iterations
+
+Future iterations will be needed to move the content that remains in the govuk-coronavirus-content repository out of the yaml file, and into collections publisher.
+
+### How to create a new hub page
+
+1. Create a new yaml file following the format of those currently in [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content/tree/master/content) repository, and add page content.
+2. Add a basic page configuration to [this file](app/services/coronavirus_pages/configuration.rb) in Collections Publisher.
+3. Visiting the Coronavirus page tab of collections publisher will [create a coronavirus page model](https://github.com/alphagov/collections-publisher/blob/bbc667e5ab1c76fc0038fb4c0b14434b6ad3283c/app/controllers/coronavirus_pages_controller.rb#L50) based on the configuration.
```
reggieb, line 43: I'm not sure about this - does the link really want to point at specific version of the controller. How will changes to the controller after this document is created, affect this behaviour. I wonder if something along the lines of:
```
Visiting the Coronavirus page tab of collections publisher will create a coronavirus page model 
based on the configuration using the service CoronavirusPages::ModelBuilder.
```
With a more generic link to the controller?

docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
...
+4. The result will be two new links are be created on the [index page](https://collections-publisher.publishing.service.gov.uk/coronavirus):
```
reggieb, line 44: Typo - there's an errant `be` I think. Pesky little blighters.

I'd suggest
```
As a result two new links will be created ...
```

docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
...
+  - Edit the \<new hub> page accordion
+  - Publish \<new hub> page
+5. The Edit link can be used to populate the accordion section on the page.
+6. The Publish link can be used to pull in the content added to the yaml file created in step one, and send it to publishing api.
```
reggieb, line 48: Does 6 need some mention of the merging of data that goes on? Something like:
```
6. The Publish link can be used to pull in the content added to the yaml file created in step
one, merge that content with the page sub section data, and send it to publishing api.
```

docs/coronavirus_page_publishing_tool.md
```diff
@@ -0,0 +1,48 @@
...
+
+Future iterations will be needed to move the content that remains in the govuk-coronavirus-content repository out of the yaml file, and into collections publisher.
+
+### How to create a new hub page
+
+1. Create a new yaml file following the format of those currently in [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content/tree/master/content) repository, and add page content.
+2. Add a basic page configuration to [this file](app/services/coronavirus_pages/configuration.rb) in Collections Publisher.
+3. Visiting the Coronavirus page tab of collections publisher will [create a coronavirus page model](https://github.com/alphagov/collections-publisher/blob/bbc667e5ab1c76fc0038fb4c0b14434b6ad3283c/app/controllers/coronavirus_pages_controller.rb#L50) based on the configuration.
+4. The result will be two new links are be created on the [index page](https://collections-publisher.publishing.service.gov.uk/coronavirus):
+  - Edit the \<new hub> page accordion
```
reggieb, line 45: I'm not sure we should be documenting the elements as "accordion". That's how they are currently being presented on the page. A restyling of the page (or a future reuse of this mechanism on other pages) could make accordion inappropriate. In the data and the models they are Sub-Sections. I think that's how they should be documented.

app/controllers/coronavirus_pages_controller.rb
```diff
@@ -115,4 +115,8 @@ def slug_unknown?
   def page_configs
     CoronavirusPages::Configuration.all_pages
   end
+
+  def toggle_state(state)
+    coronavirus_page.update(state: state)
+  end
```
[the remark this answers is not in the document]
hannako: Yep 👍

docs/coronavirus_page_publishing_tool.md
hannako: oh yes definitely, thanks 👍

docs/coronavirus_page_publishing_tool.md
hannako: Content folks call them accordions, and the design has accordion as the title of the page. Maybe i need to explain that.

reggieb: OK :+1:

PR #1101, opened 2020-07-16
Coronavirus publishing tool: remove feature flag

hannako:
Remove checks for Unreleased feature signon permission that currently hide the new tools features.

https://trello.com/c/DU5dRL03/415-remove-feature-flag-hiding-the-coronavirus-publishing-tool

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::Base
 
 private
 
-  helper_method :gds_editor?, :active_navigation_item, :coronavirus_editor?, :livestream_editor?, :unreleased_feature_user?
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -25,10 +25,6 @@ def livestream_editor?
     current_user.has_permission? "Livestream editor"
   end
 
-  def unreleased_feature_user?
-    current_user.has_permission? "Unreleased feature"
```
sihugh, line 29: I was wondering about either leaving this in, or coming up with some agreed thing to use across apps so that we don't proliferate different versions of this?

hannako: Ah so leave it in the controller for future prototypes - yeah good idea.

PR #1102, opened 2020-07-16
Stop publishing email-signup routes for Topic's

1pretz1:
These email subroutes get published when a Topic (specialist sector) is
published / republished. As part of the work to move these redirected
routes out of collections-publisher we need to stop publishing them in
the first place.

After this gets merged we should republish all the Topic's via https://github.com/alphagov/collections-publisher/blob/master/lib/tasks/publishing_api.rake#L12.
This will keep the publishing-api in sync.

---

Trello:
https://trello.com/c/Fi2Z2zDy/399-redirect-and-remove-legacy-email-signup-routes-for-collections

### DIFF
--- a/spec/presenters/topic_presenter_spec.rb
+++ b/spec/presenters/topic_presenter_spec.rb
@@ -104,11 +104,10 @@ RSpec.describe TopicPresenter do
         expect(presented_data[:public_updated_at]).to eq(the_past.iso8601)
       end
 
-      it "includes routes for latest, and email_signups in addition to base route" do
+      it "includes a /latest route in addition to the base route" do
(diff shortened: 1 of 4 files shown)


spec/presenters/topic_presenter_spec.rb
```diff
@@ -104,11 +104,10 @@
         expect(presented_data[:public_updated_at]).to eq(the_past.iso8601)
       end
 
-      it "includes routes for latest, and email_signups in addition to base route" do
+      it "includes route for latest in addition to base route" do
```
kevindew, line 107:
```suggestion
      it "includes a /latest route in addition to the base route" do
```

PR #1104, opened 2020-07-17
Add temp rake task to redirect legacy signups

1pretz1:
Adds a temporary rake task to create redirects for specialist sector
email signup URL's.

The current behaviour is for these specialist sector email-signup URL's
to be sent to collections which then redirects to email-alert-frontend.

This commit creates a redirect for each specialist sector email signup
URL which points to the relevant parent page rendered by collections.

Depends on:
https://github.com/alphagov/collections-publisher/pull/1102

---

Trello:
https://trello.com/c/Fi2Z2zDy/399-redirect-and-remove-legacy-email-signup-routes-for-collections

### DIFF
--- a/lib/tasks/redirect_legacy_email_signups.rake
+++ b/lib/tasks/redirect_legacy_email_signups.rake
@@ -1,0 +1,27 @@
+desc "Redirect legacy email signups"
+task redirect_legacy_email_signups: :environment do
+  # Exclude parent specialist sectors as they never included an email signup option
+  specialist_sector_slugs = Topic.where(state: "published").where.not(parent_id: nil).map(&:full_slug)
+
+  specialist_sector_slugs.each do |slug|
+    base_path = "/topic/#{slug}/email-signup"
+    destination = "/topic/#{slug}"
+
+    redirect = {
+      base_path: base_path,
+      document_type: "redirect",
+      schema_name: "redirect",
+      publishing_app: "collections-publisher",
+      update_type: "major",
+      redirects: [
+        { path: base_path, type: "exact", destination: destination },
+      ],
+    }
+
+    content_id = SecureRandom.uuid
+    Services.publishing_api.put_content(content_id, redirect)
+    Services.publishing_api.publish(content_id)
+
+    puts "Redirected #{base_path} to #{destination}"
+  end
+end


lib/tasks/redirect_legacy_email_signups.rake
```diff
@@ -0,0 +1,31 @@
+desc "Redirect legacy email signups"
+task redirect_legacy_email_signups: :environment do
+  specialist_sectors_info = Topic.where(state: %w[draft published]).where.not(parent_id: nil).map do |ss|
+    { state: ss.state, slug: ss.full_slug }
+  end
+
+  specialist_sectors_info.map do |specialist_sector|
+    base_path = "/topic/#{specialist_sector[:slug]}/email-signup"
+    destination = "/topic/#{specialist_sector[:slug]}"
```
kevindew, line 9: Should the destination be "/email-signup?topic=" so they remain sent to the same page as now? 

I can see in discussion on the card it's stated that this isn't possible but I think Ben just got caught with url encoding.

As an example:

```
irb(main):001:0> content_id = SecureRandom.uuid
=> "af079a0c-7333-477b-baf7-2c2efe748b97"
irb(main):002:0> redirect = {
irb(main):003:1*   base_path: "/topic/immigration-operational-guidance/non-compliance-biometric-registration/email-signup",
irb(main):004:1*   document_type: "redirect",
irb(main):005:1*   schema_name: "redirect",
irb(main):006:1*   publishing_app: "collections-publisher",
irb(main):007:1*   update_type: "major",
irb(main):008:1*   redirects: [
irb(main):009:2*     { path: "/topic/immigration-operational-guidance/non-compliance-biometric-registration/email-signup", type: "exact", destination: "/email-signup?topic=#{ERB::Util.url_encode('/topic/immigration-operational-guidance/non-compliance-biometric-registration')}" },
irb(main):010:2*   ],
irb(main):011:1* }
=> {:base_path=>"/topic/immigration-operational-guidance/non-compliance-biometric-registration/email-signup", :document_type=>"redirect", :schema_name=>"redirect", :publishing_app=>"collections-publisher", :update_type=>"major", :redirects=>[{:path=>"/topic/immigration-operational-guidance/non-compliance-biometric-registration/email-signup", :type=>"exact", :destination=>"/email-signup?topic=%2Ftopic%2Fimmigration-operational-guidance%2Fnon-compliance-biometric-registration"}]}
irb(main):012:0> Services.publishing_api.put_content(content_id, redirect)
=> #<GdsApi::Response:0x00007f0036a20c50 @http_response=<RestClient::Response 200 "{\"auth_bypa...">, @web_urls_relative_to=nil>
irb(main):013:0> Services.publishing_api.publish(content_id)
=> #<GdsApi::Response:0x00007f0036a32838 @http_response=<RestClient::Response 200 "{\"content_i...">, @web_urls_relative_to=nil>
```

lib/tasks/redirect_legacy_email_signups.rake
```diff
@@ -0,0 +1,31 @@
...
+      redirects: [
+        { path: base_path, type: "exact", destination: destination },
+      ],
+    }
+
+    content_id = SecureRandom.uuid
+    Services.publishing_api.put_content(content_id, redirect)
+
+    if specialist_sector[:state] == "published"
+      Services.publishing_api.publish(content_id, "republish")
```
kevindew, line 26: I don't think we need the "republish" argument here, these are technically new publishing of the redirect document and not a republishing of it.

lib/tasks/redirect_legacy_email_signups.rake
```diff
@@ -0,0 +1,31 @@
+desc "Redirect legacy email signups"
+task redirect_legacy_email_signups: :environment do
+  specialist_sectors_info = Topic.where(state: %w[draft published]).where.not(parent_id: nil).map do |ss|
```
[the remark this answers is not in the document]
kevindew: Also, it'd be useful if the commit or a comment said why we need a parent_id of nil. I didn't know myself so had to go work it out from the frontend.

lib/tasks/redirect_legacy_email_signups.rake
benthorner: Huh, that's strange - looks like you're right. However, we'd still have to collect the individual topic_ids for these pages, which seems like excessive effort.

We took this approach with Service Manual Publisher, and it has the advantage of following the parent page, if it were ever itself unpublished.

lib/tasks/redirect_legacy_email_signups.rake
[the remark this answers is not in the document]
1pretz1: CollectionsPublisher forces publishers to submit a redirect URL when they're archiving a topic so I'd assume they would all have one. I've double checked this from router's perspective and they all are being redirected to other topic pages.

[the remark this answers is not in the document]
1pretz1: I've added a comment which should make that part of the query more clear.

As for:
> Do we need to worry about redirects for draft topics?

I thought it'd be good to add these in for drafts too so draft and published specialist sectors email-signup paths are at least consistent in the router. Otherwise we'd have dangling /email-signup routes for the draft sectors, what do you think?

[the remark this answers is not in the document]
benthorner: Ah, of course - me being dumb. Thanks for checking @1pretz1 💯

lib/tasks/redirect_legacy_email_signups.rake
1pretz1: I've decided to take Kevin's approach of redirecting to the email-signup path as it seems more useful then redirecting to the parent page. 

I'm not sure what you mean @benthorner when you say 'excessive effort' in retrieving the topic id's?

lib/tasks/redirect_legacy_email_signups.rake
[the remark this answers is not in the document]
kevindew: > > Do we need to worry about redirects for draft topics?
> 
> I thought it'd be good to add these in for drafts too so draft and published specialist sectors email-signup paths are at least consistent in the router. Otherwise we'd have dangling non redirect /email-signup routes for the draft sectors, what do you think?

Sure that makes sense but I'd not worry too much about router consistency since that is really the domain of communication between content store and router-api (and it'd only be in the draft stack). I think we end up in a slightly weirder situation if we present the drafts to the Publishing API as it adds in a pile of content items that will never be published to live and may linger around to cause a surprise one day.

lib/tasks/redirect_legacy_email_signups.rake
benthorner: Hmm...I don't think we should do that: if the topic is unpublished in the future, these links will still work. We would actually want them to stop working.

> when you say 'excessive effort' in retrieving the topic id's

Ah, I was getting confused between `topic` and `topic_id`. I should just stop reviewing this PR, sorry 😬. Yes, that would work. However, I still think the above argument is a good one.

kevindew: Either approach you'd have the suffix links still working and returning 301. Is your concern based on whether Email Alert Frontend can handle a topic that no-longer exists? If I remember right it looks up a content item and seems to 404 on non-existent ones e.g. https://www.integration.publishing.service.gov.uk/email-signup?topic=%2Ftopic%2Fnon-existent ?

benthorner: > whether Email Alert Frontend can handle a topic that no-longer exists

I'm seeing redirecting to the parent as simpler, because it's not coupled to the behaviour of another app. It's also the approach we took with service manual signups, so consistent in that respect.

Since the redirect is to the parent, it's not blocking the user. I don't feel too strongly either way, apart from this being a second change in approach, for something that seems pretty arbitrary to me.

I'm happy for you and @1pretz1 to agree on what to do 👍.

lib/tasks/redirect_legacy_email_signups.rake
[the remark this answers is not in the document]
1pretz1: Okay dokay that's a fair point. 

I've removed drafts from the query so the only redirects getting created are for published specialist sectors.

lib/tasks/redirect_legacy_email_signups.rake
1pretz1: I've changed tact again 😄 

As consistency is king and these redirects should always be useful if pointing at the parent page as there should always be an end destination that isn't a 404. That okay with you @kevindew?

kevindew: Yup no problem, seems weird to do it that way but don't care strongly enough to argue about it.

lib/tasks/redirect_legacy_email_signups.rake
```diff
@@ -0,0 +1,29 @@
+desc "Redirect legacy email signups"
+task redirect_legacy_email_signups: :environment do
+  # Exclude parent specialist sectors as they never included an email signup option
+  specialist_sectors_info = Topic.where(state: "published").where.not(parent_id: nil).map do |ss|
+    { state: ss.state, slug: ss.full_slug }
+  end
```
kevindew, line 6: You don't need to pull out state any more. You can do `Topic.where(state: "published").where.not(parent_id: nil).pluck(:full_slug)` (presuming that is a db column rather than an instance method)

1pretz1: Doh of course, sloppy half job removal of this code! 😆

1pretz1:
I've tried running the individual lines in a console in integration (my normal testing approach for this sort of PR) and everything worked as expected. 

I'll deploy it again and give it a raz in integration to double check.

> I guess it'll also need to be deployed with or after #1102 so any publisher activity doesn't clobber the redirects (though I guess publisher activity is rare in this area).

And yes theres a 'Depends on:' section in the PR description with that PR underneath.

PR #1115, opened 2020-07-27
Coronavirus publishing tool: Tidy up

hannako:
## What

- Remove sections content from yaml test fixtures
- The first time a page is created from the config file we should no longer build its sub_sections from the [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content) yaml file. The sub_sections data doesn't live there anymore, so sub_sections will need to be added via the UI.
- Remove rake task that was needed to wipe the database prior to the tool going live
- Create a new rake task that will allow `sections_title` and `title` fields to be resynced in the event that they are changed in the yaml file.

https://trello.com/c/HZCPRtAD/419-remove-unused-code-from-collections-publisher


lib/tasks/coronavirus.rake
```diff
@@ -1,23 +1,25 @@
...
-      CoronavirusPage.any? ? "Something went wrong, #{slugs} were not deleted." : "Done"
-    else
-      puts "This task is in dry run mode. It would have removed #{count} coronavirus pages from the database: #{slugs}"
-    end
+  task :resync_titles, [:slug] => [:environment] do |_task, args|
+    page = CoronavirusPage.find_by(slug: args.slug)
+    builder ||= CoronavirusPages::ModelBuilder.new(args.slug)
+    title = builder.title
+    sections_heading = builder.sections_heading
+    page.update!(title: title, sections_title: sections_heading)
```
reggieb, line 20: As `title` and `section_heading` defined on lines 18 and 19 are only used on line 20, this would be clearer:
```ruby
page.update!(
  title: builder.title,
  section_title: builder.section_title
)
```

spec/services/coronavirus_pages/model_builder_spec.rb
```diff
@@ -98,5 +70,16 @@
...
+    it "creates sub_sections with a position that reflects their order in the content item" do
+      coronavirus_page
+      discard_changes
+      input_order = live_sections.pluck("title").each_with_index.to_h.invert
+
+      coronavirus_page.sub_sections.each do |sub_section|
+        position = sub_section.position
+        expect(input_order[position]).to eq(sub_section.title)
+      end
+    end
```
reggieb, line 83: `coronavirus_page` is defined via `let!` and therefore doesn't need calling directly as it would have already been called as the spec is started.

`input_order` does not rely on the `coronavirus_page` nor `discard_changes` calls. So should really be defined outside the `it` clause.

I'd suggest this:
```ruby
let(:input_order) {  live_sections.pluck("title").each_with_index.to_h.invert }

it "creates sub_sections with a position that reflects their order in the content item" do
  discard_changes
  coronavirus_page.sub_sections.each do |sub_section|
    position = sub_section.position
    expect(input_order[position]).to eq(sub_section.title)
  end
end
```

lib/tasks/coronavirus.rake
hannako: thanks!

spec/services/coronavirus_pages/model_builder_spec.rb
hannako: Thanks Rob. I'll remove the unnecessary call of `coronavirus_page` but leave input order where it is, as it's the only place it's used and makes it a bit faster to scan and understand.

PR #1122, opened 2020-08-17
Allow users to set a link as a "featured link" in an accordion

leenagupte:
Trello: https://trello.com/c/7Li6NbQr

# What's changed?
Add an input field to the accordion page to allow users to mark one of the links featured in the content as a "featured" link. In the frontend application, links that have "featured_link" set to true automatically have special styling applied to them (the arrow head).
 
Add a "featured_link" attribute for a sub_section. This "featured_link" should contain the path to a page on GOV.UK. 
Also adds validations to make sure that the "featured_link" appears in the body content of the accordion.
Finally, the discard draft action has been updated to allow any changes to the featured link to be reverted with the rest of the draft changes.

# Why?
At the moment the featured link is generate automatically if a topic link is included in the accordion.

We need the ability to add "featured links" (ie with an arrow and meta description) to content links as well as hub page links. 

This is specifically important for the coronavirus testing page which is a quick answer guide format, and is currently only linked to in a standard link style. 

We want to emphasise this content in the accordion as the main starting point for users who want to find out more about testing and get routed to the right testing service for their situation.

# Expected changes
Screenshot 2020-08-17 at 14 24 47
Screenshot 2020-08-14 at 12 43 38
Screenshot 2020-08-14 at 12 44 13

### DIFF
--- a/spec/controllers/sub_sections_controller_spec.rb
+++ b/spec/controllers/sub_sections_controller_spec.rb
@@ -84,6 +84,10 @@ RSpec.describe SubSectionsController, type: :controller do
       let(:featured_link) { "/#{SecureRandom.urlsafe_base64}" }
 
       it "stores the featured link" do
+        content_id = SecureRandom.uuid
+        stub_publishing_api_has_item(base_path: featured_link, content_id: content_id)
+        stub_publishing_api_has_lookups(featured_link.to_s => content_id)
+
         content = "###{Faker::Lorem.sentence}\n[Link text](#{featured_link})"
         sub_section_params.merge!(content: content, featured_link: featured_link)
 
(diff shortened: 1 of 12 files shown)


spec/controllers/sub_sections_controller_spec.rb
```diff
@@ -10,6 +10,7 @@
   let!(:live_stream) { create :live_stream, :without_validations }
   let(:title) { Faker::Lorem.sentence }
   let(:content) { "###{Faker::Lorem.sentence}" }
+  let(:featured_link) { "/#{SecureRandom.urlsafe_base64}" }
```
steveholmes, line 12: Is this still needed, as there is a duplicate line at line 83?

leenagupte: Good spot! Will remove a fix up.

PR #1128, opened 2020-09-01
Add hidden_search_terms to publishing-api payload

leenagupte:
Trello: https://trello.com/c/JkaMKLxT
Zendesk: https://govuk.zendesk.com/agent/tickets/4167533

The content on coronavirus landing and hub pages isn't being indexed in the site search.

Search has a concept of `indexable_content`. This means that it looks for [certain fields](https://github.com/alphagov/search-api/blob/fb3eafcb2edbda5a1b7aeb90d6465342910bcf00/lib/govuk_index/presenters/indexable_content_presenter.rb#L3) like `body` in the "details" hash of a content item and makes them searchable.

Unfortunately, the [coronavirus landing](https://www.gov.uk/api/content/coronavirus) and hub pages have a lot of attributes in "details" but none of them match the fields that the search-api is looking for.

One of the accepted fields is "hidden_search_terms". [smart-answers](https://www.gov.uk/api/content/am-i-getting-minimum-wage) already take advantage of "hidden_search_fields" to index content that appears in the question flow.

This change goes through all of the section/accordion text content on the coronavirus page and "flattens" it into a list of sentences.

For example, if the `sections` in the details hash looks like:

```
"sections" => [
  {
    "title" => "Blue Shadow Virus",
    "sub_sections" => [
      {
        "list" => [
          {
            "url" => "/foo",
            "label" => "Apply for a test",
          },
          {
            "url" => "/bar",
            "label" => "What to do if you test positive",
          },
        ],
        "title" => "Testing for Blue Shadow Virus",
      },
    ],
  },
]
```

Then `hidden_search_terms will be populated with:

```
[
  "Blue Shadow Virus",
  "Testing for Blue Shadow Virus",
  "Apply for a test",
  "What to do if you test positive"
]
```

A safe navigation operator (&) is being used when looping through the "lists" in subsections as the code currently allows users to add content with just headings in it, i.e. no links.

At the moment these changes only apply to the accordion content, but they can be expanded to include the other sections later.

The landing and hub pages will need to be republished for these changes to take effect.

Screenshot 2020-09-01 at 14 13 28
Screenshot 2020-09-01 at 16 22 48


app/services/coronavirus_pages/content_builder.rb
```diff
@@ -97,4 +98,22 @@ def add_live_stream(data)
       data["live_stream"] = github_live_stream_data.merge(persisted_live_stream_data)
     end
   end
+
+  def hidden_search_terms
+    sections = sub_sections_data.map do |section|
+      [section[:title], subsections(section[:sub_sections])]
+    end
+
+    sections.flatten.compact
```
kevindew, line 107: I see you're using compact to remove any nil values, are you at all concerned about empty strings? I wonder if you might want select(&:present?).

Also would you want repeated terms multiple times (I see "Popular content" from your example) or would you want to `.uniq` them so they don't distort relevancy?

app/services/coronavirus_pages/content_builder.rb
```diff
@@ -97,4 +98,22 @@ def add_live_stream(data)
...
+  end
+
+  def subsections(items)
```
kevindew, line 110: I wonder if this method name is a little broad, since it only relates to `hidden_search_terms`. I notice there is already a [`sub_sections_data` method](https://github.com/alphagov/collections-publisher/blob/1e06a43a6156c11cd0bfcaa5b0118da49d667c68/app/services/coronavirus_pages/content_builder.rb#L80) that feels quite similar in naming to this.

kevindew (commented):
Nice one, this looks good - I've added a couple of suggestions

app/services/coronavirus_pages/content_builder.rb
leenagupte: Thanks! I forgot about uniqueness.

I’ve never used `select(&:present?)` before. It doesn’t read as nicely as `compact` but is definitely more useful.

leenagupte:
@kevindew Thanks for the review! I've updated and fixed up.

kevindew (approved):
Great, looks good to me

PR #1132, opened 2020-09-08
Change /eubusiness to an exact route

sihugh:
[We're getting errors](https://sentry.io/organizations/govuk/issues/1884114207/?environment=production&project=202213&referrer=slack) because router is sending sub paths of /eubusiness to collections. We only have the /eubusiness page right now, so an exact route is the way to go.

I'll check about removing the prefix route once this is republished.

https://trello.com/c/aRd8ha9V/437-fix-app-error-in-collections

### DIFF
--- a/lib/special_route_publisher.rb
+++ b/lib/special_route_publisher.rb
@@ -24,7 +24,7 @@ class SpecialRoutePublisher
 
   def self.routes
     {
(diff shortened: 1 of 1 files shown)


hannako (approved):
AH! Sorry!

PR #1142, opened 2020-09-17
Upgrade to Ruby 2.7.1

cdccollins:
These are the necessary changes for Ruby 2.7.1

I've left the Ruby version as 2.6.6 since there is a script to update all the apps' Ruby versions at once.


Rhian (approved):
This update looks good to me

PR #1147, opened 2020-09-24
Index timeline content in search

leenagupte:
Trello: https://trello.com/c/jdBmdPef
Follows on from: PR #1128 

# What?

Allow users to find the landing page by searching for timeline content in the GOV.UK site search.

# Why?

At the moment only the content in the accordions on the landing and hub pages are being indexed in search.

It's likely that the content that appears in the timeline might not be repeated anywhere else on GOV.UK, so we need to give users a way to find it.

# Expected changes
## Content item
Screenshot 2020-09-24 at 08 55 23

## Search data
Screenshot 2020-09-24 at 08 59 26

The landing and page will need to be republished for these changes to  take effect.

### DIFF
--- a/app/services/coronavirus_pages/content_builder.rb
+++ b/app/services/coronavirus_pages/content_builder.rb
@@ -100,7 +100,7 @@ class CoronavirusPages::ContentBuilder
   end
 
   def hidden_search_terms
-    search_terms_in_sections
+    search_terms_in_sections + search_terms_in_timeline
   end
 
   def search_terms_in_sections
@@ -119,5 +119,18 @@ class CoronavirusPages::ContentBuilder
 
       [subsection[:title], labels]
     end
+  end
+
+  def search_terms_in_timeline
+    return [] if github_data["timeline"].blank?
+
+    timeline = github_data["timeline"]["list"].map do |item|
+      [
+        item["heading"],
+        MarkdownService.new.strip_markdown(item["paragraph"]),
+      ]
+    end
+
+    timeline.flatten.select(&:present?).uniq
   end
 end
--- a/app/services/coronavirus_pages/content_builder.rb
+++ b/app/services/coronavirus_pages/content_builder.rb
@@ -100,6 +100,10 @@ class CoronavirusPages::ContentBuilder
   end
 
   def hidden_search_terms
+    search_terms_in_sections
+  end
+
+  def search_terms_in_sections
     sections = sub_sections_data.map do |section|
       [section[:title], search_terms_in_sub_sections(section[:sub_sections])]
     end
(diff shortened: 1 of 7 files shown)


spec/helpers/markdown_helper_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe MarkdownHelper do
+  describe "#strip_markdown" do
```
DilwoarH, line 4: Could we add a test with content that have multiple links, ie:

`this is [a link](/blaaahhh) and this is also [a liiiiinnkkk](/bajajajja)`

app/services/coronavirus_pages/content_builder.rb
```diff
@@ -116,4 +122,12 @@ def search_terms_in_sub_sections(items)
       [subsection[:title], labels]
     end
   end
+
+  def search_terms_in_timeline
+    return [] if github_data["timeline"].blank?
+
+    github_data["timeline"]["list"].map do |item|
+      strip_markdown(item["paragraph"])
```
DilwoarH, line 130: Should you be indexing the date too?

DilwoarH (changes requested):

app/services/coronavirus_pages/content_builder.rb
leenagupte: I don't think so. Are people likely to be searching for the date? Also the date is not in a typical date format, so I think that would make it hard to search for.

spec/helpers/markdown_helper_spec.rb
leenagupte: Sure!

leenagupte: @DilwoarH Updated and fixed up.

kevindew:
This looks good. I did wonder if you need to strip markdown and instead send HTML in this field - since search API already copes with body HTML but that's probably a bad idea as we wouldn't be specifying it has a content type of text/html.

I do wonder if you can achieve the markdown stripping with Kramdown rather than Redcarpet? since we already have Kramdown as a dependency this would mean avoiding an extra dependency.

app/services/coronavirus_pages/content_builder.rb
DilwoarH: If @helengds - doesn't think we need it, I'm happy for my change request to be dismissed

leenagupte:
@DilwoarH @kevindew Thanks for your reviews. I've followed @kevindew's advice and switched to using Kramdown. It's much less code. Thanks!  

I'd appreciate it if you could re-review 🥺 🙏

DilwoarH (approved):
LGTM 🥇  - probably worth a backend dev to confirm too

kevindew (approved):

PR #1179, opened 2020-11-20
Discard drafts of Announcements

leenagupte:
Trello: https://trello.com/c/HGZZJgiD

# What's changed and why?

Now that we are allowing users to create and edit announcements via the publishing tool, we should allow them to discard any changes they make too, rather than having to manually unpick them.

# Expected changes

There are no visual changes to the user. When the discard draft button is clicked on the summary page, changes to any announcements should revert to the live ("published") version.


JonathanHallam (approved):
Tested this in integration and it works as it should. Really excellent commit history, thank you! 🥇

PR #1185, opened 2020-12-02
Include step content in the body field

sihugh:
We want step content to be indexed by search so we can find these pages more easily. Search-api can understand html, and steps are written in markdown.

We therefore want render all the step content and titles as html, not for public consumption, but so that search can index the page content.

I've tried to present the output fairly verbosely with plenty of space between sections. Fully fledged step by step pages are quite chunky, and should we experience problems it'd be nice to be able to view this bit in a debugger without it just being a mass of html.

https://trello.com/c/rsiS0jZ6/516-improve-visibility-of-import-export-step-by-steps-in-site-search

### DIFF
--- a/app/presenters/step_nav_presenter.rb
+++ b/app/presenters/step_nav_presenter.rb
@@ -70,6 +70,7 @@ private
         ],
         steps: steps,
       },
+      body: body,
     }
   end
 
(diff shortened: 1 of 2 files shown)


app/presenters/step_nav_presenter.rb
```diff
@@ -6,8 +6,9 @@ def initialize(step_nav)
 
   def render_for_publishing_api(publish_intent = PublishIntent.minor_update)
     payload = required_fields
-    payload.merge!(optional_fields)
-    payload.merge(publish_intent.present)
+    payload
+      .merge(optional_fields)
```
leenagupte, line 10: I think you need the `!` on merge, or it will be overwritten by the publish_intent.

sihugh: I've removed the change as it wasn't relevant to what I'm doing here 👍

hannako (approved):
This is a good idea. 

(We should all be more Fonz)

PR #1192, opened 2020-12-08
Add feature flag for backend code

leenagupte:
Trello: https://trello.com/c/gzWPC4LU

# What's changed and why?

The announcements UI was being hidden behind a user permission
"Unreleased feature" to stop them seeing and interacting with
announcements, however there was nothing to stop the backend code from
running.

This caused problems recently where the announcements section was
removed [1] in live because the code to populate publishing-api was
looking for the announcements in the wrong place. We added a flag to
check if there are any existing announcements to get around the problem
but since found that this had a knock-on effect when drafts are
discarded.

When the draft is discarded, announcements are retrieved from
publishing-api and populated in the database. As the local database had
announcements, any changes made to the announcements section via Github
weren't picked up because the flag we added evaluated to true.

To stop problems like occurring in the upcoming timeline feature, config
has been added that sets the conditions for when the backend code
associated with an unreleased feature should be run.

The config class relies on the presence of an environment variable. The
bonus of this is that we can enable unreleased_features differently in
different environments.

For example, if we set ENV["UNRELEASED_FEATURES"] in integration only,
that means we'll be able to test the feature in Integration, but it
would be turned off by default in staging and production.

So rather than doing something like:

```
data["announcements"] = announcements_data if coronavirus_page.announcements.any?
```

We can instead do:

```
data["announcements"] = announcements_data if Rails.application.config.unreleased_features?
```

[1]: #1191

### DIFF
--- a/app/services/coronavirus_pages/content_builder.rb
+++ b/app/services/coronavirus_pages/content_builder.rb
@@ -94,6 +94,8 @@ class CoronavirusPages::ContentBuilder
   end
 
   def persisted_live_stream_data
+    return {} unless LiveStream.any?
+
     live_stream = LiveStream.last
     {
       "video_url" => live_stream.url,
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -52,4 +52,6 @@ Rails.application.configure do
   # routes, locales, etc. This feature depends on the listen gem.
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
   config.hosts << "collections-publisher.dev.gov.uk"
+
+  config.unreleased_features = true
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -103,4 +103,6 @@ Rails.application.configure do
   # Rather than use a CSS compressor, use the SASS style to perform compression.
   config.sass.style = :compressed
   config.sass.line_comments = false
+
+  config.unreleased_features = ENV["UNRELEASED_FEATURES"].present?
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -47,4 +47,6 @@ Rails.application.configure do
 
   # Raises error for missing translations.
   # config.action_view.raise_on_missing_translations = true
+
+  config.unreleased_features = true
 end
--- a/spec/controllers/announcements_controller_spec.rb
+++ b/spec/controllers/announcements_controller_spec.rb
@@ -6,7 +6,6 @@ RSpec.describe AnnouncementsController, type: :controller do
   let(:stub_user) { create :user, :coronovirus_editor, name: "Name Surname" }
   let(:coronavirus_page) { create :coronavirus_page, :landing }
   let(:announcement) { create :announcement, coronavirus_page: coronavirus_page }
-  let!(:live_stream) { create :live_stream, :without_validations }
   let(:title) { Faker::Lorem.sentence }
   let(:path) { "/government/foo/vader/baby/yoda" }
   let(:published_at) { { "day" => "12", "month" => "12", "year" => "1980" } }
--- a/spec/controllers/reorder_sub_sections_controller_spec.rb
+++ b/spec/controllers/reorder_sub_sections_controller_spec.rb
@@ -4,7 +4,6 @@ RSpec.describe ReorderSubSectionsController, type: :controller do
   render_views
   let(:stub_user) { create :user, :coronovirus_editor, name: "Name Surname" }
   let(:coronavirus_page) { create :coronavirus_page, :of_known_type }
-  let(:live_stream) { create :live_stream, :without_validations }
   let(:slug) { coronavirus_page.slug }
   let(:raw_content_url) { CoronavirusPages::Configuration.page(slug)[:raw_content_url] }
   let(:fixture_path) { Rails.root.join "spec/fixtures/coronavirus_landing_page.yml" }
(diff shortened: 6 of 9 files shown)


kevindew:
This is a fancy looking technique, nice digging around.

I'd be a bit concerned about using Rails.env as the flag as that'll mean that you'd need to mock that if you wanted to test the feature flag being off, and I'm not sure if that causes other problems for a Rails test. Perhaps an env var would be more conventional?

When we wrestled with this on Content Publisher we wondered if we'd ever need something that listed individual features (as how this lists announcements) as we'd basically only need the flag until we gave it to users and then remove it. So rather than being a feature flag per-se it was more a flag to mark bits of the app that weren't ready. I wonder if that might be more applicable to this use case here?

leenagupte:
> So rather than being a feature flag per-se it was more a flag to mark bits of the app that weren't ready. I wonder if that might be more applicable to this use case here?

I thought that that was what this is doing! Is there another of setting the flag that I'm missing?

leenagupte:
> I'd be a bit concerned about using Rails.env as the flag as that'll mean that you'd need to mock that if you wanted to test the feature flag being off, and I'm not sure if that causes other problems for a Rails test. Perhaps an env var would be more conventional?

I suppose I'm trying to avoid having to do any puppet changes for something that's going to be relatively short-lived, hence not setting an env var. We have used Rails.env in the [past](https://github.com/alphagov/collections/pull/2099/commits/882868d7109f6203ac746d923fd2651991874e4a) though we were probably being super unconventional then too 😅

kevindew:
> > So rather than being a feature flag per-se it was more a flag to mark bits of the app that weren't ready. I wonder if that might be more applicable to this use case here?
> 
> @kevindew I thought that that was what this is doing! Is there another of setting the flag that I'm missing?

Sorry, I'll try rephrase to be more clear.

It seems unnecessary to be asking the question in code "is the _announcements_  feature turned on" because once announcements is ready you'll no longer need to ask the question. So the code of

`data["announcements"] = announcements_data if Feature.enabled?(:announcements)`

is basically the same as:

`data["announcements"] = announcements_data if Feature.live_features?`

> > I'd be a bit concerned about using Rails.env as the flag as that'll mean that you'd need to mock that if you wanted to test the feature flag being off, and I'm not sure if that causes other problems for a Rails test. Perhaps an env var would be more conventional?
> 
> I suppose I'm trying to avoid having to do any puppet changes for something that's going to be relatively short-lived, hence not setting an env var. We have used Rails.env in the [past](https://github.com/alphagov/collections/pull/2099/commits/882868d7109f6203ac746d923fd2651991874e4a) though we were probably being super unconventional then too 😅

Yeah, I see it as a 12 factor violation using `Rails.env.production?` to indicate an aspect of the deployment destinations configuration. There is some information on that in https://12factor.net/config. GOV.UK's past usage of that pattern causes all sort of problems, many of those needed to be resolved for [the introduction of publishing-e2e-tests](https://github.com/alphagov/publisher/pull/669/files). It can still lead to lingering problems too, I noticed only yesterday that we can't test the postcode check on the heroku preview app [due to a similar conditional](https://github.com/alphagov/collections/blob/37e6271dfd4af622d7aa04ce9a751781a63803b0/app/services/location_lookup_service.rb#L64).  

I don't find it too much of a pain to add environment variables but if you really wanted to avoid it you could use environment variables to turn on the feature - e.g `data["announcements"] = announcements_data if ENV["USE_DB_ANNOUNCEMENTS"]` and then you'd only need to turn this env var on during the test process and it could be removed from the code once the live environment is ready.

leenagupte:
>It seems unnecessary to be asking the question in code "is the announcements feature turned on" because once announcements is ready you'll no longer need to ask the question. So the code of
data["announcements"] = announcements_data if Feature.enabled?(:announcements)
is basically the same as:
data["announcements"] = announcements_data if Feature.live_features?

@kevindew The idea of naming the features is so that we could potentially work on more than one at a time. Of course now that announcements is live, that's not going to be a problem. I'll have a think about how this can be made more generic.

kevindew:
> > It seems unnecessary to be asking the question in code "is the announcements feature turned on" because once announcements is ready you'll no longer need to ask the question. So the code of
> > data["announcements"] = announcements_data if Feature.enabled?(:announcements)
> > is basically the same as:
> > data["announcements"] = announcements_data if Feature.live_features?
> 
> @kevindew The idea of naming the features is so that we could potentially work on more than one at a time. Of course now that announcements is live, that's not going to be a problem. I'll have a think about how this can be made more generic.

Sure, I'm not sure we have a scenario where we want a feature flag to remain for a feature once it's out of pre-release. So if we have multiple features but they all end up with the condition `Rails.env.production?` do we actually benefit from listing them individually?

leenagupte:
> > > It seems unnecessary to be asking the question in code "is the announcements feature turned on" because once announcements is ready you'll no longer need to ask the question. So the code of
> > > data["announcements"] = announcements_data if Feature.enabled?(:announcements)
> > > is basically the same as:
> > > data["announcements"] = announcements_data if Feature.live_features?
> > 
> > 
> > @kevindew The idea of naming the features is so that we could potentially work on more than one at a time. Of course now that announcements is live, that's not going to be a problem. I'll have a think about how this can be made more generic.
> 
> Sure, I'm not sure we have a scenario where we want a feature flag to remain for a feature once it's out of pre-release. So if we have multiple features but they all end up with the condition `Rails.env.production?` do we actually benefit from listing them individually?

That's a good point. I've had a stab at changing the approach to use and ENV var instead. If this looks ok, I'll start looking at the puppet bit.

kevindew (approved):
LGTM 👍

PR #1193, opened 2020-12-08
Keep existing user input on error

leenagupte:
Trello: https://trello.com/c/dYuQSgk7

## What?

Form fields are pre-populated with user's input when the create announcement form is submitted and fails validation. Previously if the user filled any of the fields out incorrectly, the submitted data disappeared for all of the other fields, even though those fields did not fail validation.

## Why?

Better user experience! It is annoying to have to fill the whole form out again if only one field was incorrect.

# Expected changes
### Before
Screenshot 2020-12-08 at 18 04 36

### After
Screenshot 2020-12-08 at 18 03 23

### Before
Screenshot 2020-12-08 at 18 04 15

### After
Screenshot 2020-12-08 at 18 03 03

### Before
Screenshot 2020-12-08 at 18 04 48

### After
Screenshot 2020-12-08 at 18 03 39

### DIFF
--- a/app/views/announcements/_form.html.erb
+++ b/app/views/announcements/_form.html.erb
@@ -4,15 +4,18 @@
     bold: true,
   },
   name: "announcement[title]",
+  value: @announcement.title,
   id: "title",
   error_message: error_items(@announcement.errors.messages, :title),
 } %>
+
 <%= render "govuk_publishing_components/components/input", {
   label: {
     text: "Enter the path to the announcement",
     bold: true,
   },
   name: "announcement[path]",
+  value: @announcement.path,
   id: "path",
   error_message: error_items(@announcement.errors.messages, :path),
   hint: "Example format: /government/news/coronavirus-covid-19",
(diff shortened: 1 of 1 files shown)


app/views/announcements/_form.html.erb
```diff
@@ -4,29 +4,52 @@
...
   name: "announcement[published_at]",
   id: "published_at",
   legend_text: legend,
   hint: "For example, 31 3 2020",
   error_message: error_items(@announcement.errors.messages, :published_at),
+  items: [
+    {
+      name: "day",
+      width: 2,
+      value: @announcement.published_at&.day
```
DilwoarH, line 38: what will happen if there isn't a published date? does this fall back to nil?

leenagupte: Yes, it's doesn't show a value. The `&` is a safety navigation operator and is the equivalent of doing:
```
@announcement.published_at&.day if @announcement.published_at.present? && @announcement.published_at.day.present?
```

DilwoarH (approved):

PR #1195, opened 2020-12-09
Remove finder publisher

sihugh:
The last of the finders published from here was removed some time ago and now [redirects to gov.uk/search/all](https://www.gov.uk/api/content/government/organisations/latest).

Finders are generally published from search-api or specialist-publisher these days


benthorner (approved):
Nice ✨. I didn't realise finders had gotten into so many different publishing apps!

PR #1196, opened 2020-12-09
Add routes for translated pages of /eubusiness

sihugh:
~This should not be merged before adding v67.2.0 or greater of gds-api-adapters to this repo.~ DONE

The first commit simplifies the SpecialRoutePublisher class, as it doesn't need to be that flexible at the moment. Context is in that commit message.

I've removed the `public_updated_at` entry from the payload as this is set by default downstream.

`document_type` is "special_route" by default, which is probably fine for entirely hard-coded pages that are not referenced in search in any way. This means we can leave it out of the new configuration.

I've made the content_ids of each of the translations the same, as this indicates that they differ only by their locale, and should be linked together as translations of each other should we wish to use this to help with our navigation between translations.

I've added a task to publish a single page, as this seems useful. Unpublishing them is more a moveable feast, so I haven't built anything for that yet. I think the most likely outcome is that we want to redirect to the English version.

The routes are test published as part of [the end to end tests](https://github.com/alphagov/publishing-e2e-tests/blob/19d76bdf8699b83b2eb706d5501fdd26d10eef01/Makefile#L153)


lib/special_route_publisher.rb
```diff
@@ -1,38 +1,83 @@
...
+        base_path: "/eubusiness.de",
+        locale: "de",
+        title: "Trade with the UK from 1 January 2021 as a business based in the EU",
+        description: "The UK has left the EU. On 31 December 2020 the UK will leave the EU single market and customs union. From 1 January 2021 the rules for trading with the UK will change.",
+      },
+      {
+        content_id: "bb986a97-3b8c-4b1a-89bf-2a9f46be9747",
+        base_path: "/eubusiness.es",
+        locale: "es",
+        title: "Trade with the UK from 1 January 2021 as a business based in the EU",
```
hannako, line 50: Should these be the translated titles and descriptions? I can push the translations to this branch if so.

hannako (approved):

lib/special_route_publisher.rb
sihugh: They can be, and it's preferable that they are so that we're consistent with use of the locale. I don't think they'll be used anywhere yet though, so it doesn't need to block us. Thanks!

sihugh:
As the person who opened the PR, I can't actually click the button to approve @hannako 's additions. But I approve them in spirit, so merge away. ✅

PR #1202, opened 2020-12-10
Edit announcement

Rosa-Fox:
[Trello](https://trello.com/c/MLHLiI8L/800-edit-an-announcement-using-the-ui)

- Announcements are editable via the user interface.
- Draft content is sent to the publishing api.

edit-announcement

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -30,6 +30,20 @@ class AnnouncementsController < ApplicationController
     redirect_to coronavirus_page_path(@coronavirus_page.slug), message
   end
 
+  def edit
+    @announcement = coronavirus_page.announcements.find(params[:id])
+  end
+
+  def update
+    @announcement = coronavirus_page.announcements.find(params[:id])
+
+    if @announcement.update(announcement_params) && draft_updater.send
+      redirect_to coronavirus_page_path(@coronavirus_page.slug), notice: "Announcement was successfully updated."
+    else
+      render :edit
+    end
+  end
+
 private
 
   def coronavirus_page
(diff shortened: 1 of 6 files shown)


spec/controllers/announcements_controller_spec.rb
```diff
@@ -55,6 +56,59 @@
     end
   end
 
+  describe "GET /coronavirus/:coronavirus_page_slug/announcements/:id/edit" do
+    before do
+      stub_user.permissions << "Unreleased feature"
```
leenagupte, line 61: I think we should a test or two to make sure user can't access the page if they don't have the right permissions.

app/controllers/announcements_controller.rb
```diff
@@ -25,7 +25,7 @@ def edit
   def update
     @announcement = Announcement.find(params[:id])
     @coronavirus_page = @announcement.coronavirus_page
-    if @announcement.update(announcement_params)
+    if @announcement.update(announcement_params) && draft_updater.send
```
leenagupte, line 28: Nice atomic commit 🤗

leenagupte (commented):
I've tested this locally and it looks good 🎉 

There are a couple of inline comments with minor suggestions and a request for a couple of permissions tests.

app/controllers/announcements_controller.rb
Rosa-Fox: 🥇 :)

spec/controllers/announcements_controller_spec.rb
Rosa-Fox: Done :)

app/controllers/announcements_controller.rb
leenagupte: Oh dear, your nice atomic commit got a bit broken in the rebase. Nevermind!

Rosa-Fox: oh damn it!!

leenagupte (approved):
LGTM 👍

PR #1204, opened 2020-12-11
Make announcements section live

leenagupte:
Trello: https://trello.com/c/ekjW492n

# What's changed and why?

We're ready to make the announcements section, so we need to remove the "Unreleased feature" feature flag.

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -1,6 +1,5 @@
 class AnnouncementsController < ApplicationController
   include DateFormatHelper
-  before_action :require_unreleased_feature_permissions!
   before_action :require_coronavirus_editor_permissions!
   layout "admin_layout"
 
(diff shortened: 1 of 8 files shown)


Rosa-Fox (commented):
Hi @leenagupte,


Only thing I found are that unreleased feature permissions are still being used in [this test](https://github.com/alphagov/collections-publisher/blob/49d192294ed77d37596317d65b05e650da067400/spec/controllers/announcements_controller_spec.rb#L104) but once that is gone then should be all good to approve.

leenagupte:
> Hi @leenagupte,
> 
> Only thing I found are that unreleased feature permissions are still being used in [this test](https://github.com/alphagov/collections-publisher/blob/49d192294ed77d37596317d65b05e650da067400/spec/controllers/announcements_controller_spec.rb#L104) but once that is gone then should be all good to approve.

Thank you! Fixing now.

Rosa-Fox (approved):

PR #1214, opened 2020-12-17
Tidy up specialist publisher tests

hannako:
## What
The `#publish` test in special_route_publisher_spec.rb is very similar to the spec for the publishing_api rake tasks in lib/tasks/publishing_api_spec.rb. As both tests stub the same call, the [CI](https://ci.integration.publishing.service.gov.uk/job/collections-publisher/job/master/1061/console) is intermittently failing. 

- Refactor the rake task spec to check if publishing api receives the call to put and publish, rather than assert that the specialist publisher receives it.
- Add a test for find_route to the special_route_publisher_spec.rb

### DIFF
--- a/spec/lib/special_route_publisher_spec.rb
+++ b/spec/lib/special_route_publisher_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 require_relative "../../lib/special_route_publisher"
 
 RSpec.describe SpecialRoutePublisher do
-  describe "publish" do
+  describe "#publish" do
     it "calls 'publish' on SpecialRoutePublisher with a route" do
       route = {
         document_type: "answer",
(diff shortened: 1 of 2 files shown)


spec/lib/special_route_publisher_spec.rb
```diff
@@ -2,33 +2,6 @@
 require_relative "../../lib/special_route_publisher"
 
 RSpec.describe SpecialRoutePublisher do
-  describe "publish" do
```
leenagupte, line 5: I think it might be better to remove the test for the rake task, and perhaps add a test for `find_route` in here instead, so that tests are together?

hannako: yep. uno momento!

leenagupte (approved):
👍

PR #1215, opened 2020-12-17
Remove need for at least one live stream in db

leenagupte:
Trello: https://trello.com/c/1YohbtSc

# What's changed and why?
Adds a guard so that the merging of live stream data doesn't fail if the
database doesn't contain a live stream.

The live stream is an optional component on the coronavirus landing
page, and there hasn't been a live stream in months.

However the code expected there to be at least one live stream in the
database.

Adding the guard means that we no longer have to create an unnecessary
live stream object to test things like announcements and sub_sections.

### DIFF
--- a/app/services/coronavirus_pages/content_builder.rb
+++ b/app/services/coronavirus_pages/content_builder.rb
@@ -94,6 +94,8 @@ class CoronavirusPages::ContentBuilder
   end
 
   def persisted_live_stream_data
+    return {} unless LiveStream.any?
+
     live_stream = LiveStream.last
     {
       "video_url" => live_stream.url,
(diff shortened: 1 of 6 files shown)


kevindew (commented):
Change looks good - should there be a unit test for ContentBuilder that covers this condition?

leenagupte:
@kevindew This was merged by accident. Not sure how I managed it, without one approving review 🤔

kevindew:
> @kevindew This was merged by accident. Not sure how I managed it, without one approving review 🤔

Oh yes that is strange, the repo setting seem to state that you need one approving review 🤔

PR #1230, opened 2021-01-08
Change TimelineEntry#contents type and validate heading length

kevindew:

We created this model with a DB column type of string, which in MySQL equates to a VARCHAR of max length 255. This causes us a problem as the current value we have in the timeline content is 679 characters. To resolve this I've switched the type to be text which has a max length of 64 KB.

I felt that there wasn't a real need for the heading field to be greater than 255 characters, so I haven't changed that. However it did seem a sufficiently small length that it could catch a user out, so I've added a validation. I didn't think there was much chance of a user being caught out with the 64KB content field so I didn't add a validation there.


leenagupte (approved):
💯

PR #1235, opened 2021-01-12
Add create timeline page

leenagupte:
Trello: https://trello.com/c/Gqe4lgaI

# What?

Create a page to allow new timeline entries to be stored in the database. 

Changes the markdown editor component to allow users to see the full range of menu options, whilst keeping the existing menu restrictions on the "step" and "sub-section" edit pages.

Only users with "Unreleased feature" permissions should be able to access the page whilst the timeline feature is being built.

The timeline entry is only persisted to the database. Whilst we are calling `DraftUpdater`, to send changes to the publishing-api, the actual creation of the payload for publishing-api will happen in another story.

# Expected changes

URL of page: http://collections-publisher.dev.gov.uk/coronavirus/landing/timeline_entries/new

screenshot-collections-publisher dev gov uk-2021 01 12-10_15_33

Error messages:

Screenshot 2021-01-13 at 17 15 59

### DIFF
--- a/app/controllers/timeline_entries_controller.rb
+++ b/app/controllers/timeline_entries_controller.rb
@@ -7,9 +7,27 @@ class TimelineEntriesController < ApplicationController
     @timeline_entry = coronavirus_page.timeline_entries.new
   end
 
+  def create
+    @timeline_entry = coronavirus_page.timeline_entries.new(timeline_entry_params)
+
+    if @timeline_entry.save && draft_updater.send
+      redirect_to coronavirus_page_path(coronavirus_page.slug), notice: "Timeline entry was successfully created."
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
 private
+
+  def timeline_entry_params
+    params.require(:timeline_entry).permit(:heading, :content)
+  end
 
   def coronavirus_page
     @coronavirus_page ||= CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
+  end
+
+  def draft_updater
+    @draft_updater ||= CoronavirusPages::DraftUpdater.new(coronavirus_page)
   end
 end
--- a/app/controllers/timeline_entries_controller.rb
+++ b/app/controllers/timeline_entries_controller.rb
@@ -1,6 +1,15 @@
 class TimelineEntriesController < ApplicationController
   before_action :require_coronavirus_editor_permissions!
   before_action :require_unreleased_feature_permissions!
+  layout "admin_layout"
 
-  def new; end
+  def new
+    @timeline_entry = coronavirus_page.timeline_entries.new
+  end
+
+private
+
+  def coronavirus_page
+    @coronavirus_page ||= CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
+  end
 end
--- a/app/controllers/timeline_entries_controller.rb
+++ b/app/controllers/timeline_entries_controller.rb
@@ -1,0 +1,6 @@
+class TimelineEntriesController < ApplicationController
+  before_action :require_coronavirus_editor_permissions!
+  before_action :require_unreleased_feature_permissions!
+
+  def new; end
+end
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -24,6 +24,11 @@
           <md-link class="app-c-markdown-editor__toolbar-button" title="Link" aria-label="Link">
             <%= render "components/markdown_editor/link.svg" %>
           </md-link>
+          <% if controls.include?(:blockquote) %>
+            <md-quote class="app-c-markdown-editor__toolbar-button" title="Blockquote" aria-label="Blockquote" data-gtm="markdown-toolbar-blockquote" data-gtm-click-tracking="true">
+              <%= render "components/markdown_editor/blockquote.svg" %>
+            </md-quote>
+          <% end %>
           <% if controls.include?(:bullets) %>
             <md-unordered-list class="app-c-markdown-editor__toolbar-button" title="Bullets" aria-label="Bullets">
               <%= render "components/markdown_editor/bullets.svg" %>
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -29,6 +29,11 @@
               <%= render "components/markdown_editor/blockquote.svg" %>
             </md-quote>
           <% end %>
+          <% if controls.include?(:numbered_list) %>
+            <md-ordered-list class="app-c-markdown-editor__toolbar-button" title="Numbered list" aria-label="Numbered list" data-gtm="markdown-toolbar-list" data-gtm-click-tracking="true">
+              <%= render "components/markdown_editor/numbered_list.svg" %>
+            </md-ordered-list>
+          <% end %>
           <% if controls.include?(:bullets) %>
             <md-unordered-list class="app-c-markdown-editor__toolbar-button" title="Bullets" aria-label="Bullets">
               <%= render "components/markdown_editor/bullets.svg" %>
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -1,7 +1,7 @@
 <%
   textarea ||= {}
   component_classes = %w[app-c-markdown-editor govuk-form-group]
-  bullets_button ||= false
+  controls ||= []
 %>
 <%= tag.div class: component_classes do %>
 
@@ -16,7 +16,7 @@
           <md-link class="app-c-markdown-editor__toolbar-button" title="Link" aria-label="Link">
             <%= render "components/markdown_editor/link.svg" %>
           </md-link>
-          <% if bullets_button %>
+          <% if controls.include?(:bullets) %>
             <md-unordered-list class="app-c-markdown-editor__toolbar-button" title="Bullets" aria-label="Bullets">
               <%= render "components/markdown_editor/bullets.svg" %>
             </md-unordered-list>
--- a/app/views/components/_markdown_editor.html.erb
+++ b/app/views/components/_markdown_editor.html.erb
@@ -2,12 +2,21 @@
   textarea ||= {}
   component_classes = %w[app-c-markdown-editor govuk-form-group]
   controls ||= []
+  error_message ||= ""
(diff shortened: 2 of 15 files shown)


app/views/timeline_entries/_form.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    name: "timeline_entry[content]",
+    id: "content",
+    rows: 20,
+    value: @timeline_entry.content
+  },
+  heading_buttons: true,
+  blockquote_button: true,
+  numbered_list_button: true,
+  bullets_button: true
+} %>
```
kevindew, line 27: It looks like this is missing the error display that is shown for the heading field.

For example: 

Screenshot 2021-01-12 at 11 06 25

app/views/timeline_entries/new.html.erb
```diff
@@ -1 +1,35 @@
...
+    },
+    {
+      text: "Add timeline entry"
+    },
+  ]
+
+  metadata = {
+    "Status" => "Draft",
+    "Last saved" => format_full_date_and_time(DateTime.now),
+  }
```
kevindew, line 19: It doesn't look like this metadata variable is used - unless I'm missing something?

app/views/timeline_entries/new.html.erb
```diff
@@ -1 +1,35 @@
...
+<% content_for :breadcrumbs, render('shared/steps/step_breadcrumb', links: links) %>
+<% content_for :title, formatted_title(@coronavirus_page) %>
+<% content_for :context, "Add timeline entry" %>
+<% if @timeline_entry.errors.any? %>
+  <%= render "shared/sub_sections/form_errors", resource: @timeline_entry %>
+<% end %>
+
+<div class="covid-edit-page govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @timeline_entry, announcement: @timeline_entry, url: coronavirus_page_timeline_entries_path do  %>
```
kevindew, line 31: I don't think this "announcement" argument does anything, is it something caught in a copy & paste from announcements?

kevindew: I did wonder if you need the url and timeline_entry referenced here and discovered a TIL in the Rails manual.

It looks like `form_for` is soft deprecated and the conventional Rails way would be to use [`form_with`](https://edgeguides.rubyonrails.org/form_helpers.html). So this should probably be `<%= form_with url: coronavirus_page_timeline_entries_path do %>`

app/controllers/timeline_entries_controller.rb
```diff
@@ -7,9 +7,27 @@ def new
     @timeline_entry = coronavirus_page.timeline_entries.new
   end
 
+  def create
+    @timeline_entry = coronavirus_page.timeline_entries.new(timeline_entry_params)
+
+    if @timeline_entry.save && draft_updater.send
+      redirect_to coronavirus_page_path(coronavirus_page.slug), notice: "Timeline entry was successfully created."
+    else
+      render :new
```
kevindew, line 16: this should probably have a status of unprocessable entity to indicate that validation failed - but maybe don't bother if it's a nightmare for consistency with the rest of the app.

kevindew (commented):
Looking great 👍  I've got a few minor comments here, nothing too significant to resolve.

spec/components/markdown_editor_spec.rb
```diff
@@ -22,9 +22,9 @@
     assert_select ".govuk-textarea[id='markdown-editor']"
   end
 
-  it "does not render bullet list toolbar button if hide_bullets_button is true" do
+  it "renders bullet list toolbar button if bullets_button is true" do
     render "components/markdown_editor",
-           hide_bullets_button: true,
+           bullets_button: true,
```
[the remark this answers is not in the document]
leenagupte: Yeah, I didn't know about it either until I tried to make this change.

spec/controllers/timeline_entries_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      get :new, params: { coronavirus_page_slug: coronavirus_page.slug }
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it "cannot be accessed by users without Coronavirus editor permissions" do
+      stub_user.permissions = ["signin", "Unreleased feature"]
+      get :new, params: { coronavirus_page_slug: coronavirus_page.slug }
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
```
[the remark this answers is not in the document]
leenagupte: I'll change this to set the permissions explicitly.

app/views/timeline_entries/_form.html.erb
leenagupte: Ah! It looks like this is missing in the component. I'll add it.

app/views/timeline_entries/new.html.erb
leenagupte: I think you're right. I'll remove it.

app/views/timeline_entries/new.html.erb
leenagupte: You caught me with the copy and paste error. Oops.

leenagupte: Hmmm. The form stops submitting when I switch to a `form_with`.

app/views/components/_markdown_editor.html.erb
```diff
@@ -3,6 +3,7 @@
   component_classes = %w[app-c-markdown-editor govuk-form-group]
   heading_buttons ||= false
   blockquote_button ||= false
+  numbered_list_button ||= false
```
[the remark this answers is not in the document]
leenagupte: My initial thought was to have all of the controls turned on by default, but I couldn't get it to work with the default set to "true", the options were always on. That's why I went with this approach of having to turn on the toolbar options you want.

I used the booleans approach because that seemed to follow the existing pattern for components.  So are you suggesting, having say `controls ||= [:headings, :blockquotes]` set here, so in this example, if you don't pass in controls, you'll see the heading and blockquotes buttons, but that if you want to see more, you have to explicitly pass them in? e.g. `controls: [:headings, :blockquotes, :numbered_list]`?

app/views/timeline_entries/new.html.erb
leenagupte: Ah! Figured it out. I need to set `local: true`

kevindew: Oh gosh `local: true` sounds a strange attribute 🤔

app/views/components/_markdown_editor.html.erb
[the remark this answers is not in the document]
kevindew: Yup, that sounds a good approach to me. Can do that in a separate PR if you like to not block this one?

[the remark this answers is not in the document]
leenagupte: I've got a separate commit ready. Will push that first to check everything is alright before splitting it up and fixing up.

leenagupte:
@kevindew Thanks for your review. I've addressed all of your comments. As mentioned in one of the replies, the change to passing a list of optional controls to the markdown editor is in a separate commit to check the approach is ok before splitting it up.

I've also switched to using `form_with` and tried to explain in the commit message why `local` needs to be set to true.

I've also added a feature test that I forgot to add the first time. Oops.

If get a chance can you re-review?

app/views/timeline_entries/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<% content_for :breadcrumbs, render('shared/steps/step_breadcrumb', links: links) %>
+<% content_for :title, formatted_title(@coronavirus_page) %>
+<% content_for :context, "Add timeline entry" %>
+<% if @timeline_entry.errors.any? %>
+  <%= render "shared/sub_sections/form_errors", resource: @timeline_entry %>
+<% end %>
+
+<div class="covid-edit-page govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @timeline_entry,
```
kevindew, line 26: I don't think this model does anything - since we reference the [model directly in the form](https://github.com/alphagov/collections-publisher/pull/1235/files#diff-0ed636e7df998d6718cbb06b6a880f21c098ce3020e9bed952165b9018b01794R21). So I think we can remove this without consequences.

app/views/timeline_entries/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+                  url: coronavirus_page_timeline_entries_path,
+                  local: true do %>
```
kevindew, line 28: I did some reading about this, and wow what a super annoying quirk! The good news is that Rails recognised this is weird too and in [Rails 6.1 this is going to default to be a local form](https://github.com/rails/rails/pull/40946) and you have to specify remote to change that.

Rather than specifying `local` here on this form I'd suggest we change config/application.rb to have `config.action_view.form_with_generates_remote_forms = false` so that this configuration is applied globally rather than needed on individual forms. We can leave a note with that to say we can remove that line once we've adopted Rails 6.1 config.

app/views/timeline_entries/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    bold: true
+  },
+  textarea: {
+    name: "timeline_entry[content]",
+    id: "content",
+    rows: 20,
+    value: @timeline_entry.content
+  },
+  error_message: error_items(@timeline_entry.errors.messages, :content),
+  controls: [:headings, :blockquote, :numbered_list, :bullets]
```
kevindew, line 24: Yay - I like this 👍

kevindew (commented):
Looking great Leena, just a couple of things that I think we can tweak with the form_with

app/views/timeline_entries/new.html.erb
leenagupte: Ah, yes, I think this was one of the many options I tried to get the form to submit before stumbling on the `local` issue. I'll remove it.

app/views/timeline_entries/new.html.erb
leenagupte: This is much nicer! I'll add a comment to config explaining why it's there.

leenagupte:
@kevindew Thanks for your comments. I've addressed them and fixed up. Could you re-re-review?

kevindew (approved):
Nice one 👍

PR #1251, opened 2021-01-20
Add rake task to sync timeline entries

JonathanHallam:
This adds a rake task to sync timeline entries with the database. Borrows heavily from [previous work](https://github.com/alphagov/collections-publisher/commit/116f59fcdd61e8bbd3a4a301e8ed366fee03b7ab) to do similar work for announcements.

Trello - https://trello.com/c/Hlr7PxjA/1050-migrate-timeline-entries-from-govuk-coronavirus-content-to-collections-publisher

### DIFF
--- a/app/services/coronavirus_pages/timeline_entry_builder.rb
+++ b/app/services/coronavirus_pages/timeline_entry_builder.rb
@@ -1,0 +1,27 @@
+class CoronavirusPages::TimelineEntryBuilder
+  def create_timeline_entries
+    return if timeline_entries_from_yaml.empty?
+
+    TimelineEntry.transaction do
+      coronavirus_page.timeline_entries.delete_all
+
+      timeline_entries_from_yaml.reverse.each do |entry|
+        coronavirus_page.timeline_entries.create!(
+          heading: entry["heading"],
+          content: entry["paragraph"],
+        )
+      end
+    end
+  end
+
+private
+
+  def timeline_entries_from_yaml
+    @github_data ||= YamlFetcher.new(coronavirus_page.raw_content_url).body_as_hash
+    @github_data.dig("content", "timeline", "list")
+  end
+
+  def coronavirus_page
+    @coronavirus_page ||= CoronavirusPage.find_by(slug: "landing")
+  end
+end
--- a/lib/tasks/coronavirus.rake
+++ b/lib/tasks/coronavirus.rake
@@ -23,4 +23,11 @@ namespace :coronavirus do
     puts "title: #{page.title}"
     puts "sections_title: #{page.sections_title}"
   end
+
+  desc "Sync timeline entries with entries in the yaml file"
+  task sync_timeline_entries: :environment do
+    CoronavirusPages::TimelineEntryBuilder.new.create_timeline_entries
+
+    puts "Timeline Entries synced for coronavirus landing page..."
+  end
 end
--- a/spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
+++ b/spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
@@ -1,0 +1,67 @@
+require "rails_helper"
+
+RSpec.describe CoronavirusPages::TimelineEntryBuilder do
+  let(:fixture_path) { Rails.root.join "spec/fixtures/coronavirus_landing_page.yml" }
+  let(:github_content) { YAML.safe_load(File.read(fixture_path)) }
+
+  let(:coronavirus_page) { create(:coronavirus_page, :landing) }
+
+  before do
+    stub_request(:get, Regexp.new(coronavirus_page.raw_content_url))
+      .to_return(status: 200, body: github_content.to_json)
+  end
+
+  it "creates new timeline entries from the yaml file in reverse order" do
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.count).to eq(2)
+
+    first_entry = coronavirus_page.timeline_entries.first
+
+    expect(first_entry.heading).to eq(github_entry_list.second["heading"])
+    expect(first_entry.content).to eq(github_entry_list.second["paragraph"])
+  end
+
+  it "removes existing timeline entries before creating new ones" do
+    create(:timeline_entry, coronavirus_page: coronavirus_page)
+    create(:timeline_entry, coronavirus_page: coronavirus_page)
+    create(:timeline_entry, coronavirus_page: coronavirus_page)
+
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.count).to eq(2)
+  end
+
+  it "doesn't remove existing timeline entries if the YAML file is empty" do
+    described_class.new.create_timeline_entries
+
+    github_content = YAML.safe_load(File.read(fixture_path))
+    github_content.dig("content", "timeline", "list").clear
+
+    stub_request(:get, Regexp.new(coronavirus_page.raw_content_url))
+      .to_return(status: 200, body: github_content.to_json)
+
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.count).to be(2)
+  end
+
+  it "keeps the timeline entries in the right order" do
+    new_data = { "heading" => "heading", "paragraph" => "paragraph" }
+
+    github_content = YAML.safe_load(File.read(fixture_path))
+    github_content.dig("content", "timeline", "list").unshift(new_data)
+
+    stub_request(:get, Regexp.new(coronavirus_page.raw_content_url))
+      .to_return(status: 200, body: github_content.to_json)
+
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.last.heading).to eq("heading")
+    expect(coronavirus_page.timeline_entries.last.position).to eq(1)
+  end
+
+  def github_entry_list
+    github_content.dig("content", "timeline", "list")
+  end
+end


app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,27 @@
+class CoronavirusPages::TimelineEntryBuilder
+  def create_timeline_entries
+    if timeline_entries_from_yaml.any?
+      Announcement.transaction do
```
leenagupte, line 4: Copy and paste error: This transaction should be on `TimelineEntry`.

leenagupte (commented):
I've tried this locally and this all works like a charm. 🎉 

There's only one request. Can a test be added to check that timeline entries are being added in the expected order? 
Adding the test here means that we'll know to update it when the [positioning PR](https://github.com/alphagov/collections-publisher/pull/1253) is added. Or conversely if the positioning PR goes in first, then we can be confident that timeline entries from the YAML are being given the correct position.

JonathanHallam:
@leenagupte sure thing, will add it in in the morning :)

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+  def create_timeline_entries
+    if timeline_entries_from_yaml.any?
+      TimelineEntry.transaction do
+        coronavirus_page.timeline_entries.delete_all
+
+        timeline_entries_from_yaml.each do |entry|
+          TimelineEntry.create!(
+            heading: entry["heading"],
+            content: entry["paragraph"],
+            position: entry["position"],
```
kevindew, line 11: This isn't actually going to do anything because of: https://github.com/alphagov/collections-publisher/blob/3f8ee18910a0507a7824ae26b552de1d65942411/app/models/timeline_entry.rb#L5

It also looks like the yaml file doesn't have a position

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,29 @@
+class CoronavirusPages::TimelineEntryBuilder
+  def create_timeline_entries
+    if timeline_entries_from_yaml.any?
+      TimelineEntry.transaction do
+        coronavirus_page.timeline_entries.delete_all
+
+        timeline_entries_from_yaml.each do |entry|
```
kevindew, line 7: I think this will need to add the items in reverse order to get the positioning correct because of the callback. As per: https://github.com/alphagov/collections-publisher/pull/1253/files#diff-8b6d13a1a94024bdc57bf166e53622328267634808dbea57021a9e514a5b4ceeR71

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+          TimelineEntry.create!(
```
kevindew, line 8: Minor, this could be a little simpler with `coronavirus_page.timeline_entry.create!(` and you can save specifying the coronavirus_page.id.

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,29 @@
+class CoronavirusPages::TimelineEntryBuilder
+  def create_timeline_entries
+    if timeline_entries_from_yaml.any?
```
kevindew, line 3: Minor it'd be a more idiomatic Ruby to have this as a guard clause. E.g. `return if timeline_entries_from_yaml.empty?`

Do you definitely want to abort if there are no entries though or should it delete all items? I'm not sure why we're anticipate this being empty to know what we need to be cautious of.

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+      end
+    end
+  end
+
+private
+
+  def timeline_entries_from_yaml
+    @github_data ||= YamlFetcher.new(coronavirus_page.raw_content_url).body_as_hash
+    @github_data["content"]["timeline"]["list"]
```
kevindew, line 23: Minor - this looks like a dig: https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig

lib/tasks/coronavirus.rake
```diff
@@ -23,4 +23,11 @@ namespace :coronavirus do
     puts "title: #{page.title}"
     puts "sections_title: #{page.sections_title}"
   end
+
+  desc "Sync timeline entries with entries in the yaml file"
+  task sync_timeline_entries: :environment do
+    CoronavirusPages::TimelineEntryBuilder.new.create_timeline_entries
+
+    puts "Timeline Entries synced for coronavirus landing page..."
+  end
```
kevindew, line 32: I'm not sure if it breaks convention, but it'd be good to have a test for the rake task - to be honest the amount of code here is sufficiently short it'd probably be reasonable for this to be written in the rake task, however I'm not sure if that breaks a convention.

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    github_content = YAML.safe_load(File.read(fixture_path))
+    github_content["content"]["timeline"]["list"].unshift(new_data)
+
+    stub_request(:get, Regexp.new(coronavirus_page.raw_content_url))
+      .to_return(status: 200, body: github_content.to_json)
+
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.first.heading).to eq("heading")
+    expect(coronavirus_page.timeline_entries.first.position).to eq(3)
```
kevindew, line 66: I don't think this test is right as unshift inserts this at the start of the array - whereas this tests that the item has the last position.

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    entry = create(:timeline_entry, coronavirus_page: coronavirus_page)
+
+    described_class.new.create_timeline_entries
+
+    first_entry = coronavirus_page.timeline_entries.first
+
+    expect(coronavirus_page.timeline_entries.count).to eq(1)
+    expect(first_entry.heading).to eq(entry.heading)
+    expect(first_entry.content).to eq(entry.content)
+    expect(first_entry.created_at).to eq(entry.created_at)
```
kevindew, line 51: I don't think you need this much in the way of assertions here. It'd be sufficient to confirm the collection isn't empty. E.g. `expect(coronavirus_page.timeline_entries).not_to be_empty`

kevindew (commented):
Nice one Jon, this is looking clean - I think there's a problem with positioning to resolve though.

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: yep, absolutely right, rushed this in, should have checked, sorry

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: i did try this, having copied from https://github.com/alphagov/collections-publisher/commit/116f59fcdd61e8bbd3a4a301e8ed366fee03b7ab couldn't get it working for the life of me. Just get an error along the lines of 
`undefined method `timeline_entry' for #<CoronavirusPage:0x00005629a55ac1f0>`

kevindew: Oops I typo'd it's `coronavirus_page.timeline_entries` (plural not singular)

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: yep, excellent shout

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: nice one, sorted thanks :)

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
JonathanHallam: I've fixed this, it is expected to be able to add things to the start and have them written to the last position. That's so that the newest entries can be added to the top of the yaml whilst the oldest end up in position 1. The tests now reflect this.

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
JonathanHallam: nice one, have cleaned this test up, thanks

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: yeah, took a while to get my head round the ordering of things, sorry. This is sorted now.

lib/tasks/coronavirus.rake
JonathanHallam: So I've basically just copied this from [here](https://github.com/alphagov/collections-publisher/blob/e47d09b50f80942e9db7eb30200c5fffe5524d0e/lib/tasks/coronavirus.rake#L31) hoping to follow convention :D. I think the rake task test should probably be added to another story if we feel strongly about it because I don't think I've added enough here to warrant it

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: This was also just copied from some [work that Leena already did](https://github.com/alphagov/collections-publisher/blob/e47d09b50f80942e9db7eb30200c5fffe5524d0e/app/services/coronavirus_pages/announcements_builder.rb#L3). I think it's just to protect us from losing data if someone removes all the entries from the yaml, Leena might know a bit more as to why this is desirable. I've changed the syntax of the guard clause.

app/services/coronavirus_pages/timeline_entry_builder.rb
JonathanHallam: sorted :)

lib/tasks/coronavirus.rake
kevindew: Sure, sounds like there isn't a convention for testing rake tasks. That's a shame but doesn't seem worth introducing one for this given it's only a temporary task.

app/services/coronavirus_pages/timeline_entry_builder.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    TimelineEntry.transaction do
+      coronavirus_page.timeline_entries.delete_all
+
+      timeline_entries_from_yaml.reverse.each do |entry|
+        coronavirus_page.timeline_entries.create!(
+          heading: entry["heading"],
+          content: entry["paragraph"],
+        )
+      end
```
kevindew, line 13: Nice one - the logic for this looks correct now.

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    stub_request(:get, Regexp.new(coronavirus_page.raw_content_url))
+      .to_return(status: 200, body: github_content.to_json)
+
+    described_class.new.create_timeline_entries
+
+    expect(coronavirus_page.timeline_entries.last.heading).to eq("heading")
+    expect(coronavirus_page.timeline_entries.last.position).to eq(1)
+  end
+
+  def github_entry_list
```
kevindew, line 64: Minor - there's a couple of other places in the code that could use this method (lines 39 and 53)

kevindew (approved):
This looks good - I've tried it out locally and it works 🎉 

It looks like some of the code here could be [replaced by utilising the helper Leena added](https://github.com/alphagov/collections-publisher/blob/f588d9da3c3f05b498ed385cecdddf22c1508f56/spec/support/coronavirus_helpers.rb#L2), but I'm assuming this will only live for a short period of time so there isn't too much of a need to optimise it.

spec/services/coronavirus_pages/timeline_entry_builder_spec.rb
JonathanHallam: Ah nice spot, should have seen this. I've been able to use it on 39 but because `github_content` has been redefined on line 52 trying to use it on 53 breaks things.

kevindew: Oh sorry - hadn't quite appreciated that it was using the let (not sure what I was thinking) I imagine like 39 also does something undesirable as that'll have a different `github_content` in scope to this (isn't variable scope fun?)

JonathanHallam: oh blimey yes good point, I'll undo that one too! Changed things which made the tests pass but for all the wrong reasons! All this and it'll be the version you reviewed that actually gets merged :D

PR #1253, opened 2021-01-21
Add new timeline entries front of list

leenagupte:
Trello: https://trello.com/c/0oeC49ub

# What's changed and why?

Add new timeline entries to the front of the list (i.e. position 1) so that content designers can be sure that the order of the entries they see in the publishing tool matches the order of the timeline entries shown to the user.

For example, a content designer might add an entry for  "2 January", and later add an entry for "24 - 26 January". By default we need to display these entries in this order:

"24 -26 January"
"2 January"

A page to allow users to reorder timeline entries will be added in a later PR.

Paired with @Rosa-Fox 

### DIFF
--- a/app/models/coronavirus_page.rb
+++ b/app/models/coronavirus_page.rb
@@ -16,4 +16,10 @@ class CoronavirusPage < ApplicationRecord
       announcement.update!(position: index)
     end
   end
+
+  def make_timeline_entry_positions_sequential
+    timeline_entries.sort_by(&:position).each.with_index(1) do |timeline_entry, index|
+      timeline_entry.update!(position: index)
+    end
+  end
 end
--- a/app/models/coronavirus_page.rb
+++ b/app/models/coronavirus_page.rb
@@ -12,14 +12,18 @@ class CoronavirusPage < ApplicationRecord
   end
 
   def make_announcement_positions_sequential
-    announcements.sort_by(&:position).each.with_index(1) do |announcement, index|
-      announcement.update!(position: index)
-    end
+    make_positions_sequential(announcements)
   end
 
   def make_timeline_entry_positions_sequential
-    timeline_entries.sort_by(&:position).each.with_index(1) do |timeline_entry, index|
-      timeline_entry.update!(position: index)
+    make_positions_sequential(timeline_entries)
+  end
+
+private
+
+  def make_positions_sequential(collection)
+    collection.sort_by(&:position).each.with_index(1) do |object, index|
+      object.update_column(:position, index)
     end
   end
 end
--- a/app/models/timeline_entry.rb
+++ b/app/models/timeline_entry.rb
@@ -2,4 +2,10 @@ class TimelineEntry < ApplicationRecord
   belongs_to :coronavirus_page
   validates :heading, presence: true, length: { maximum: 255 }
   validates :content, presence: true
+  before_create :set_position
+
+  def set_position
+    coronavirus_page.timeline_entries.update_all("position = position + 1")
+    self.position = 1
+  end
 end
--- a/app/models/timeline_entry.rb
+++ b/app/models/timeline_entry.rb
@@ -3,9 +3,14 @@ class TimelineEntry < ApplicationRecord
   validates :heading, presence: true, length: { maximum: 255 }
   validates :content, presence: true
   before_create :set_position
+  after_destroy :set_parent_positions
 
   def set_position
     coronavirus_page.timeline_entries.update_all("position = position + 1")
     self.position = 1
+  end
+
+  def set_parent_positions
+    coronavirus_page.make_timeline_entry_positions_sequential
   end
 end
--- a/app/services/coronavirus_pages/draft_discarder.rb
+++ b/app/services/coronavirus_pages/draft_discarder.rb
@@ -63,7 +63,7 @@ module CoronavirusPages
     end
 
     def update_timeline_entries
-      coronavirus_page.timeline_entries.destroy_all
+      coronavirus_page.timeline_entries.delete_all
       coronavirus_page.timeline_entries = timeline_entries
     end
 
--- a/app/services/coronavirus_pages/draft_discarder.rb
+++ b/app/services/coronavirus_pages/draft_discarder.rb
@@ -68,11 +68,10 @@ module CoronavirusPages
     end
 
     def timeline_entries
-      timeline_entries_from_payload.each_with_index.map do |attributes, index|
+      timeline_entries_from_payload.reverse.map do |attributes|
         TimelineEntry.new(
           heading: attributes[:heading],
           content: attributes[:paragraph],
-          position: index + 1,
         )
       end
     end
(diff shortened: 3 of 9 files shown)


app/models/timeline_entry.rb
```diff
@@ -2,4 +2,14 @@ class TimelineEntry < ApplicationRecord
...
+  after_create :set_position
+
+  def set_position
+    coronavirus_page.timeline_entries.each do |timeline_entry|
+      new_position = timeline_entry.position.present? ? timeline_entry.position + 1 : 1
+      timeline_entry.update!(position: new_position)
+    end
+
+    update!(position: 1)
+  end
```
kevindew, line 15: There's an aspect of dubiousness that we briefly persist an invalid entry into the database by allowing nil. I wonder if this might be better as a before_save block where you set the position as 1 and update_all the other ones to be one greater? It has a small risk of not being in a transaction that you wouldn't necessarily have the item persisted, but then you have a similar risk in an after_create that the callback falls.

app/models/coronavirus_page.rb
```diff
@@ -11,15 +11,9 @@ def topic_page?
...
-      announcement.update!(position: index)
-    end
-  end
-
-  def make_timeline_entry_positions_sequential
-    timeline_entries.sort_by(&:position).each.with_index(1) do |timeline_entry, index|
-      timeline_entry.update!(position: index)
+  def make_positions_sequential(collection)
+    collection.sort_by(&:position).each.with_index(1) do |object, index|
+      object.update!(position: index)
```
kevindew, line 16: Nice refactoring 👍

kevindew (commented):
Nice job. I've got a suggestion for the callback handling, but it's not too big a deal if it's not to your tastes.

I wonder if the thing that made this challening is the enforcement of trying to keep a strict sequence in single increments? I imagine the tricky thing about that is that it's a somewhat unusual concept in databases where the state of one row is coupled to a whole collection,  where every create, update or delete operation requires the whole series to be updated.

It does strike me that being more forgiving with the meaning of the sequence is the only real simplicity trade off available. For example if you weren't concerned about it being incremented by 1 then you'd not have to do anything when things were deleted and accept keys like 1, 3, 5. If you were prepared to have it be ordered via a descending key then you'd be also be ablt to simplify creating leaving the only ordering management on updates.

However, this seems like all the reasonable steps to maintain a strict ordering so looks good to me.

app/models/timeline_entry.rb
```diff
@@ -2,4 +2,14 @@ class TimelineEntry < ApplicationRecord
   belongs_to :coronavirus_page
   validates :heading, presence: true, length: { maximum: 255 }
   validates :content, presence: true
+  after_create :set_position
+
+  def set_position
+    coronavirus_page.timeline_entries.each do |timeline_entry|
+      new_position = timeline_entry.position.present? ? timeline_entry.position + 1 : 1
+      timeline_entry.update!(position: new_position)
+    end
```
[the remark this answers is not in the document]
leenagupte: Thanks! This is much nicer. I don't think adding a new timeline entry should update the "updated_at" time of the existing entries if it's just the position that's changed, so this is better.

Will update and fix up.

leenagupte:
@kevindew Thanks for your review. I've made some changes and fixed up.

>If you were prepared to have it be ordered via a descending key then you'd be also be able to simplify creating leaving the only ordering management on updates.

The tricky part with showing things in reverse order comes in the[ upcoming reorder page](https://github.com/alphagov/collections-publisher/pull/1256). If we show things in descending order, what does it mean when a user moves an entry up one position?

app/models/timeline_entry.rb
kevindew: Ah sorry I meant `before_create`

app/models/timeline_entry.rb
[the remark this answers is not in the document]
kevindew: Great stuff - a thing to bear in mind is that this will cause an imbalance with https://github.com/alphagov/collections-publisher/pull/1253/files#diff-abcc24eb29072787841a2fdc3b350199c9a9123251ca8d8b406631c22bbe4979R16 which will update the timetamp. You can change that to `update_column(:position, index)` to avoid that.

app/models/timeline_entry.rb
leenagupte: If I switched to using a `before_create`, `set_position` could be changed to:

```
  def set_position
    coronavirus_page.timeline_entries.update_all("position = position + 1")
    self.position = 1
  end
```

Is that what you're thinking? And then I can switch the position column back to not accepting nils.

kevindew:
> @kevindew Thanks for your review. I've made some changes and fixed up.
> 
> > If you were prepared to have it be ordered via a descending key then you'd be also be able to simplify creating leaving the only ordering management on updates.
> 
> The tricky part with showing things in reverse order comes in the[ upcoming reorder page](https://github.com/alphagov/collections-publisher/pull/1256). If we show things in descending order, what does it mean when a user moves an entry up one position?

Sure, well I understand that the JS just works out the orderings on the fly rather than using the data from the DB so it's already somewhat abstract - and thus could be done without any JS changes. You'd likely need the controller to flip the ordering along the lines of:

```
    base_position = coronavirus_page.timeline_entries.count + 1
    reordered_timeline_entries.each do |timeline_entry_data|
      timeline_entry = coronavirus_page.timeline_entries.find(timeline_entry_data["id"])
      timeline_entry.update!(position: base_position - timeline_entry_data["position"])
    end
```

app/models/timeline_entry.rb
kevindew: Yup - that's it. Hopefully I've not missed something that makes that renders that untenable.

leenagupte:
@kevindew Thanks for your review. I've made the changes to use the `before_create` callback instead of `after_create`, and to be consistent about not setting `updated_at` if just the position is changing. Can you re-re-review?

leenagupte:
> Sure, well I understand that the JS just works out the orderings on the fly rather than using the data from the DB so it's already somewhat abstract - and thus could be done without any JS changes. You'd likely need the controller to flip the ordering along the lines of:
> 
> ```
>     base_position = coronavirus_page.timeline_entries.count + 1
>     reordered_timeline_entries.each do |timeline_entry_data|
>       timeline_entry = coronavirus_page.timeline_entries.find(timeline_entry_data["id"])
>       timeline_entry.update!(position: base_position - timeline_entry_data["position"])
>     end
> ```

I haven't had a chance to look at this logic. But I think being this could end up being a bit of a brain-twister on re-order page.

kevindew (commented):
Nice one - this looks good. Sorry to drop a late comment in.

app/models/coronavirus_page.rb
```diff
@@ -11,9 +11,9 @@ def topic_page?
     slug == "landing"
   end
 
-  def make_announcement_positions_sequential
-    announcements.sort_by(&:position).each.with_index(1) do |announcement, index|
-      announcement.update!(position: index)
+  def make_positions_sequential(collection)
```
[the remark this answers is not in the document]
leenagupte: You're right. This is better. I've updated and fixed up.

kevindew (approved):
Great stuff 👍

PR #1254, opened 2021-01-22
Add rescue condition for featured link description

GDSNewt:
At the moment we're pulling out the featured link descriptions from the content item to display in the action link. For now we are content for external links to be featured without this description so we can just add a rescue condition to circumvent any 404s we receive from the publishing-api. Adding/editing meta descriptions for non-govuk featured links can be added later if we determine there is a need for them. 

With this condition the test shows that we can build a new subsection with the description being nil

[Trello](https://trello.com/c/a8aT8ExJ/1051-allow-featured-links-in-accordions-to-link-to-external-links)

### DIFF
--- a/app/presenters/sub_section_json_presenter.rb
+++ b/app/presenters/sub_section_json_presenter.rb
@@ -131,9 +131,15 @@ class SubSectionJsonPresenter
   end
 
   def description_for_featured_link(base_path)
+    return nil if external_link?(base_path)
+
     content_item(base_path)["description"]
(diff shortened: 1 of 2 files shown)


leenagupte:
@GDSNewt It would be great if you could update the commit message with any other approaches you tried, and why they were discarded. I can't review this, because it is identical to the Spike PR I created as a proof of concept, and even then I wasn't sure if that was the best approach to take.

kevindew:
Superseded by https://github.com/alphagov/collections-publisher/pull/1259

PR #1258, opened 2021-01-28
Delete timeline entries

leenagupte:
Trello: https://trello.com/c/rwtOUXAZ

# What's changed?

Allow users to delete a timeline entry from the summary page.

Screenshot 2021-01-28 at 09 53 31

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -17,16 +17,19 @@ class AnnouncementsController < ApplicationController
   end
 
   def destroy
-    @announcement = coronavirus_page.announcements.find(params[:id])
-    @coronavirus_page = @announcement.coronavirus_page
-    attrs = @announcement.attributes.except("id", "created_at", "updated_at")
-    if @announcement.destroy && draft_updater.send
-      message = { notice: "Announcement was successfully deleted." }
-    else
-      draft_updater.rebuild_announcement(attrs)
-      message = { alert: "Announcement couldn't be deleted" }
+    announcement = coronavirus_page.announcements.find(params[:id])
+    message = { notice: "Announcement was successfully deleted." }
+
+    Announcement.transaction do
+      announcement.destroy!
+
+      unless draft_updater.send
+        message = { alert: "Announcement couldn't be deleted" }
+        raise ActiveRecord::Rollback
+      end
     end
-    redirect_to coronavirus_page_path(@coronavirus_page.slug), message
+
+    redirect_to coronavirus_page_path(coronavirus_page.slug), message
   end
 
   def edit
--- a/app/controllers/sub_sections_controller.rb
+++ b/app/controllers/sub_sections_controller.rb
@@ -3,15 +3,13 @@ class SubSectionsController < ApplicationController
   layout "admin_layout"
 
   def new
-    @coronavirus_page = CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
-    @sub_section = @coronavirus_page.sub_sections.new
+    @sub_section = coronavirus_page.sub_sections.new
   end
 
   def create
-    @coronavirus_page = CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
-    @sub_section = @coronavirus_page.sub_sections.new(sub_section_params)
+    @sub_section = coronavirus_page.sub_sections.new(sub_section_params)
     if @sub_section.save && draft_updater.send
-      redirect_to coronavirus_page_path(@coronavirus_page.slug), notice: "Sub-section was successfully created."
+      redirect_to coronavirus_page_path(coronavirus_page.slug), notice: "Sub-section was successfully created."
     else
       @sub_section.errors.add :base, draft_updater.errors.to_sentence
       render :new
@@ -19,15 +17,13 @@ class SubSectionsController < ApplicationController
   end
 
   def edit
-    @coronavirus_page = CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
-    @sub_section = @coronavirus_page.sub_sections.find(params[:id])
+    @sub_section = coronavirus_page.sub_sections.find(params[:id])
   end
 
   def update
-    @sub_section = SubSection.find(params[:id])
-    @coronavirus_page = @sub_section.coronavirus_page
+    @sub_section = coronavirus_page.sub_sections.find(params[:id])
     if @sub_section.update(sub_section_params) && draft_updater.send
-      redirect_to coronavirus_page_path(@coronavirus_page.slug), notice: "Sub-section was successfully updated."
+      redirect_to coronavirus_page_path(coronavirus_page.slug), notice: "Sub-section was successfully updated."
     else
       @sub_section.errors.add :base, draft_updater.errors.to_sentence
       render :edit
@@ -35,8 +31,7 @@ class SubSectionsController < ApplicationController
   end
 
   def destroy
-    @sub_section = SubSection.find(params[:id])
-    @coronavirus_page = @sub_section.coronavirus_page
+    @sub_section = coronavirus_page.sub_sections.find(params[:id])
     attrs = @sub_section.attributes.except("id", "created_at", "updated_at")
     if @sub_section.delete && draft_updater.send
       message = { notice: "Sub-section was successfully deleted." }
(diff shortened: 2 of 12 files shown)


app/controllers/timeline_entries_controller.rb
```diff
@@ -31,6 +31,18 @@ def update
     end
   end
 
+  def destroy
+    @timeline_entry = coronavirus_page.timeline_entries.find(params[:id])
+
+    message = if @timeline_entry.destroy && draft_updater.send
+                { notice: "Timeline entry was successfully deleted." }
+              else
+                { alert: "Timeline entry couldn't be deleted" }
```
kevindew: Just wondered if these should be in the locale file? it's not so much a suggestion, just wondering what consistently is and isn't in it.

spec/support/coronavirus_feature_steps.rb
```diff
@@ -356,6 +356,21 @@ def then_the_timeline_entry_is_updated
...
+  set_up_github_data(@coronavirus_page)
+
+  page.accept_alert "Are you sure?" do
+    page.find("a[href=\"/coronavirus/landing/timeline_entries/#{@timeline_entry.id}\"]", text: "Delete").click
+  end
+end
+
+def then_i_can_see_the_timeline_entry_has_been_deleted
+  expect(page).to have_text("Timeline entry was successfully deleted.")
+  expect(page).not_to(have_text(@timeline_entry.heading))
```
kevindew, line 371:
```suggestion
  expect(page).not_to have_text(@timeline_entry.heading)
```

Just minor slip out of RSpec conventions

spec/support/coronavirus_feature_steps.rb
```diff
@@ -356,6 +356,21 @@ def then_the_timeline_entry_is_updated
   expect(@timeline_entry.reload.content).to eq("##Form content")
 end
 
+# Deleting timeline entries
+
+def when_i_delete_a_timeline_entry
+  set_up_github_data(@coronavirus_page)
+
+  page.accept_alert "Are you sure?" do
+    page.find("a[href=\"/coronavirus/landing/timeline_entries/#{@timeline_entry.id}\"]", text: "Delete").click
```
kevindew, line 365: This seems rather specific would `click_link("Delete")` work?

app/controllers/timeline_entries_controller.rb
```diff
@@ -33,12 +33,14 @@ def update
...
   def destroy
     @timeline_entry = coronavirus_page.timeline_entries.find(params[:id])
+    backup = @timeline_entry.attributes.except("id", "created_at", "updated_at")
 
-    message = if @timeline_entry.destroy && draft_updater.send
-                { notice: "Timeline entry was successfully deleted." }
-              else
-                { alert: "Timeline entry couldn't be deleted" }
-              end
+    if @timeline_entry.destroy && draft_updater.send
```
kevindew, line 38: It seems like this destroy would reset the ordering of everything and the recreate would then move this item to the end?

I imagine the right thing to do is to put things into a transaction and abort the transaction if updating the Publishing API failed so the data remains intact.

app/controllers/sub_sections_controller.rb
```diff
@@ -3,52 +3,51 @@ class SubSectionsController < ApplicationController
...
     end
-    redirect_to coronavirus_page_path(@coronavirus_page.slug), message
+    redirect_to coronavirus_page_path(coronavirus_page.slug), message
   end
 
 private
 
+  def coronavirus_page
+    @coronavirus_page ||= CoronavirusPage.find_by(slug: params[:coronavirus_page_slug])
+  end
```
kevindew, line 49: This is great 👍

kevindew (commented):
Just a few comments - the most painful one is likely the ordering one with the transaction suggestion, hopefully it's a good suggestion.

spec/support/coronavirus_feature_steps.rb
leenagupte: When the [Reorder page PR](https://github.com/alphagov/collections-publisher/pull/1256) is ready, they'll be more than one Delete link in the Timeline section, so I'm going to leave this in ready for the rebase.

app/controllers/timeline_entries_controller.rb
leenagupte: They should. I'll move them in separate commit. There needs to be another review of the content as I think some of it was added at the same time as locale file work.

app/controllers/timeline_entries_controller.rb
leenagupte: Good shout. I'll wrap the destroy and publishing-api calls in a transaction, and do the same for announcements and sub-sections too.

leenagupte:
@kevindew Thanks for your review. I've addressed and responded to your comments.

spec/controllers/timeline_entries_controller_spec.rb
```diff
@@ -145,6 +145,36 @@
...
+    end
+
+    it "redirects to the coronavirus page" do
+      delete :destroy,
+             params: {
+               id: timeline_entry.id,
+               coronavirus_page_slug: coronavirus_page.slug,
+             }
+
+      expect(subject).to redirect_to(coronavirus_page_path(coronavirus_page.slug))
```
kevindew, line 164:
```suggestion
      expect(response).to redirect_to(coronavirus_page_path(coronavirus_page.slug))
```

spec/controllers/timeline_entries_controller_spec.rb
```diff
@@ -173,6 +173,31 @@
...
+             params: {
+               id: timeline_entry.id,
+               coronavirus_page_slug: coronavirus_page.slug,
+             }
+
+      expect(coronavirus_page.timeline_entries.count).to eq(2)
+      expect(coronavirus_page.timeline_entries.first.position).to eq(2)
+      expect(coronavirus_page.timeline_entries.first.heading).to eq("Skywalker")
+      expect(coronavirus_page.timeline_entries.second.position).to eq(1)
+      expect(coronavirus_page.timeline_entries.second.heading).to eq("Amidala")
```
kevindew, line 199: This seems to be asserting on a bunch of things outside of this test and the logic in the code. 

You could probably simplify it down to:

```suggestion
      stub_publishing_api_isnt_available

      params = { id: timeline_entry.id, coronavirus_page_slug: coronavirus_page.slug }
      expect { delete :destroy, params: params }
         .not_to change { coronavirus_page.reload.timeline_entries.count }
         .from(2)
```

spec/controllers/announcements_controller_spec.rb
```diff
@@ -70,18 +68,16 @@
...
-      stub_any_publishing_api_put_content
-        .to_return(status: 500)
+    it "doesn't delete the announcement if draft_updater fails" do
+      stub_publishing_api_isnt_available
+
       expect { subject }.not_to(change { Announcement.count })
       expect(announcement.title).to eq Announcement.last.title
       expect(announcement.path).to eq Announcement.last.path
-      expect(announcement.id).not_to eq Announcement.last.id
+      expect(announcement.id).to eq Announcement.last.id
```
kevindew, line 80: I don't think you need these last three assertions now it's not recreated.

spec/controllers/announcements_controller_spec.rb
```diff
@@ -127,7 +123,6 @@
     end
 
     it "updates the announcements" do
-      announcement
```
kevindew, line 130: Yay - let! is much less mysterious than this variable usage.

app/controllers/announcements_controller.rb
```diff
@@ -18,15 +18,18 @@ def create
 
   def destroy
     @announcement = coronavirus_page.announcements.find(params[:id])
```
kevindew, line 20: Minor - I guess this isn't used in the view so doesn't need an @ variable?

kevindew (commented):
Looks great - I've left a few minor comments but I imagine you'll want to finish https://github.com/alphagov/collections-publisher/pull/1256 and rebase first.

spec/controllers/timeline_entries_controller_spec.rb
leenagupte: Ah, I was trying to assert that the position doesn't change, as if the timeline entry was being recreated rather than rolled back, the positions would reverse order.

kevindew (commented):
Looks good - just the question about consistency with flash messaging and locale file.

app/controllers/announcements_controller.rb
```diff
@@ -17,16 +17,19 @@ def create
...
-    @announcement = coronavirus_page.announcements.find(params[:id])
-    @coronavirus_page = @announcement.coronavirus_page
-    attrs = @announcement.attributes.except("id", "created_at", "updated_at")
-    if @announcement.destroy && draft_updater.send
-      message = { notice: "Announcement was successfully deleted." }
-    else
-      draft_updater.rebuild_announcement(attrs)
-      message = { alert: "Announcement couldn't be deleted" }
+    announcement = coronavirus_page.announcements.find(params[:id])
+    message = { notice: "Announcement was successfully deleted." }
```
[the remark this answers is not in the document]
leenagupte: Probably, but I do think this is outside the scope of this card. I'll create a card to do a "locale audit".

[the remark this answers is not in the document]
kevindew: Ok cool 👍  - the locale file conventions are complex.

kevindew (approved):

PR #1262, opened 2021-02-01
Add a Coronavirus namespace for Coronavirus related code

kevindew:

This applies a `Coronavirus` module to models and controllers that are associated with Coronavirus. This has been done to make this code distinct from the rest of the app - for example a model of `Announcement` feels like something generic that can be used anywhere, whereas `Coronavirus::Announcement` makes it clear it's only for `Coronavirus`. Hopefully this will also make the future removal of Coronavirus code easier.

As this is quite a delicate, tedious changes that are hard to review I've tried to break up this task into logical bits and only do the first of this in this PR. This means the state of the code following this PR will have a few non-ideal things. Following this PR I plan to do the following things:

- Put associated services and presenters into the same namespace
- Do something similar with LiveStreams, as they're Coronavirus specific but handled differently to everything else
- Rename `Coronavirus::CoronavirusPage` to `Coronavirus::Page`
- Rename factories so they're prefixed with `coronavirus`

A divisive thing to consider as part of this was what a namespace should be and the scope of it. I decided to go with `Coronavirus` since it matched the path of /coronavirus and so it could be broad enough for future Coronavirus code. This isn't without downsides though, it means that it's not clear announcements/sub-sections/timeline entries are related to a Page (an alternative namespacing could have been based on CoronavirusPage) and it makes it hard to directly replace the services namespaced under CoronavirusPages, as they need a bit of consideration as to how they're names still make sense.

This PR is best reviewed commit by commit with further details in the commit messages. A query string of [?w=1](https://github.com/alphagov/collections-publisher/commit/3f042086b3b61f41037198a8f7ca64891463fa86?w=1) will make https://github.com/alphagov/collections-publisher/commit/3f042086b3b61f41037198a8f7ca64891463fa86 much easier to review.


JonathanHallam:
From a code point of view this is looking nice, thank you so much for breaking the commits up in the way you have, it made looking at this nice and easy. I'm struggling to find things to say about it as none of the changes are that drastic and the commits are so clean. I'm happy with this from a code perspective given all of the future work that you've put in the commit message. I know Rosa wanted a look as well so I'll not press the green tick just yet. Thanks for this, makes our space a bit easier to understand :)

Rosa-Fox (approved):
Brilliant work @kevindew. I've looked through it all/run locally and can't find anything missing... other than potentially also name spacing the test fixtures? There are not many of them though and those which are coronavirus related (most of them) have `coronavirus` in their file name.

PR #1265, opened 2021-02-02
Rename CoronavirusPages module to Coronavirus::Pages

kevindew:

This is open as a draft as it depends on https://github.com/alphagov/collections-publisher/pull/1262. I've set that as the base branch for this PR

This change has been made so that these classes could be within the
Coronavirus namespace. These classes didn't seem to make much sense
without some relationship to page so I decided to create an additional
namespace of Pages with Coronavirus so they could be moved with minimal
changes.

There is probably scope to rethink the purpose of some of these classes
as they seem to venture outside of what would be considered to be a
service (Configuration) and there is a Presenter class here that seems
to have cross over with the presenters directory of classes.

### DIFF
--- a/app/controllers/coronavirus/announcements_controller.rb
+++ b/app/controllers/coronavirus/announcements_controller.rb
@@ -64,7 +64,7 @@ module Coronavirus
     end
 
     def draft_updater
-      @draft_updater ||= CoronavirusPages::DraftUpdater.new(@coronavirus_page)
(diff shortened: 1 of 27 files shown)


JonathanHallam (approved):
This looks good, I've run the tests locally and it's happy. Thanks so much for the GitHub `?w=1` modifier knowledge, made this so much easier!!

PR #1267, opened 2021-02-03
Put Live streams under Coronavirus namespace

kevindew:

This moves the various LiveStream code to be under a `Coronavirus` namespace to bring consistency with the other Coronavirus code.

As part of this it changes the route for `/live_stream` to now be `/coronavirus/live_stream`. It also renames the Youtube URL validator from `UrlValidator` to` YoutubeUrlValidator` and fixes some inconsistent code approaches in the tests.

More info in the commits.


JonathanHallam (approved):
This looks great, thank you :)

app/models/coronavirus/live_stream.rb
```diff
@@ -1,3 +1,3 @@
 class Coronavirus::LiveStream < ApplicationRecord
-  validates :url, url: true, presence: true
+  validates :url, youtube_url: true, presence: true
```
leenagupte, line 2: What's the difference between `url` and `youtube_url`? Or does this make `youtube_url` an alias of `url`?

kevindew: It represents the renaming of the underlying validator.

Prior to this change we had a validator, UrlValidator, which had the job of validating that a string was a YouTube URL. This was rather surprising as the name, UrlValidator, implies that it validates that a string is a URL. I renamed that to be YoutubeUrlValidator to reflect that it actually validates that a string is a YouTube URL.

PR #1271, opened 2021-02-05
Add coronavirus prefix to coronavirus database tables (⚠️  dangerous PR)

kevindew:

This makes these models consistent with `coronavirus_pages` in having a
`coronavirus_` prefix in the table name.

This migration must be applied at a quiet time for the application as
the deployment of it risks errors from when the migration has run until
a new instance of the app is running.

It's not ideal to deal with this risk of errors but the alternative is
far more painful - which would be creating new tables that are
backfilled and kept in sync until the tables are swapped. Given that
this application frequently goes many hours without a user request it
seems reasonable to take this risk.

### DIFF
--- a/app/models/coronavirus/timeline_entry.rb
+++ b/app/models/coronavirus/timeline_entry.rb
@@ -1,4 +1,6 @@
 class Coronavirus::TimelineEntry < ApplicationRecord
+  self.table_name = "coronavirus_timeline_entries"
+
   belongs_to :page, foreign_key: "coronavirus_page_id"
   validates :heading, presence: true, length: { maximum: 255 }
   validates :content, presence: true
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,9 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_26_122043) do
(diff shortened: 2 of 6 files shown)


db/schema.rb
```diff
@@ -221,6 +221,8 @@
     t.index ["uid"], name: "index_users_on_uid", unique: true
   end
 
+  add_foreign_key "coronavirus_sub_sections", "coronavirus_pages"
+  add_foreign_key "coronavirus_timeline_entries", "coronavirus_pages"
```
leenagupte, line 225: Hmmm. I wonder why announcements never had a foreign key? 🤔

app/models/coronavirus/timeline_entry.rb
```diff
@@ -1,4 +1,7 @@
 class Coronavirus::TimelineEntry < ApplicationRecord
+  self.table_name = "coronavirus_timeline_entries"
+
+  belongs_to :coronavirus_page
```
leenagupte, line 4: Should this be `belongs_to :page`?

kevindew: Yes it should, looks like a rebasing mistake but it's a bit of a worry that it all still built fine with that. Thanks for catching.

leenagupte (approved):
👍

PR #1275, opened 2021-02-09
Move hard-coded Coronavirus pages content from controllers and views to locale file

leenagupte:

# What?

Whilst the [locale file card](https://trello.com/c/0Nm1iIcz/973-put-content-for-coronavirus-publishing-tool-subsections-and-announcements-into-an-en-locale-file) was being worked on, more content was added to views and controllers.

This change moves does another audit of controllers, views etc, for sub-sections, announcements, and timeline entries and makes sure all of the content has been moved to the locale file.

## Content moved
- [x] Summary page views
- [x] Announcements controller
- [x] Announcements views
- [x] Reorder Announcements controller
- [x] Reorder Announcements views
- [x] Timeline entries controller
- [x] Timeline entries views
- [x] Reorder Timeline entries controller
- [x] Reorder Timeline entries views
- [x] Subsections controller
- [x] Subsections views
- [x] Reorder Subsections controller
- [x] Reorder Subsections views
- [x] Livestream controller
- [x] Livestream views
- [x] Pages controller
- [x] Feature tests


app/views/coronavirus/timeline_entries/_form.html.erb
```diff
@@ -11,7 +11,7 @@
 
 <%= render "components/markdown_editor", {
   label: {
-    text: "Content",
+    text: t("coronavirus.pages.timeline_entries.form.content"),
```
kevindew, line 14: I think this and the previous key would benefit from having a suffix of `_label` I thought they were heading and content of the form rather than references to form field labels.

app/views/coronavirus/reorder_sub_sections/index.html.erb
```diff
@@ -9,17 +9,17 @@
       href: coronavirus_page_path(slug: @page.slug)
     },
     {
-      text: "Reorder"
+      text: t("coronavirus.pages.sub_sections.reorder.title", section: @page.sections_title.downcase)
```
kevindew, line 12: Is it intentional that this doesn't match the previous text of "Reorder" ?

app/views/coronavirus/reorder_timeline_entries/index.html.erb
```diff
@@ -9,13 +9,13 @@
       href: coronavirus_page_path(slug: @page.slug)
     },
     {
-      text: t("coronavirus.pages.timeline_entries.reorder.heading")
+      text: t("coronavirus.pages.timeline_entries.reorder.title")
```
kevindew, line 12: ⭐  for consistency.

config/locales/en.yml
```diff
@@ -52,6 +52,13 @@ en:
         subtopic_edit:
           accordions: "Edit %{page_name} accordions"
           something: Edit something else on the
+      live_stream:
+        update:
+          success: Draft live stream url updated!
```
kevindew, line 96: Corr are we excited enough about this to give users an exclamation point?

config/locales/en.yml
```diff
@@ -53,6 +53,34 @@ en:
           accordions: "Edit %{page_name} accordions"
           something: Edit something else on the
       live_stream:
+        title: Update live stream URL
+        instructions:
+          one:
+            step: 1. Add the URL
+            summary:
+              title: Help with YouTube
+              content: |
```
kevindew, line 73:
```suggestion
              content_markdown: |
```

config/locales/en.yml
```diff
@@ -1,11 +1,22 @@
 en:
   coronavirus:
     pages:
+      error:
```
kevindew, line 4: Sorry another comment on key organisation. 

This seems a touch arbitrary to have just one section of error within coronavirus.pages when so much is in there - and there's probably a few other errors in there. Perhaps it could be under the action it's defined under `update` or alternative idea would be putting all flash messages for the controller under one header of `flashes`

config/locales/en.yml
```diff
@@ -62,7 +62,7 @@ en:
           something_else: Edit something else on the landing page
         subtopic_edit:
           accordions: "Edit %{page_name} accordions"
-          something: Edit something else on the
+          something: Edit something else on the %{page_name}
```
kevindew, line 65: Great catch on this missing variable 👍 

It seems a curious choice of the original implementer to call this `something` rather than `something_else` as that means the opposite.

kevindew: Oh and I just notice two lines above it's called `something_else` in a different context. 

I have an awkward feeling I may have just missed this in a review. Any chance you can fix? 🙏

spec/support/coronavirus_feature_steps.rb
```diff
@@ -583,11 +583,11 @@ def and_i_see_a_draft_updated_message
   end
 
   def and_i_choose_a_major_update
-    choose("Major")
+    choose(I18n.t("coronavirus.pages.publish.major_type"))
   end
 
   def and_i_publish_the_page
-    click_on("Publish")
+    click_on(I18n.t("coronavirus.pages.publish.button_text"))
```
kevindew, line 590: Nice one catching all of these missed bits 👍

kevindew (commented):
Good stuff Leena - I think I've only caught a couple of genuine problems.

Most of my concerns seem to be based around a somewhat arbitrary approach to how we're organising the keys for these locales. Might be easier to chat over that to clear that up?

app/controllers/coronavirus/announcements_controller.rb
```diff
@@ -11,21 +11,21 @@ def new
     def create
       @announcement = page.announcements.new(announcement_params)
       if @announcement.save && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: "Announcement was successfully created."
+        redirect_to coronavirus_page_path(page.slug), notice: I18n.t("coronavirus.pages.announcements.new.success")
```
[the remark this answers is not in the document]
leenagupte: I didn't move the announcements section in the locale file. It was always under "pages". I left it there because a page has announcements. But I don't think the pages controller and page model are related like that are they?

Also, I used `new.success` because I thought of it in terms of what the user is doing, i.e. adding a "new" announcement.

config/locales/en.yml
leenagupte: I didn't change the error message! That is a debate for another time.

app/views/coronavirus/reorder_sub_sections/index.html.erb
leenagupte: Yeah, I followed the pattern of the other pages where the metadata text matches the page title.

app/controllers/coronavirus/announcements_controller.rb
```diff
@@ -11,21 +11,21 @@ def new
...
       announcement = page.announcements.find(params[:id])
-      message = { notice: "Announcement was successfully deleted." }
+      message = { notice: I18n.t("coronavirus.pages.announcements.delete.success") }
 
       Announcement.transaction do
         announcement.destroy!
 
         unless draft_updater.send
-          message = { alert: "Announcement couldn't be deleted" }
+          message = { alert: I18n.t("coronavirus.pages.announcements.delete.failed") }
```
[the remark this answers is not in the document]
leenagupte: Ah, I went with matching the user action instead.

app/controllers/coronavirus/reorder_announcements_controller.rb
```diff
@@ -24,10 +24,10 @@ def update
       end
 
       if success
-        message = "Announcements were successfully reordered."
+        message = I18n.t("coronavirus.pages.announcements.reorder.success")
         redirect_to coronavirus_page_path(page.slug), notice: message
       else
-        message = "Sorry! Announcements have not been reordered: #{draft_updater.errors.to_sentence}."
+        message = I18n.t("coronavirus.pages.announcements.delete.failed", error: draft_updater.errors.to_sentence)
```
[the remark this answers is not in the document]
leenagupte: Ah good spot! You caught me in a copy and paste error. I can add a commit later that `!`'s the coronavirus content. Thanks.

app/views/coronavirus/reorder_announcements/index.html.erb
```diff
@@ -9,17 +9,17 @@
...
 %>
 <% content_for :breadcrumbs, render('shared/steps/step_breadcrumb', links: links) %>
 <% content_for :title, formatted_title(@page)%>
-<% content_for :context, "Reorder announcements" %>
+<% content_for :context, t("coronavirus.pages.announcements.reorder.title") %>
 
 <div class="covid-manage-page govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render_markdown "Use the up/down buttons on the right to reorder the announcements, or click and hold on a section to reorder using drag and drop." %>
+    <%= render_markdown t("coronavirus.pages.announcements.reorder.form.markdown_hint") %>
```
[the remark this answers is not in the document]
leenagupte: I didn't delete and content. I think the content itself is a discussion for another time.

config/locales/en.yml
```diff
@@ -7,9 +7,10 @@ en:
         preview: Preview
         publish: Publish
       announcements:
-        confirm: Are you sure?
-        reorder: Reorder
-        add: Add announcement
+        summary:
+          title: Announcements
+          reorder: Reorder
+          add: Add announcement
```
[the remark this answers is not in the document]
leenagupte: "Pages" doesn't really mean anything. I think something that shows that this is content on the summary page would be more helpful. `coronavirus.summary.announcements`? Then everything on the summary page would be in one place?

app/controllers/coronavirus/reorder_announcements_controller.rb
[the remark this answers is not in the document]
leenagupte: We're already setting [config.action_view.raise_on_missing_translations = true](https://github.com/alphagov/collections-publisher/blob/master/config/application.rb#L27) 🎉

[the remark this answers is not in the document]
kevindew: Oh great - you might find it reads nicer in controllers to use the helper from the view with `helpers.t` rather than `I18n.t!` it feel a bit more consistent with the views (but would still need I18n for tests)

kevindew:
As per our discussion I had a play around with moving the translation keys so the key always mapped logically back to the place it was used (view or controller).

The commit is here with more details in the commit message: https://github.com/alphagov/collections-publisher/commit/5a377ece92fc8ff1dfb4fcf341161a1bce26b3ec

Main thing I'm looking for with the change above is to see if we can establish agreement on a) a logical mapping is beneficial and scalable and b) is the one I've proposed good. It doesn't have to go in with this PR.

app/views/coronavirus/reorder_announcements/index.html.erb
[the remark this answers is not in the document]
leenagupte: It's not in a component yet... so perhaps: `t("coronavirus.pages.announcements.reorder.hint_markdown")`

config/locales/en.yml
leenagupte: This is the only error that applies to the whole controller as it's triggered by a before action. All of the other errors are action specific so I'm inclined to leave this where it is.

app/controllers/coronavirus/reorder_announcements_controller.rb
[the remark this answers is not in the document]
leenagupte: I've never seen `helpers.t` used anywhere else. Is that a new standard?

config/locales/en.yml
kevindew: Sure, I guess I find it confusing just due to how broad the key of coronavirus.pages is, given that spans multiple controllers.

app/controllers/coronavirus/reorder_announcements_controller.rb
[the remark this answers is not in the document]
kevindew: In Rails 5 [the ability to call helpers in controllers](https://bigbinary.com/blog/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers) was added. So instead of having to do `include MyHelper` you can access them directly from a helper object.

There isn't any specific standard for using `helpers.t` I'm just theorising that this would be an easier read as there is symmetry between controllers with `helpers.t("key") and views with `<%= t("key") %>`. I feel that having `I18n.t!("key") in controllers, while views have `<%= t("key") %>`, makes you wonder why one is a bang and the other isn't.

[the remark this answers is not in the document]
leenagupte: Do we still need the `!` if we're already raising an error on missing translation?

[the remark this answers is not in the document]
kevindew: Yup. `I18n.t` doesn't make any use of `config.action_view.raise_on_missing_translations` that configuration value only applies when you use Rails' `t` helper.

leenagupte:
@kevindew Thanks for your review. I've had a look at restructuring the content in the locale file. What I've found out is that a lot of the confusion comes from how the content for "Edit something else" page (i.e. the "prepare" page) and the index page was spread throughout the file. I also found out that none of the view partials in the "shared" folder were actually being shared. So as the view and controller structure is still a bit all over the place, I've tried structuring the content by "screen". i.e. the page that the user sees. So for example,  the content is grouped under:

- index page
- summary page
- new timeline entry page
- edit announcement page
- reorder sub-sections page
- livestream page
- edit something else page
etc

Hopefully this will make it easy to go straight from the page needing to be updated to the section in the locale page that contains the content.

This means that there are a few places where content is duplicated. For example, the button text for "Publish" appears in three places, but I thought that it was preferable to keep the livestream, prepare and summary pages separate from each other so that the content is easier to find, rather than trying to DRY them up and making a mess.

Also, the only partials that are being shared are the forms for the new and edit pages for announcements, timeline entries and sub-sections. I've put them in as their own entry for now as this seemed like the tidiest option.

kevindew:
Thanks for this @leenagupte and great job spotting the shared folder isn't actually used for any sharing. Nice one on the `helpers.t` too.

I've been trying to understand the screen organisation system you've introduced. There are a few bits that have confused me:

-  if things are screens then [what is a form](https://github.com/alphagov/collections-publisher/blob/222e36a2dd18ac827cea73754e7bd5e4e223018c/config/locales/en.yml#L49-L51)? should all shared things have their own keys?
- it seems to mix internal terminology with external terminology - for example we have a keys for sub_sections (which are known externally as accordions) but then renames to prepare to [something_else_from_github](https://github.com/alphagov/collections-publisher/blob/222e36a2dd18ac827cea73754e7bd5e4e223018c/config/locales/en.yml#L133) - and introduce new terminology such as `coronavirus.summary`
- I don't really understand the logic behind the flash message organisation, some seem to be [defined based on the screen they'll be seen on](https://github.com/alphagov/collections-publisher/pull/1275/files#diff-f5f9513cfc84e9e7b0679941c0cd03a7eb4d7bae232c7d13ae84d51ea29b0dbfR22), [others not](https://github.com/alphagov/collections-publisher/pull/1275/files#diff-f5f9513cfc84e9e7b0679941c0cd03a7eb4d7bae232c7d13ae84d51ea29b0dbfR44), some [seem to be embedded in actions keys](https://github.com/alphagov/collections-publisher/blob/222e36a2dd18ac827cea73754e7bd5e4e223018c/config/locales/en.yml#L14-L18)

I do wonder if it might have been better to put together a quick proposal of this new approach and discuss before applying it everywhere, as I feel pretty lousy about asking questions on the fundamentals when there are 16 commits of implementation - nor can I imagine that you want to be questioned after doing all that. I'm also assuming, given you've taken a new approach, that you didn't like the idea of doing this in an approach that was consistent with Content Publisher? 

> This means that there are a few places where content is duplicated. For example, the button text for "Publish" appears in three places, but I thought that it was preferable to keep the livestream, prepare and summary pages separate from each other so that the content is easier to find, rather than trying to DRY them up and making a mess.

Yup, repeating is fine. It seems much less dangerous than having the same key used in multiple places and having someone think they're only editing it in one place.

leenagupte:
I didn't follow the content-publisher of naming keys after controllers (is that right?) as that assumes that the controller/view patterns make sense, and I don't think that's true for coronavirus pages.  Also I had a look at your example commit, but couldn't see that pattern replicated (using "controller" and "view" in the key names) in content-publisher. Also it tightly couples content to code rather than where it's seen, which I don't think is a good idea for coronavirus pages.

As I mentioned previously, I find PagesController confusing. From looking at it I can't tell what each action really does, so I didn't want to perpetuate that confusion in the locale file. For example, the PagesController contains actions for the index page, summary pages, and "something else" page, but not the live-stream page or the sub-section pages (I'm excluding announcements and timeline entries from this as they follow the pattern established for sub-sections). I believe that PagesController needs to be broken up and the "something else" stuff should be moved out into their own controller etc. However, I think that should be done in a separate PR as the content changes in this one have already snowballed at bit.

The aim is group content by what the user sees and what they were doing. i.e. what page does the content appear on? Which action errored etc. So the "delete subsection" action happens on the summary page, so it and it's associated messages (success, failed) appear there. Edit subsection has it's own page, so everything associated with what the user is doing on that page is grouped together, even the success message even though that's shown on the summary page.

For complex pages like [summary pages](https://collections-publisher.integration.publishing.service.gov.uk/coronavirus/landing), the content is grouped by the section it appears on the page, hence the "actions" key.

To fully understand what bits of the code the content related to, I had to pull apart the locale file. I resaved it in 16 commits, because saving it way that was easier for me. _I_ need a visual to understand what the restructure would look like, hence the restructuring of the file before pushing it up for a discussion.

The original scope of this PR was just to make sure all of the announcements, sub-sections and timeline entries content was moved to the locale file, but these changes have now far exceeded that. 

I think rather than continuing with this PR, the changes should be broken up into separate pieces. So rather than doing this, I think for example, the "prepare"/"something else" page should be renamed and split out from the PagesController, and it's content and views consolidated (I don't think it needs to be split into 5 partials). Ditto the live_stream changes, as these seem to be the least contentious. Then the announcements, timeline entries and sub-sections can be updated in their own individual PRs rather than in one go. It will be easier to focus on these sections once the noise of the  "prepare"/"something else" and live stream fixes have been removed.

kevindew:
> I didn't follow the content-publisher of naming keys after controllers (is that right?) as that assumes that the controller/view patterns make sense, and I don't think that's true for coronavirus pages. Also I had a look at your example commit, but couldn't see that pattern replicated (using "controller" and "view" in the key names) in content-publisher. Also it tightly couples content to code rather than where it's seen, which I don't think is a good idea for coronavirus pages.

Yeah, the content publisher approach is to have the keys correlate with the view structure. If you have a view of [app/views/image/edit.html.erb](https://github.com/alphagov/content-publisher/blob/master/app/views/images/edit.html.erb) then you have locale keys prefixed with "en.image.edit" and that correlates to a file stored at [config/locales/view/edit.yml](https://github.com/alphagov/content-publisher/blob/master/config/locales/en/images/edit.yml).

The key -> view naming structure is a common Rails convention, in fact (In googling this) I just learnt that they actually have a shorthand functionality that makes use of this: https://guides.rubyonrails.org/i18n.html#lazy-lookup. It actually seems to be that we initially followed that convention in this app with the [initial commit adding it](https://github.com/alphagov/collections-publisher/commit/c6c760a1e6b23d497fbaad93a66032cbc0d64040) but looks like the subsequent commits lost it.

I think there are always going to be scenarios where locale file mapping is challenging to understand and I think the key advantage you get with doing it based on file structures is that there isn't ambiguity to this organisation (which I see is how we got into this problem in the first place).

> As I mentioned previously, I find PagesController confusing. From looking at it I can't tell what each action really does, so I didn't want to perpetuate that confusion in the locale file. For example, the PagesController contains actions for the index page, summary pages, and "something else" page, but not the live-stream page or the sub-section pages (I'm excluding announcements and timeline entries from this as they follow the pattern established for sub-sections). I believe that PagesController needs to be broken up and the "something else" stuff should be moved out into their own controller etc. However, I think that should be done in a separate PR as the content changes in this one have already snowballed at bit.
> 
> The aim is to group content by what the user sees and what they were doing. i.e. what page does the content appear on? Which action errored etc. So the "delete subsection" action happens on the summary page, so it and it's associated messages (success, failed) appear there. Edit subsection has it's own page, so everything associated with what the user is doing on that page is grouped together, even the success message even though that's shown on the summary page.
> 
> For complex pages like [summary pages](https://collections-publisher.integration.publishing.service.gov.uk/coronavirus/landing), the content is grouped by the section it appears on the page, hence the "actions" key.
> 
> To fully understand what bits of the code the content related to, I had to pull apart the locale file. I resaved it in 16 commits, because saving it way that was easier for me. _I_ need a visual to understand what the restructure would look like, hence the restructuring of the file before pushing it up for a discussion.

Thanks, that all makes sense. The situation I'm a bit worried about is that this new structure is that it's somewhat arbitrary organisation so it may just slip back into the same problems of before. Maybe it won't? 

> The original scope of this PR was just to make sure all of the announcements, sub-sections and timeline entries content was moved to the locale file, but these changes have now far exceeded that.
> 
> I think rather than continuing with this PR, the changes should be broken up into separate pieces. So rather than doing this, I think for example, the "prepare"/"something else" page should be renamed and split out from the PagesController, and it's content and views consolidated (I don't think it needs to be split into 5 partials). Ditto the live_stream changes, as these seem to be the least contentious. Then the announcements, timeline entries and sub-sections can be updated in their own individual PRs rather than in one go. It will be easier to focus on these sections once the noise of the "prepare"/"something else" and live stream fixes have been removed.

Cool, that sounds good 👍

leenagupte:
Closing as this PR is going to be broken up into smaller chunks.

PR #1281, opened 2021-02-11
Fix components guide

DilwoarH:
This change fixes the components guide endpoint in collections publisher. There was a bug in the yaml file which stopped the guide from rendering, this has now been fixed and a small test added to make sure the components guide loads.

### DIFF
--- a/app/views/components/docs/markdown_editor.yml
+++ b/app/views/components/docs/markdown_editor.yml
@@ -28,7 +28,8 @@ examples:
         id: markdown-editor
   with_bullet_list_button:
     data:
-      controls: [:bullets]
+      controls:
+        - :bullets
       label:
         text: Body
       textarea:
@@ -36,7 +37,8 @@ examples:
         id: markdown-editor
   with_headings_buttons:
     data:
-      controls: [:headings]
+      controls:
+        - :headings
       label:
         text: Body
       textarea:
@@ -44,7 +46,8 @@ examples:
         id: markdown-editor
   with_blockquote_button:
     data:
-      controls: [:blockquote]
+      controls:
+        - :blockquote
       label:
         text: Body
       textarea:
@@ -52,7 +55,8 @@ examples:
         id: markdown-editor
   with_numbered_list_button:
     data:
-      controls: [:numbered_list]
+      controls:
+        - :numbered_list
       label:
         text: Body
       textarea:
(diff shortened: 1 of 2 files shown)


app/views/components/docs/markdown_editor.yml
```diff
@@ -28,31 +28,35 @@ examples:
         id: markdown-editor
   with_bullet_list_button:
     data:
-      controls: [:bullets]
+      controls:
+        - :bullets
```
kevindew, line 32: Nice one fixing these 👍

spec/features/components_guide_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.feature "Visit Component Guide" do
```
kevindew, line 3: Great stuff adding a test for this.

I was a bit surprised to have it as a feature test but since there isn't a concept of requests tests for this project (instead has the older approach of controller tests) then I agree feature is the place to put this stuff.

spec/features/components_guide_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  include ComponentGuideSteps
```
kevindew, line 4: It seems superfluous to have a helper file for this. I'd just add any methods into here (as per: https://github.com/alphagov/collections-publisher/blob/master/spec/features/access_control_spec.rb#L42-L62)

spec/features/components_guide_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  describe "Component Guide" do
```
kevindew, line 6: I don't think you need a describe

spec/features/components_guide_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    scenario "User views the component guide page" do
+      when_i_visit_the_component_guide_index_page
+      i_see_a_page_title
+      i_see_subheadings
```
kevindew, line 10: These two assertion seem to actually test parts of govuk_publishing_components gem by checking for HTML and text which the gem provides. 

I'd probably just do:
```
  when_i_visit_the_component_guide_index_page
  then_i_receive_a_valid_response
end

# ...

def then_i_receive_a_valid_response
  expect(page.status_code).to eq(200)
end
```

DilwoarH: Done

spec/features/components_guide_spec.rb
DilwoarH: Done

spec/features/components_guide_spec.rb
DilwoarH: Done

kevindew (approved):
Nice one 👍

PR #1285, opened 2021-02-17
Subsection bug

GDSNewt:
In the previous configuration we had accidentally allowed the user to save changes to a sub-section even if there were errors with it. This wraps the sub-section changes in a transaction whilst we call on the draft updater, so they can be rolled back in the event that the draft updater finds errors.

There will be no visual changes for the user 

[Trello card](https://trello.com/c/p6Rq4Gj8/163-stop-saving-subsections-with-invalid-markdown-the-band-aid-method)

### DIFF
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -23,11 +23,17 @@ module Coronavirus
 
     def update
       @sub_section = page.sub_sections.find(params[:id])
-      if @sub_section.update(sub_section_params) && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: "Sub-section was successfully updated."
-      else
+
+      SubSection.transaction do
+        @sub_section.update!(sub_section_params)
+        raise ActiveRecord::Rollback unless draft_updater.send
+      end
+
+      if draft_updater.errors.any?
         @sub_section.errors.add :base, draft_updater.errors.to_sentence
         render :edit
+      else
+        redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully updated." }
       end
     end
 
--- a/spec/controllers/coronavirus/sub_sections_controller_spec.rb
+++ b/spec/controllers/coronavirus/sub_sections_controller_spec.rb
@@ -144,6 +144,15 @@ RSpec.describe Coronavirus::SubSectionsController do
       expect(sub_section.title).to eq(title)
       expect(sub_section.content).to eq(content)
     end
+
+    context "given invalid content" do
+      let(:content) { "invalid content" }
+
+      it "doesn't change the subsection" do
+        expect { subject }.not_to(change { sub_section.reload.attributes })
+        expect(subject.body).to include("Unable to parse markdown")
+      end
+    end
   end
 
   describe "DELETE /coronavirus/:page_slug/sub_sections/:id" do


spec/controllers/coronavirus/sub_sections_controller_spec.rb
```diff
@@ -141,8 +144,28 @@
...
+            title: "valid title",
+            content: "invalid content",
+          },
+        }
+      end
+
+      it "doesn't change the subsection" do
+        subject
+        sub_section.reload
+        expect(sub_section.content).not_to eq("invalid content")
```
kevindew, line 166: We can probably use a more concise means to confirm nothing has changed which matches closer to the `it` description.

```suggestion
        expect { subject }.not_to(change { sub_section.reload.attributes })
```

kevindew (commented):
This looks great, you've got a good handle on the problem here.

I've made a few suggestions that can make this more concise.

spec/controllers/coronavirus/sub_sections_controller_spec.rb
```diff
@@ -124,7 +124,10 @@
       {
         id: sub_section,
         page_slug: page.slug,
-        sub_section: sub_section_params,
+        sub_section: {
+          title: "valid title",
+          content: "###valid content",
+        },
```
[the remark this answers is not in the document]
JonathanHallam: I've had a bit of a look at this and you're right, I think we jumped the gun. In my head the title and content were being used to create all the subsections which would have meant that the test didn't actually represent a change. This isn't actually the case as most of them use the factory which has its own faked out content. We'll just drop this commit.

spec/controllers/coronavirus/sub_sections_controller_spec.rb
```diff
@@ -141,8 +144,28 @@
...
+      let(:params) do
+        {
+          id: sub_section,
+          page_slug: page.slug,
+          sub_section: {
+            title: "valid title",
+            content: "invalid content",
+          },
+        }
+      end
```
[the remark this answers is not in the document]
JonathanHallam: you absolutely can, thanks!

spec/controllers/coronavirus/sub_sections_controller_spec.rb
JonathanHallam: nice one, i like it :)

app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -23,11 +23,22 @@ def edit
 
     def update
       @sub_section = page.sub_sections.find(params[:id])
-      if @sub_section.update(sub_section_params) && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: "Sub-section was successfully updated."
-      else
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+      errors = nil
```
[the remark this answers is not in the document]
JonathanHallam: nice one, thanks, should have spotted this! Have fixed this and all the other related comments

spec/controllers/coronavirus/sub_sections_controller_spec.rb
```diff
@@ -144,6 +144,17 @@
       expect(sub_section.title).to eq(title)
       expect(sub_section.content).to eq(content)
     end
+
+    context "given invalid content" do
+      let(:content) { "invalid content" }
+
+      it "doesn't change the subsection" do
+        subject
+        sub_section.reload
```
kevindew, line 153: These two lines aren't needed (well the first one `subject` actually renders the assertion on line 154 a non-check as subject is memo'd)

JonathanHallam: ah nice one, dangers of copying and pasting things!

kevindew (approved):
Awesome looks good 👍

PR #1291, opened 2021-02-22
Move prepare actions

leenagupte:
Follows on from: #1275

# What's changed?
Anything to do with the `prepare` action (i.e updating drafts, publishing) is being separated into it’s own controller/route, a bit like the livestream page, so we can keep it separate from the bits that can actually be edited in collections-publisher.

# Why?
When the functionality was first added to collections-publisher to allow users to publish changes to the coronavirus pages, it made sense that the action to get the changes ready for publish was called "prepare".

It made sense because all of the content was updated in github and pulled across to collections publisher to be published, but wasn’t actually being stored.

Now, though with content being stored and updated in collections-publisher directly, the name "prepare" has become confusing.

Also, the page is rendered from the `prepare` action in the PagesController along with the actions to render the index page and the summary pages. So it's very difficult to tell which methods are need to edit content in collections-publisher and which to bring content to across from Github.

# Expected changes

The path to the page is changing 
from: `/coronavirus/<slug>/prepare`
to: `/coronavirus/<slug>/github_changes`


app/views/coronavirus/pages/prepare.html.erb
```diff
@@ -5,78 +5,68 @@
...
-        } %>
-      <% end %>
-    </div>
-    <div class="govuk-!-padding-bottom-2">
-      <%= render "govuk_publishing_components/components/heading", {
-        text: t("coronavirus.pages.preview.heading"),
-        padding: true
+      <%= render "govuk_publishing_components/components/button", {
+          text: t("coronavirus.pages.publish.button_text"),
+          margin_bottom: true,
```
kevindew: This looks a good iteration - does it change visually at all?

config/routes.rb
```diff
@@ -17,6 +17,10 @@ def matches?(request)
   root to: redirect("/step-by-step-pages", status: 302)
 
   namespace :coronavirus do
+    resources :github_changes, only: %i[show update], param: :slug do
+      post "publish", to: "github_changes#publish", on: :member
+    end
```
kevindew, line 22: I feel like a resource might be the wrong way to define this compared to the approach we've taken with sub sections and timeline entries.

We'll end up with a URL here of `/coronavirus/github_changes/:slug` which feels quite different to `coronavirus/pages/:slug/sub_sections`.

I'd instead consider defining routes within the pages resource to be consistent with the other bits. I'd have thought you'd not really want a show route, that's perhaps more of an index (since there isn't necessarily a github_content resource to show) and have update and publish not need a parameter.

spec/controllers/coronavirus/github_changes_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
+
+  describe "PUT /coronavirus/github_changes/:slug" do
+    it "updates page content in publishing-api" do
+      put :update, params: { slug: coronavirus_page.slug }
+
+      assert_publishing_api_put_content(coronavirus_page.content_id)
```
kevindew, line 33: I assume these are copied over from before rather than your additions so I'm sorry for making a comment. I just feel this and the publish calls could probably benefit from an assertion on status code too (in this same test)

app/controllers/coronavirus/pages_controller.rb
```diff
@@ -74,7 +53,7 @@ def redirect_to_index_if_slug_unknown
     end
 
     def publish_page
-      Services.publishing_api.publish(page.content_id, update_type)
+      Services.publishing_api.publish(page.content_id, "minor")
       change_state("published")
```
kevindew, line 57: Should this be inlined as per the GithubContentController? 

Also, a problem for another day, but I wonder if maybe we should think about moving this code into a service. it seems one of those annoying things where it is a small amount of code to keep in sync which is a bit dubious, but not really enough to justify doing anything with.

Also while you're clearing this file up. I think nothing makes use of: https://github.com/alphagov/collections-publisher/blob/779caa2a05c7fd3e632fda35bf10789af58afaa5/app/controllers/coronavirus/pages_controller.rb#L63-L72 anymore.

config/locales/en.yml
```diff
@@ -1,6 +1,31 @@
...
+          five:
+            step: 5. Publish changes
+            radio_heading: Update type
+            radio_options:
+              major: Major
+              minor: Minor
+            button_text: Publish
+          six:
+            step: 6. Check live
+            button_text: View live version
```
kevindew, line 28: This looks good 👍

kevindew (commented):
This is looking great - I think the only thing I've commented on that is a bit of a pain is route definitition.

app/views/coronavirus/pages/prepare.html.erb
leenagupte: I think it adds an extra line between sections. Barely noticeable in my humble opinion.

config/routes.rb
leenagupte: Oh God I ummed and ahhed about this for ages. 

It's a weird one, because "github_changes" isn't part of the "page" model, which is what I assume the "pages" resource refers to, so it looked odd when I tried to add the resource under "pages" like announcements and timeline entries are etc.

I tried putting the routes under "pages" but that made the "update" and "publish" routes weird because they expect an id.

That's when I tried the current position of making it separate, but not quite like timeline entries.

Is your suggestion to define the individual routes inside "pages", rather than a resource? e.g.
```
resources :pages,
.
.
.
  get "/github_changes/:slug/"
  put "/github_changes/:slug/"
  post "/github_changes/:slug/"
```
etc?

However the route is defined, we will still need the slug.

spec/controllers/coronavirus/github_changes_controller_spec.rb
leenagupte: I had those expectations, and took them out. I'll put them back in tomorrow.

app/controllers/coronavirus/pages_controller.rb
leenagupte: I'm going to say all of this is out of scope for this change. That's why I specifically said I wasn't going to clean up the PagesController any further as part of this change as it's a very slippery slope down the rabbit hole.

kevindew: I can understand your concern. I'd have thought only the service thing is a pain to do here (which I wasn't expecting as part of this PR - perhaps when I said another day I should have been more explicit as some undefined time in the future). I'd have thought it'd be very quick to remove change_state and the `with_longer_timeout` method.

config/routes.rb
kevindew: > It's a weird one, because "github_changes" isn't part of the "page" model, which is what I assume the "pages" resource refers to, so it looked odd when I tried to add the resource under "pages" like announcements and timeline entries are etc.

I get what you mean, I think announcements etc are all similarly part and not part of a page model too so it makes sense to handle them all consistently. With github_content we've got something that's not really very RESTful which always makes things harder.

By the way you get a slug by default by being within the `pages` resource.

I didn't have a particular suggestion for implementation but inline seems good. I guess that would come out with something like: 

```
    resources :pages,
              path: "",
              only: %i[index show update],
              param: :slug do
      get "github_content", to: "github_content#index", on: :member
      post "github_content/update", to: "github_content#update", as: "update_github_content", on: :member
      post "github_content/publish", to: "github_content#publish", as: "publish_github_content",  on: :member
```

producing:

```
➜  collections-publisher git:(timeline-summary) ✗ rails routes | grep github_content
              github_content_coronavirus_page GET    /coronavirus/:slug/github_content(.:format)                                          coronavirus/github_content#index
       update_github_content_coronavirus_page POST   /coronavirus/:slug/github_content/update(.:format)                                   coronavirus/github_content#update
      publish_github_content_coronavirus_page POST   /coronavirus/:slug/github_content/publish(.:format)                                  coronavirus/github_content#publish
```

And comparing that to a resources approach:

```
    resources :pages,
              path: "",
              only: %i[index show update],
              param: :slug do

      resources :github_content, only: %i[index] do
        post "update", as: :update, on: :collection
        post "publish", on: :collection
      end
```

You seem to get slightly odder route names out that have an `_index` suffix (this is [because `github_content` isn't a plural](https://github.com/rails/rails/blob/4e53e5ab0d15f3540d60bbbf4bcf70be575c5567/actionpack/lib/action_dispatch/routing/mapper.rb#L1218-L1220))

```
➜  collections-publisher git:(timeline-summary) ✗ rails routes | grep github_content
 update_coronavirus_page_github_content_index POST   /coronavirus/:page_slug/github_content/update(.:format)                              coronavirus/github_content#update
publish_coronavirus_page_github_content_index POST   /coronavirus/:page_slug/github_content/publish(.:format)                             coronavirus/github_content#publish
        coronavirus_page_github_content_index GET    /coronavirus/:page_slug/github_content(.:format)                                     coronavirus/github_content#index
```

The latter approach is probably more consistent with the rest of this app but is a bit more magical - and would require a bit of a fight to lose the `_index` suffix.

app/controllers/coronavirus/pages_controller.rb
leenagupte: There are quite a few little things that should be fixed in `PagesController`, that's why I tried to be disciplined in the changes in this PR and limit them only to things that are directly related to the Github changes page.

I plan to fix the niggles with `PagesController` in a separate PR.

config/routes.rb
leenagupte: Um, I called it `github_change` which is a plural.

kevindew: Oh great - that'd resolve the weird `_index` suffix then.

leenagupte: Also, I think the convention is to use `show` if an id is being passed in: https://edgeguides.rubyonrails.org/routing.html#crud-verbs-and-actions

leenagupte: I'm happy to give this a go though and see what happens:

```
resources :pages,
              path: "",
              only: %i[index show update],
              param: :slug do
      get "github_changes", to: "github_changes#show", on: :member
      post "github_changes/update", to: "github_changes#update", on: :member
      post "github_changes/publish", to: "github_changes#publish", on: :member
```

kevindew: The id isn't for the github_changes resource though, it's for the coronavirus page. Same way you can have /coronavirus/:slug/sub-sections which is an index.

kevindew: Great stuff 👍  (sorry my previous reply was out of sync your other one)

leenagupte: I'm sure I tried using `index` in my first attempt and it wouldn't pass the slug through 🤔

leenagupte: I've been having a little play about with the routes:

If they stay as they are, i.e.

```
namespace :coronavirus do
    resources :github_changes, only: %i[show update], param: :slug do
      post "publish", to: "github_changes#publish", on: :member
    end
```

We get:

```
publish_coronavirus_github_change  POST   /coronavirus/github_changes/:slug/publish(.:format)    coronavirus/github_changes#publish
coronavirus_github_change          GET    /coronavirus/github_changes/:slug(.:format)            coronavirus/github_changes#show
                                   PATCH  /coronavirus/github_changes/:slug(.:format)            coronavirus/github_changes#update
                                   PUT    /coronavirus/github_changes/:slug(.:format)            coronavirus/github_changes#update
```
Which gives us an extra unnecessary PATCH route.

Moving the individual routes inside the `pages` resource looks a bit messy in the routes file, but seems to give nicer routes:

```
namespace :coronavirus do    
    resources :pages,
              path: "",
              only: %i[index show],
              param: :slug do

      get "github_changes", to: "github_changes#show", on: :member
      put "github_changes/update", to: "github_changes#update", on: :member
      post "github_changes/publish", to: "github_changes#publish", on: :member


Gives:

github_changes_coronavirus_page GET           /coronavirus/:slug/github_changes(.:format)            coronavirus/github_changes#show
github_changes_update_coronavirus_page PUT    /coronavirus/:slug/github_changes/update(.:format)     coronavirus/github_changes#update
github_changes_publish_coronavirus_page POST  /coronavirus/:slug/github_changes/publish(.:format)    coronavirus/github_changes#publish

````

And `index` and `show` seem to be interchangeable when it comes to the path helper methods

namespace :coronavirus do    
    resources :pages,
              path: "",
              only: %i[index show],
              param: :slug do

      get "github_changes", to: "github_changes#index", on: :member
      put "github_changes/update", to: "github_changes#update", on: :member
      post "github_changes/publish", to: "github_changes#publish", on: :member


Gives:

github_changes_coronavirus_page         GET    /coronavirus/:slug/github_changes(.:format)           coronavirus/github_changes#index
github_changes_update_coronavirus_page  PUT    /coronavirus/:slug/github_changes/update(.:format)    coronavirus/github_changes#update
github_changes_publish_coronavirus_page POST   /coronavirus/:slug/github_changes/publish(.:format)   coronavirus/github_changes#publish

```

Adding the routes as a resource under `pages` gives some strange routes which require an id as well as the extra PATCH route:

```

namespace :coronavirus do    
    resources :pages,
              path: "",
              only: %i[index show],
              param: :slug do

      resources :github_changes, only: %i[index update] do
        post "publish", to: "github_changes#publish", on: :member
      end


Gives:

publish_coronavirus_page_github_change POST   /coronavirus/:page_slug/github_changes/:id/publish(.:format)   coronavirus/github_changes#publish
coronavirus_page_github_changes        GET    /coronavirus/:page_slug/github_changes(.:format)               coronavirus/github_changes#index coronavirus_page_github_change         PATCH  /coronavirus/:page_slug/github_changes/:id(.:format)           coronavirus/github_changes#update
                                       PUT    /coronavirus/:page_slug/github_changes/:id(.:format)                                 coronavirus/github_changes#update
```

If we add a param to the resource, which seems to defeat the purpose of nesting the resource in the first place, we no longer need to pass in an ID, but we still get the extra PATCH route:

```
namespace :coronavirus do    
    resources :pages,
              path: "",
              only: %i[index show],
              param: :slug do

      resources :github_changes, only: %i[index update], param: :slug do
        post "publish", to: "github_changes#publish", on: :member
      end


Gives:

publish_coronavirus_page_github_change POST   /coronavirus/:page_slug/github_changes/:slug/publish(.:format)   coronavirus/github_changes#publish
coronavirus_page_github_changes        GET    /coronavirus/:page_slug/github_changes(.:format)                 coronavirus/github_changes#index
coronavirus_page_github_change         PATCH  /coronavirus/:page_slug/github_changes/:slug(.:format)           coronavirus/github_changes#update
                                       PUT    /coronavirus/:page_slug/github_changes/:slug(.:format)           coronavirus/github_changes#update
```

In conclusion, my preference is add "individual" routes inside the `pages` resource.

leenagupte:
@kevindew I've made the changes as discussed and fixed up. Can you re-review?

spec/controllers/coronavirus/github_changes_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
+
+  describe "POST /coronavirus/:slug/github_changes/update" do
+    it "updates page content in publishing-api" do
+      put :update, params: { slug: coronavirus_page.slug }
+
+      assert_publishing_api_put_content(coronavirus_page.content_id)
+      expect(response).to have_http_status(:redirect)
```
kevindew, line 34:
```suggestion
      expect(response).to redirect_to(github_changes_coronavirus_page_path(coronavirus_page.slug))
```

I feel like just saying you expect a redirect seems a bit unusual for a response assertion and it's more common to assert the redirect destination.

spec/controllers/coronavirus/github_changes_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  describe "POST /coronavirus/:slug/github_changes/publish" do
+    it "publishes page content" do
+      post :publish, params: { slug: coronavirus_page.slug }
+
+      assert_publishing_api_publish(coronavirus_page.content_id)
+      expect(response).to have_http_status(:redirect)
```
kevindew, line 43: Same point as above regarding redirect.

kevindew (commented):
Looks great Leena - just one thing I've flagged to address.

kevindew (approved):
Awesome 👍

PR #1292, opened 2021-02-22
Fix subsection bug for the create method.

JonathanHallam:
This fixes a bug whereby it was possible to save a new subsection with invalid content. It's also brought to light a lack of error handling that we have not only for the create and update methods.
 
We've been ignoring ActiveRecord validation errors and relying on the validation that we have in the presenters to show users their errors. We're checking the validity of sub_sections before we make any attempt to save them. This means that we're not at any risk of saving invalid subsections and that we are using errors as created by the model layer.

Trello - https://trello.com/c/p6Rq4Gj8/163-stop-saving-subsections-with-invalid-markdown-the-band-aid-method

### DIFF
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -11,7 +11,7 @@ module Coronavirus
       @sub_section = page.sub_sections.new(sub_section_params)
 
       unless @sub_section.valid?
-        render :new
+        render :new, status: :unprocessable_entity
         return
       end
 
@@ -22,7 +22,7 @@ module Coronavirus
 
       if draft_updater.errors.any?
         flash.now["alert"] = draft_updater.errors.to_sentence
-        render :new
+        render :new, status: :internal_server_error
       else
         redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully created." }
       end
@@ -37,7 +37,7 @@ module Coronavirus
       @sub_section.assign_attributes(sub_section_params)
 
       unless @sub_section.valid?
-        render :edit
+        render :edit, status: :unprocessable_entity
         return
       end
 
@@ -48,7 +48,7 @@ module Coronavirus
 
       if draft_updater.errors.any?
         flash.now["alert"] = draft_updater.errors.to_sentence
-        render :edit
+        render :edit, status: :internal_server_error
       else
         redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully updated." }
       end
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -34,14 +34,20 @@ module Coronavirus
 
     def update
       @sub_section = page.sub_sections.find(params[:id])
+      @sub_section.assign_attributes(sub_section_params)
+
+      unless @sub_section.valid?
+        render :edit
+        return
+      end
 
       SubSection.transaction do
-        @sub_section.update!(sub_section_params)
+        @sub_section.save!
         raise ActiveRecord::Rollback unless draft_updater.send
       end
 
       if draft_updater.errors.any?
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+        flash.now["alert"] = draft_updater.errors.to_sentence
         render :edit
       else
         redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully updated." }
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -9,11 +9,22 @@ module Coronavirus
 
     def create
       @sub_section = page.sub_sections.new(sub_section_params)
-      if @sub_section.save && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: "Sub-section was successfully created."
-      else
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+
+      unless @sub_section.valid?
         render :new
+        return
+      end
+
+      SubSection.transaction do
+        @sub_section.save!
+        raise ActiveRecord::Rollback unless draft_updater.send
+      end
+
+      if draft_updater.errors.any?
+        flash.now["alert"] = draft_updater.errors.to_sentence
+        render :new
+      else
+        redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully created." }
       end
     end
 
--- a/spec/controllers/coronavirus/sub_sections_controller_spec.rb
+++ b/spec/controllers/coronavirus/sub_sections_controller_spec.rb
@@ -157,6 +157,13 @@ RSpec.describe Coronavirus::SubSectionsController do
         expect(subject.body).to include("Unable to parse markdown")
       end
     end
+
+    context "given a featured link that is not in the content" do
+      it "displays the expected error" do
+        sub_section_params.merge!(featured_link: "/banana")
+        expect(subject.body).to include("Featured link does not exist in accordion content")
+      end
+    end
   end
 
   describe "DELETE /coronavirus/:page_slug/sub_sections/:id" do
--- a/spec/controllers/coronavirus/sub_sections_controller_spec.rb
+++ b/spec/controllers/coronavirus/sub_sections_controller_spec.rb
@@ -70,6 +70,10 @@ RSpec.describe Coronavirus::SubSectionsController do
     context "subsection content error" do
       let(:content) { "bad_content" }
 
+      it "doesn't save to the database" do
+        expect { subject }.not_to(change { Coronavirus::SubSection.count })
+      end
+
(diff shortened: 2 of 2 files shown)


app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -9,11 +9,22 @@ def new
...
+        if @sub_section.save
+          unless draft_updater.send
+            errors = draft_updater.errors.to_sentence
+            raise ActiveRecord::Rollback
+          end
+        end
+      end
+
+      if !@sub_section.created_at || errors
+        @sub_section.errors.add :base, errors
```
leenagupte, line 24: Can the error be rendered using an `flash["alert"]` instead? 

e.g.
```
if draft_updater.errors.any?
  flash["alert"] = draft_updater.errors.to_sentence
  render :new
end
```

app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -9,11 +9,22 @@ def new
...
-      else
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+      errors = nil
+
+      SubSection.transaction do
+        if @sub_section.save
+          unless draft_updater.send
+            errors = draft_updater.errors.to_sentence
+            raise ActiveRecord::Rollback
+          end
```
leenagupte, line 19: I don't think the conditional is needed. I think this can be written in similar style to `update`

```
@sub_section.save!
raise ActiveRecord::Rollback unless draft_updater.send
```

app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -9,11 +9,18 @@ def new
...
-      else
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+
+      SubSection.transaction do
+        if @sub_section.save
+          raise ActiveRecord::Rollback unless draft_updater.send
+        end
+      end
+
+      if !@sub_section.created_at || draft_updater.errors.any?
```
kevindew, line 19: I'm a bit confused by this dual conditional here - particularly as we only test one. 

I'm assuming the two parts of it are needed because the item can be invalid or the draft update can fail, which actually makes me think we've missed something on the update method - as that will just raise an exception.

It looks like we've also now lost the ability to display model errors with `flash["alert"] = draft_updater.errors.to_sentence`.

Perhaps both the methods need a little bit of a quick pseudo code think through and to decide what to test as both seem to have hidden surprises.

app/controllers/coronavirus/sub_sections_controller.rb
JonathanHallam: Sorted after a talk in person 🎉

app/controllers/coronavirus/sub_sections_controller.rb
JonathanHallam: Conditional is not needed. Sorted after a talk offline 🥇

app/controllers/coronavirus/sub_sections_controller.rb
JonathanHallam: This work has made us realise that we weren't dealing with errors properly as they came from the model. We've updated to avoid the need for the conditional and to ensure that errors are caught by our model layer.

kevindew: Great stuff 👍 

It'd be good to have tests for both scenarios - since they're quite confusing and easy to overlook.

I also feel like it'd be good if the code explicitly catered for them so it was easier to follow. I wondered if something like this would work: 

```
@sub_section = page.sub_sections.new(sub_section_params)

unless @sub_section.valid?
  render :new
  return
end

SubSection.transaction do
  @sub_section.save!
  raise ActiveRecord::Rollback unless draft_updater.send
end

if draft_updater.errors
  redirect_to coronavirus_page_path(page.slug), { notice: "Sub-section was successfully created." }
else
  flash["alert"] = draft_updater.errors.to_sentence
  render :new
end
```

leenagupte: Ooh this is much nicer! ❤️

JonathanHallam: I'd not seen the `valid?` method before, this is really clean, thank you :)

kevindew: > Ooh this is much nicer! ❤️ I forget now, but I'm assuming if the call to `valid?` fails, it will display the model errors?

Oh thanks - yeah, once you've called `valid?` the model has it's `errors` property populated and will fall into this condition: https://github.com/alphagov/collections-publisher/blob/7cf7c0f4a870c3b57df531767db7e4d8d7250c6b/app/views/coronavirus/sub_sections/edit.html.erb#L25-L27

What I'm less clear on is whether setting `flash["alert"]` works as I'm not sure where that's handled (I copied it from your earlier suggestion)

app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -9,11 +9,22 @@ def new
...
       @sub_section = page.sub_sections.new(sub_section_params)
-      if @sub_section.save && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: "Sub-section was successfully created."
-      else
-        @sub_section.errors.add :base, draft_updater.errors.to_sentence
+
+      unless @sub_section.valid?
+        render :new
+        return
+      end
```
kevindew, line 16: I'm not sure if we're testing this conditional in the tests?

From what I can tell there's a bunch of code (too much in my opinion) that test the draft_updater.send fails with: https://github.com/alphagov/collections-publisher/blob/8a4d67ddcbe7d6a053ddab189099456f7fa5ba4c/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L56-L84 but nothing for the situation where the model is invalid.

Similar situation with the update method. where it seems that the validation code is perhaps indirectly tested with https://github.com/alphagov/collections-publisher/blob/8a4d67ddcbe7d6a053ddab189099456f7fa5ba4c/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L161-L167 but I'm not sure and seems a bit cryptic.

kevindew (commented):
Great code looks good in the controllers, I'm a bit concerned it's not sufficiently tested so have left a comment to that effect.

Were you able to check that using `flash["alert"] = draft_updater.errors.to_sentence` does work ok?

app/controllers/coronavirus/sub_sections_controller.rb
JonathanHallam: So our thinking for this was that it would be covered by the featured link tests. That block is tested for create as well as update here https://github.com/alphagov/collections-publisher/blob/master/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L98-L105. I do agree that this is slightly opaque but I'm happy enough to leave it that way until the work to tidy up the validation is done as I think it's important to be explicit about it once we're more knowingly relying on validation in the model.

The testing for the `flash["alert"] stuff is done here https://github.com/alphagov/collections-publisher/blob/master/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L77-L79 because that's how that message is displayed unless I'm misunderstanding something.

kevindew: > So our thinking for this was that it would be covered by the featured link tests. That block is tested for create as well as update here https://github.com/alphagov/collections-publisher/blob/master/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L98-L105. I do agree that this is slightly opaque but I'm happy enough to leave it that way until the work to tidy up the validation is done as I think it's important to be explicit about it once we're more knowingly relying on validation in the model.

Oh gosh, I didn't see that one. That does seem a bit cryptic.

It didn't seem like we'd have a particular motivation to fix this as part of the validation work so I think it's probably better sorted here. I quickly put together a commit that makes the 3 cases for each clear, if you're happy with it feel free to cherry pick it onto this PR: https://github.com/alphagov/collections-publisher/commit/6701bb0bd02443e55c8b3c8fc79c43c8eb06054d

> The testing for the `flash["alert"] stuff is done here https://github.com/alphagov/collections-publisher/blob/master/spec/controllers/coronavirus/sub_sections_controller_spec.rb#L77-L79 because that's how that message is displayed unless I'm misunderstanding something.

I was wondering more how that played out visually since it's a different way to set the error to how we did it before. It looks like it displays almost the same but skips the "There is a problem" line and is placed differently.

Screenshot 2021-02-24 at 19 10 48

It's worth checking with others, but this doesn't seem too big a concern for me.

One thing I have noticed though is that the `flash["alert"]` hangs around until the next page because we don't redirect. We'll need to change them to `flash.now["alert"]`.

For example: (note that it should be impossible to have validation errors and draft updater errors at the same time)

Screenshot 2021-02-24 at 19 13 04

JonathanHallam: awesome, thanks for this, that break up of tests is lovely, we've put your commit in and adjusted the `flash` stuff. As for the difference in placement, I'm not too worried about this as it will all be brought into line when the model validation stuff happens.

kevindew (approved):
Nice one

PR #1294, opened 2021-02-24
Tidy up `PagesController`

leenagupte:
Follows on from: #1291

# What's changed and why?

In preparation for moving the content from the `PagesController` to a locale file, the controller has been tidied up a little. This means removing some unused methods, and retrieving models from the database rather than building them in memory.

There's also a rather controversial final commit that removes the redirect to the index page and instead throws a 404. The bonus to this is that we won't have to try to figure out where to put the content for "slug unknown" error message later.

### DIFF
--- a/app/controllers/coronavirus/pages_controller.rb
+++ b/app/controllers/coronavirus/pages_controller.rb
@@ -31,25 +31,12 @@ module Coronavirus
 
   private
 
-    def initialise_pages
-      page_configs.keys.map do |page|
-        Pages::ModelBuilder.new(page.to_s).page
-      end
-    end
-
     def page
       @page ||= Pages::ModelBuilder.call(slug)
     end
 
     def draft_updater
       @draft_updater ||= Pages::DraftUpdater.new(page)
-    end
-
-    def redirect_to_index_if_slug_unknown
-      if slug_unknown?
-        flash[:alert] = "'#{slug}' is not a valid page.  Please select from one of those below."
-        redirect_to coronavirus_pages_path
-      end
     end
 
     def publish_page
@@ -64,8 +51,21 @@ module Coronavirus
       params[:slug]
     end
 
+    def redirect_to_index_if_slug_unknown
+      if slug_unknown?
+        flash[:alert] = "'#{slug}' is not a valid page.  Please select from one of those below."
+        redirect_to coronavirus_pages_path
+      end
+    end
+
     def slug_unknown?
       !page_configs.key?(slug.to_sym)
+    end
+
+    def initialise_pages
+      page_configs.keys.map do |page_config|
+        Pages::ModelBuilder.new(page_config.to_s).page
+      end
     end
 
     def page_configs
--- a/app/controllers/coronavirus/pages_controller.rb
+++ b/app/controllers/coronavirus/pages_controller.rb
@@ -54,21 +54,10 @@ module Coronavirus
 
     def publish_page
       Services.publishing_api.publish(page.content_id, "minor")
-      change_state("published")
+      page.update!(state: "published")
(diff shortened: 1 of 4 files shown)


app/controllers/coronavirus/pages_controller.rb
```diff
@@ -54,23 +54,12 @@ def redirect_to_index_if_slug_unknown
...
-      prior_timeout = Services.publishing_api.client.options[:timeout]
-      Services.publishing_api.client.options[:timeout] = 10
-
-      begin
-        yield
-      ensure
-        Services.publishing_api.client.options[:timeout] = prior_timeout
-      end
-    end
-
```
kevindew, line 73: Yay ⭐

app/controllers/coronavirus/pages_controller.rb
```diff
@@ -26,13 +26,13 @@ def discard
...
       end
-      redirect_to coronavirus_page_path(slug), message
+      redirect_to coronavirus_page_path(page.slug), message
     end
 
   private
 
     def page
-      @page ||= Pages::ModelBuilder.call(slug)
+      @page ||= Page.find_by(slug: slug)
```
kevindew: I think the model builder served a purpose that would now be lost of how to initially create a page if one does not exist in the database - effectively acting as a lazy load mechanism. Is there an alternative process as to how that is supposed to happen?

By the way, this should be `find_by!` so that this will return a 404 if this fails.

app/controllers/coronavirus/pages_controller.rb
```diff
@@ -1,7 +1,6 @@
 module Coronavirus
   class PagesController < ApplicationController
     before_action :require_coronavirus_editor_permissions!
-    before_action :redirect_to_index_if_slug_unknown, only: %w[show]
```
kevindew, line 4: This seems good 👍  I don't find it _too_ scandalous 😆  Seems weird we'd give these pages a special redirect handling.

app/controllers/coronavirus/pages_controller.rb
```diff
@@ -11,7 +10,7 @@ def index
     end
 
     def show
-      page
+      return not_found if page.blank?
```
kevindew, line 13: As mentioned on an earlier comment this can be avoided with the bang variant of `find_by` (well if I remember correctly)

kevindew (commented):
This is a nice cleanup @leenagupte 🥇 

My only area of concern is that the page model builder provides some useful functionality for initialisation of pages that would be lost here.

leenagupte:
> This is a nice cleanup @leenagupte 🥇
> 
> My only area of concern is that the page model builder provides some useful functionality for initialisation of pages that would be lost here.

The initialisation still happens when the `index` page is loaded. That still uses `ModelBuilder` and hasn't been removed. It's called by the `initialise_pages` before action.

app/controllers/coronavirus/pages_controller.rb
leenagupte: The initialisation still happens when the index page is loaded. That still uses `ModelBuilder` and hasn't been removed. It's called by the `initialise_pages` before action.

app/controllers/coronavirus/pages_controller.rb
leenagupte: The initialisation still happens when the index page is loaded. That still uses `ModelBuilder` and hasn't been removed. It's called by the `initialise_pages` before action.

kevindew: Oh cool - sorry I didn't realise that.

kevindew (approved):
Looks good 👍

PR #1296, opened 2021-02-25
Move announcements content to locale file

leenagupte:
Follows on from: #1275

# What's changed?

Moves controller and view content for announcements to a locale file

### DIFF
--- a/app/views/coronavirus/announcements/_form.html.erb
+++ b/app/views/coronavirus/announcements/_form.html.erb
@@ -1,6 +1,6 @@
 <%= render "govuk_publishing_components/components/input", {
   label: {
-    text: "Enter the title of the announcement",
+    text: t("coronavirus.announcements.form.title.label"),
     bold: true,
   },
   name: "announcement[title]",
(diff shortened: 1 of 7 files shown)


app/views/coronavirus/announcements/_form.html.erb
```diff
@@ -11,25 +11,25 @@
...
   value: @announcement.path,
   id: "path",
   error_message: error_items(@announcement.errors.messages, :path),
-  hint: "Example format: /government/news/coronavirus-covid-19",
+  hint: t("coronavirus.announcements.form.path.hint"),
 } %>
 
 <% legend = capture do %>
-  <span class="govuk-heading-s govuk-!-margin-bottom-0"><%= "Enter the date of publication" %></span>
+  <span class="govuk-heading-s govuk-!-margin-bottom-0"><%= t("coronavirus.announcements.form.date.label") %></span>
```
kevindew, line 25: Might be better to refer to this as `date.legend` since it isn't technically used for a label.

app/views/coronavirus/announcements/edit.html.erb
```diff
@@ -1,15 +1,15 @@
 <%
 links = [
   {
-    text: 'Coronavirus pages',
+    text: t("coronavirus.pages.show.link_text"),
```
kevindew, line 4: This shouldn't refer to a key for a different view. It'd be more conventional to repeat the definition (as painful as that may seem).

kevindew (commented):
Looks good, just a few minor comments.

app/views/coronavirus/announcements/edit.html.erb
leenagupte: I'm not doing that. I'll leave it as hardcoded text.

leenagupte: It might be conventional, but that's just pointless. It's a link back to the summary page, it should always be the same.

leenagupte: A better solution is to move all of that links stuff to a shared partial. Something for another day.

kevindew: Yup, a partial sounds like the right approach to avoid the repetition.

I don't mind it as hardcoded text, that seems to have a much lower contagion risk than sharing the view copy.

kevindew (approved):

PR #1299, opened 2021-02-25
Tidy up featured link testing

GDSNewt:
As inclusion of featured links is handled at a model level this adds a
test to ensure that the model is validating featured links correctly.
It also removes old functionality that gave the json presenter
responsibility for featured link errors.

### DIFF
--- a/app/presenters/coronavirus/sub_section_json_presenter.rb
+++ b/app/presenters/coronavirus/sub_section_json_presenter.rb
@@ -29,12 +29,6 @@ class Coronavirus::SubSectionJsonPresenter
         title: title,
         sub_sections: sub_sections,
       }
-
-    if @sub_section.featured_link.present? && !link_set_as_featured?
-      errors << "Featured link does not exist in accordion content"
-    end
-
-    @output
   end
 
   def errors
@@ -48,14 +42,6 @@ class Coronavirus::SubSectionJsonPresenter
 
   def sub_sections
     content_groups.map { |content_group| sub_section_hash_from_content_group(content_group) }
-  end
-
-  def link_set_as_featured?
-    featured_links = @output[:sub_sections].flat_map do |section|
-      section[:list].map { |item| item[:featured_link] }
-    end
-
-    featured_links.any?
   end
 
   # Groups the sub section content into an array of arrays, such that:
--- a/spec/models/coronavirus/sub_section_spec.rb
+++ b/spec/models/coronavirus/sub_section_spec.rb
@@ -33,5 +33,19 @@ RSpec.describe Coronavirus::SubSection do
       expect(sub_section).not_to be_valid
       expect(sub_section.errors).to have_key(:content)
     end
+
+    it "validates that the featured link is in content" do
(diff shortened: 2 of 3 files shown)


app/presenters/coronavirus/sub_section_json_presenter.rb
```diff
@@ -50,14 +44,6 @@ def sub_sections
     content_groups.map { |content_group| sub_section_hash_from_content_group(content_group) }
   end
 
-  def link_set_as_featured?
```
leenagupte, line 53: I think you still need to do this so that the link has the `featured` flag set to true in the publishing-api payload, e.g.

```
{
  url: "/getting-tested-for-coronavirus?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
  label: "Find out about coronavirus testing",
  description: "Find out whether you need a PCR test or rapid lateral flow test for coronavirus (COVID-19)",
  featured_link: true
}
```

JonathanHallam: as far as we can tell the only place that that method was being used was here https://github.com/alphagov/collections-publisher/pull/1299/files#diff-221c8bf8dbf2b355bc4c0e9cb9ee9e8424bd94bbd3078a19efea804a65d9bba6L33 in order to check whether we needed to add extra errors

spec/models/coronavirus/sub_section_spec.rb
```diff
@@ -33,5 +33,12 @@
       expect(sub_section).not_to be_valid
       expect(sub_section.errors).to have_key(:content)
     end
+
+    it "fails if featured link is not in content" do
+      sub_section.featured_link = "/bananas"
+
+      expect(sub_section).not_to be_valid
+      expect(sub_section.errors).to have_key(:featured_link)
+    end
```
kevindew, line 42: I think we should also have a test for when the featured link is valid, as I don't think that's actually tested at the model level. 

The other two validations in this file seem to be implicitly tested by: https://github.com/alphagov/collections-publisher/blob/6554f8303cc642fde31ca41619580a4dbfee1837/spec/models/coronavirus/sub_section_spec.rb#L17-L21

kevindew (commented):
Nice one 👍  - it's great to see us chipping away at the presenter validation layer.

spec/models/coronavirus/sub_section_spec.rb
JonathanHallam: Good shout, I've added this in 🎉

kevindew (approved):
Nice one

app/presenters/coronavirus/sub_section_json_presenter.rb
JonathanHallam: As far as we could tell the only place that this was in the validation in the json presenter which has now been removed. https://github.com/alphagov/collections-publisher/pull/1299/files#diff-221c8bf8dbf2b355bc4c0e9cb9ee9e8424bd94bbd3078a19efea804a65d9bba6L33

It's not being used anywhere else in the repo so i think it's safe

PR #1301, opened 2021-02-26
Move reorder announcements content to locale file

leenagupte:
Follows on from: #1275

# What's changed?

Moves controller and view content for reordering announcements to a locale file

### DIFF
--- a/app/controllers/coronavirus/reorder_announcements_controller.rb
+++ b/app/controllers/coronavirus/reorder_announcements_controller.rb
@@ -26,10 +26,10 @@ module Coronavirus
       end
 
       if success
-        message = "Announcements were successfully reordered."
+        message = helpers.t("coronavirus.reorder_announcements.update.success")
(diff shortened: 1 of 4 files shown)


app/controllers/coronavirus/reorder_announcements_controller.rb
```diff
@@ -26,10 +26,10 @@ def update
       end
 
       if success
-        message = "Announcements were successfully reordered."
+        message = helpers.t("coronavirus.reorder_announcements.update.success")
```
DilwoarH, line 29: Just curious (non blocker) but why does this need `helpers.t` and the other ones just need `t`?

leenagupte: It's because it's in the controller. `t` is a helper method that's only available in views. The other way to write this is `I18n.t`, but using `helpers` allows the controller to access the same helper method that the view uses.

DilwoarH (approved):

PR #1307, opened 2021-03-01
Make DraftUpdater errors more specific

leenagupte:

# What's changed and why?

The first step to wrapping Coronavirus controller actions in transactions is to return "useful" errors from DraftUpdater.
I've made an attempt by splitting the existing `DraftUpdaterError` into two more specific errors.

I'd appreciate feedback on whether this is the right starting point for the transactions work.

### DIFF
--- a/app/services/coronavirus/pages/draft_updater.rb
+++ b/app/services/coronavirus/pages/draft_updater.rb
@@ -1,6 +1,7 @@
 module Coronavirus::Pages
   class DraftUpdater
-    DraftUpdaterError = Class.new(StandardError)
+    class UpdateFailedError < StandardError; end
+    class PayloadInvalidError < StandardError; end
 
     attr_reader :page
 
@@ -18,7 +19,7 @@ module Coronavirus::Pages
       if content_builder.success?
         Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
       else
-        raise DraftUpdaterError, content_builder.errors.to_sentence
+        raise PayloadInvalidError, content_builder.errors.to_sentence
       end
     end
 
@@ -27,7 +28,7 @@ module Coronavirus::Pages
       page.update!(state: "draft")
     rescue GdsApi::HTTPErrorResponse => e
       error_handler(e, "Failed to update the draft content item. Try saving again.")
-    rescue DraftUpdaterError => e
+    rescue UpdateFailedError => e
       error_handler(e)
     end
 
--- a/app/controllers/coronavirus/github_changes_controller.rb
+++ b/app/controllers/coronavirus/github_changes_controller.rb
@@ -8,13 +8,11 @@ module Coronavirus
     end
 
     def update
-      message = if draft_updater.send
-                  { notice: helpers.t("coronavirus.github_changes.update.success") }
-                else
-                  { alert: draft_updater.errors.to_sentence }
-                end
+      draft_updater.send
 
-      redirect_to github_changes_coronavirus_page_path(page.slug), message
+      redirect_to github_changes_coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.github_changes.update.success")
+    rescue Coronavirus::Pages::DraftUpdater::PayloadInvalidError => e
+      redirect_to github_changes_coronavirus_page_path(page.slug), alert: e.message
     end
 
     def publish


kevindew:
Nice one, picking this up. The [update](https://github.com/alphagov/collections-publisher/pull/1307/files#diff-38a63efaf05f6e528cec8bf5c5f61a6398efe03ea663fedc61756bcfd45d21e6R10) definition looks good. Though I don't think we should use [`error_handler`](https://github.com/alphagov/collections-publisher/pull/1307/files#diff-1799ac1656087e4c121d217cf411c664d159dca89e7d6c065a8ec0c976601423R32) as that is involved in DraftUpdater having state (through a collection of errors).

One thing to consider is that DraftUpdater is probably overly complex in how helpful it tries to be for error handling with the variety it provides for users and that is likely a source of additional complexity - and probably creates some dubious scenarios by using exception messages directly in errors (for example: https://github.com/alphagov/collections-publisher/blob/ca14ae8058f4115fe60eaefba8e4102a2c8ae19e/app/services/coronavirus/pages/content_builder.rb#L25)

I imagine a good exercise would be to map out the error scenarios and how they'd be represented with exceptions (hopefully we can collate a few of them together)?

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -27,7 +28,7 @@ def send
       page.update!(state: "draft")
     rescue GdsApi::HTTPErrorResponse => e
       error_handler(e, "Failed to update the draft content item. Try saving again.")
-    rescue DraftUpdaterError => e
+    rescue UpdateFailedError => e
```
leenagupte, line 31: Should this error be re-raised so that forces any transactions in the controller to be rolled back?

leenagupte: Will this error every be thrown?

kevindew: With this change it can't be caught as it was catching the error raised at line 22 which was DraftUpdaterError, however the types don't match. It seems a rather cryptic implementation.

I'd have thought you'd not need a catch here but to just let the exception raise.

As you're heading towards multiple exception types you'll want to have a think about what you view code to look like as otherwise you may be heading towards catching a bunch of different exception types which could be arduous.

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -1,6 +1,7 @@
 module Coronavirus::Pages
   class DraftUpdater
-    DraftUpdaterError = Class.new(StandardError)
+    class UpdateFailedError < StandardError; end
+    class PayloadInvalidError < StandardError; end
```
kevindew, line 4: Minor, these would be better suited to inherit from RuntimeError.

leenagupte:
> https://github.com/alphagov/collections-publisher/blob/ca14ae8058f4115fe60eaefba8e4102a2c8ae19e/app/services/coronavirus/pages/content_builder.rb#L25

I saw this 😨. I _think_ this is trying to catch errors connecting to GitHub? I wonder if, rather than `DraftUpdater`, the errors should be raised in `ContentBuilder` in [`validate_content`](https://github.com/alphagov/collections-publisher/blob/master/app/services/coronavirus/pages/content_builder.rb#L42). 

This is based on the assumption that after all of the model validation has been moved back into the model,`ContentBuilder` should only be concerned with validating the content from GitHub.

kevindew:
> > https://github.com/alphagov/collections-publisher/blob/ca14ae8058f4115fe60eaefba8e4102a2c8ae19e/app/services/coronavirus/pages/content_builder.rb#L25
> 
> I saw this 😨. I _think_ this is trying to catch errors connecting to GitHub? I wonder if, rather than `DraftUpdater`, the errors should be raised in `ContentBuilder` in [`validate_content`](https://github.com/alphagov/collections-publisher/blob/master/app/services/coronavirus/pages/content_builder.rb#L42).
> 
> This is based on the assumption that after all of the model validation has been moved back into the model,`ContentBuilder` should only be concerned with validating the content from GitHub.

Yeah, GitHub connection sounds a good shout for it. I had assumed it might be for when descriptions aren't found but I think your reasoning sounds more likely - perhaps it is both though?

leenagupte:
@kevindew After talking in Slack, I'm going to switch focus to `ContentBuilder` and change that class to raise errors rather than adding to a list of errors. Will close this PR.

Thanks for the feedback/review!

PR #1309, opened 2021-03-02
Raise error in sub section json presenter

leenagupte:

# What's changed?
Raise an error in `SubSectionJsonPresenter` when a sub-section contains invalid markdown.

# Why?
This is part of the work to apply a consistent approach to error handling for Coronavirus pages. 

Rather than storing errors in a list, we want to surface these errors back up to the controller so that when we start saving changes to the page in transactions, we can rely on the exceptions being thrown at different points in the code to force the database changes to be rolled back.

### DIFF
--- a/app/presenters/coronavirus/sub_section_json_presenter.rb
+++ b/app/presenters/coronavirus/sub_section_json_presenter.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Coronavirus::SubSectionJsonPresenter
+  class MarkdownInvalidError < RuntimeError; end
+
   HEADER_PATTERN = PatternMaker.call(
     "starts_with hashes then perhaps_spaces then capture(title) and nothing_else",
     hashes: "#+",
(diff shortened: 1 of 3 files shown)


kevindew (approved):
This change seems a good iterative step 👍 

I'm a bit confused by this part of the first commit message: "The benefit of doing this is that we will eventually be able to surface this error in the Subsection controller" - I'd have thought we don't intend to need this exception in the controller as that would be handled by model validation? 


Also I'm a little unclear what we actually want a subsection controller to look like when this work is finished, do you have something in mind?

leenagupte:
> This change seems a good iterative step 👍
> 
> I'm a bit confused by this part of the first commit message: "The benefit of doing this is that we will eventually be able to surface this error in the Subsection controller" - I'd have thought we don't intend to need this exception in the controller as that would be handled by model validation?

The validation will be handled in the model eventually, but I was working off of the assumption that that piece of work might not be finished by the time I start updating the controllers.

> 
> Also I'm a little unclear what we actually want a subsection controller to look like when this work is finished, do you have something in mind?

The subsection controller won't change much. Calls to `save!` and `draft_updater` will still be wrapped in a transaction, but transaction will rollback if draft_updater throws an exception (e.g. couldn't connect to GitHub, there's invalid content in GitHub, or couldn't connect to publishing-api) rather than having to manually throw an `ActiveRecord::Rollback`.

PR #1311, opened 2021-03-03
Moving validation to model

JonathanHallam:
Raising this as a draft PR to get an idea of what our thinking is. 
### Things to ignore
Some changes are in the wrong commit, this was after we realised some problems and will be fixed before things get too much further forward.
Variable naming - we haven't done it for some things, again will be fixed
Tests failing - Some tests are failing at the moment, this is due to not calling things properly and is just something we haven't gotten around to yet.

### Things to consider
Is this the right approach? There's obviously a lot more functionality to add but does making a new model make the most sense or is there something we haven't considered.
Is it right to make models in models? Because we don't have any new views being made for the `ContentGroup` model, we've decided to make the `SubSection` model responsible for creating the `ContentGroups`. Is there another way to do this that might be better?

### How to review
The code is very much incomplete, probably best to look at the commit messages to best understand what it is we want to be doing here and taking the code with a pinch of salt.

### DIFF
--- a/db/migrate/20210226152201_create_content_groups.rb
+++ b/db/migrate/20210226152201_create_content_groups.rb
@@ -1,0 +1,12 @@
+class CreateContentGroups < ActiveRecord::Migration[6.0]
+  def change
+    create_table :content_groups do |t|
+      t.references :coronavirus_sub_section, foreign_key: true, null: false
+      t.string :header
+      t.string :links
+      t.integer :position
+
+      t.timestamps
+    end
+  end
+end
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -35,6 +35,7 @@ module Coronavirus
     def update
       @sub_section = page.sub_sections.find(params[:id])
       @sub_section.assign_attributes(sub_section_params)
+      @sub_section.reload_content_groups
 
       unless @sub_section.valid?
         render :edit, status: :unprocessable_entity
--- a/app/models/coronavirus/content_group.rb
+++ b/app/models/coronavirus/content_group.rb
@@ -1,0 +1,7 @@
+class Coronavirus::ContentGroup < ApplicationRecord
+  belongs_to :sub_section, foreign_key: "coronavirus_sub_section_id"
+  validates :links, presence: true
+  validates :sub_section, presence: true
+
+  serialize :links, Array
+end
--- a/app/models/coronavirus/content_group.rb
+++ b/app/models/coronavirus/content_group.rb
@@ -1,7 +1,24 @@
 class Coronavirus::ContentGroup < ApplicationRecord
+  LINK_PATTERN = PatternMaker.call(
+    "starts_with perhaps_spaces within(sq_brackets,capture(label)) then perhaps_spaces and within(brackets,capture(url))",
+    label: '\s*\w.+',
+    url: '\s*(\b(https?)://)?[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]\s*',
+  )
+
   belongs_to :sub_section, foreign_key: "coronavirus_sub_section_id"
   validates :links, presence: true
-  validates :sub_section, presence: true
+  # validates :coronavirus_sub_section_id, presence: true
+  validate :links_in_correct_format
 
   serialize :links, Array
+
+  def links_in_correct_format
+    links.each do |link|
+      errors.add(:links, "Unable to parse markdown: '#{link}'") unless is_link?(link)
+    end
+  end
+
+  def is_link?(text)
+    LINK_PATTERN =~ text
+  end
 end
--- a/app/models/coronavirus/sub_section.rb
+++ b/app/models/coronavirus/sub_section.rb
@@ -13,7 +13,7 @@ class Coronavirus::SubSection < ApplicationRecord
   validates :title, :content, presence: true
   validates :page, presence: true
   validate :featured_link_must_be_in_content
-  before_create :create_content_groups
+  after_create :create_content_groups
 
   def featured_link_must_be_in_content
     if featured_link.present? && !content.include?(featured_link)
@@ -39,7 +39,9 @@ class Coronavirus::SubSection < ApplicationRecord
       content_group[:header] = group.shift if is_header?(group[0])
       content_group[:links] = group
       content_group[:position] = index
+      # content_group[:coronavirus_sub_section_id] = id
       a = Coronavirus::ContentGroup.new(content_group)
+      a.coronavirus_sub_section_id = id
       a.save!
     end
   end
--- a/app/models/coronavirus/sub_section.rb
+++ b/app/models/coronavirus/sub_section.rb
@@ -14,7 +14,6 @@ class Coronavirus::SubSection < ApplicationRecord
   validates :page, presence: true
   validate :featured_link_must_be_in_content
 
-  before_create :content_super_group
   validate :all_content_groups_valid?
   after_create :save_validated_content_groups
 
@@ -65,6 +64,14 @@ class Coronavirus::SubSection < ApplicationRecord
     content_super_group.each do |group|
       group.coronavirus_sub_section_id = id
       group.save!
+    end
+  end
+
+  def reload_content_groups
+    existing_content_groups = content_groups.map(&:id)
+    if all_content_groups_valid?
+      save_validated_content_groups
+      existing_content_groups.each { |x| Coronavirus::ContentGroup.find(x).destroy! }
     end
   end
 end
--- a/app/models/coronavirus/sub_section.rb
+++ b/app/models/coronavirus/sub_section.rb
@@ -2,6 +2,8 @@ class Coronavirus::SubSection < ApplicationRecord
   self.table_name = "coronavirus_sub_sections"
 
   belongs_to :page, foreign_key: "coronavirus_page_id"
+  has_many :content_groups, dependent: :destroy, foreign_key: "coronavirus_sub_section_id"
+
   validates :title, :content, presence: true
   validates :page, presence: true
   validate :featured_link_must_be_in_content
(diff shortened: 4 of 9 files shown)


db/migrate/20210226152201_create_content_groups.rb
```diff
@@ -0,0 +1,11 @@
+class CreateContentGroups < ActiveRecord::Migration[6.0]
+  def change
+    create_table :content_groups do |t|
```
leenagupte, line 3: I'm a bit confused about the structure of this model? It has both a header and links. Does that mean the plan is to replace the SubSection model with a ContentGroup? 

Does the ContentGroup contain all of the content in a subsection? If so, how do we know which order the header and links should be presented in?

What happens if there is more than one header in the subsection?

leenagupte (commented):
I think I'm a bit unclear on the relationship between sub_sections and content_groups. I think it's a has one relationship? Is that correct?

I _think_ an assumption has been made that sub_sections will only have one heading in the content, but I this isn't true.

Did you consider having `sub_section_lines` instead of `content_groups`? So each sub_section has many `sub_section_lines`. Each sub_section_line has a "type" of either  "header" or "link" and has a "position" to define the order of the content in the sub_section.

Happy to chat about this in person.

db/migrate/20210226152201_create_content_groups.rb
JonathanHallam: So a content group is a header and its associated links. Pretty much matches what a content group is [here](https://github.com/alphagov/collections-publisher/blob/2c3acd32c13adfa6be393adadd8d3bc5ee8a89cd/app/presenters/coronavirus/sub_section_json_presenter.rb#L55). This means that we can remove functionality from the json presenter pretty directly. 

This means that a `SubSection` has many `ContentGroups` and each content group is aware of its position relative to one another. There's also no need for a `ContentGroup` to have a heading, the only required attribute is links. 

Sorry, should have made this clear in the commit messages.

leenagupte: Ok, that sort of makes sense. So would a featured link be classed as a ContentGroup that doesn't have a header?
Also, `links` are being stored as a string at the moment. Do you have plans to store them as an array?

JonathanHallam: yeah I think we'd like to store links as an array going forward. As for the featured links we haven't given that a huge amount of thought yet. I think we need to look into it a bit more so that we can understand the best way to store information on which link is featured. As is stands a featured link is just an array item within links but we don't have any way of knowing which content group a featured link for a sub section is in, only that one exists.

leenagupte: My main concern is that it's not obvious what a "Content group" is and I think that will trip people up in the future.

Also, isn't a featured link it's own content group?

JonathanHallam: I think you might be right there, do you think a 'lines' model would be better?

Not necessarily, you could have a content group like;
```
header: "header"
links: ["[link](/link)", "[link2](/link2)"]
```
Where `/link2` is the featured link. The content group as it is has no knowledge on whether it has a featured link in the "links" section.

leenagupte: If we define content_groups as what's in the content_item, I think featured links appear in their own group. As it in the [Work and financial support accordion](https://www.gov.uk/api/content/coronavirus):

```
{
   "title":"Work and financial support",
   "sub_sections":[
      {
         "list":[
            {
               "url":"/coronavirus/worker-support?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
               "label":"Getting financial help and staying safe at work",
               "description":"Get coronavirus (COVID-19) support if you’re working less or not working - and find out how to stay safe at work.",
               "featured_link":true
            }
         ],
         "title":null
      },
      {
         "list":[
            {
               "url":"/guidance/check-if-you-could-be-covered-by-the-coronavirus-job-retention-scheme?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
               "label":"Furlough: check if your employer can put you on temporary paid leave"
            },
            {
               "url":"/redundancy-your-rights?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
               "label":"Redundancy: your rights"
            },
            {
               "url":"/test-and-trace-support-payment?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
               "label":"Apply for a Test and Trace Support Payment"
            },
            {
               "url":"https://www.citizensadvice.org.uk/work/coronavirus-if-youre-worried-about-working?priority-taxon=774cee22-d896-44c1-a611-e3109cce8eae",
               "label":"Citizens Advice: advice if you’re worried about working"
            }
         ],
         "title":"Popular content"
      }
   ]
},
```

leenagupte: Then again, that might be because at the moment featured links only appear at the top of the accordion.

I don't know if a `lines` models is _better_. Just different. featured links do throw a bit of a spanner in the works, but I don't think they can be ignored. I think they need to be front and centre of any modelling decision as they can be complicated.

kevindew:
Woop amazing progress here on a tricky problem.

The first thing that I thought when seeing this was wondering whether you actually want an active record model for the the content. I don't think you need to persist that data to the database as it will end up duplicated (once in sub section other in content group). Instead you can use a plain old Ruby object (PORO) - you could even have it use [ActiveModel](https://guides.rubyonrails.org/active_model_basics.html#validations) if you want it to have it's own validations.

JonathanHallam:
> Woop amazing progress here on a tricky problem.
> 
> The first thing that I thought when seeing this was wondering whether you actually want an active record model for the the content. I don't think you need to persist that data to the database as it will end up duplicated (once in sub section other in content group). Instead you can use a plain old Ruby object (PORO) - you could even have it use [ActiveModel](https://guides.rubyonrails.org/active_model_basics.html#validations) if you want it to have it's own validations.

Thanks :)

Plan here would be to remove the content from the content field in subsections so that content groups is the only place it's stored. I think one of the last things that we'll do is to write a rake task that ports everything over as eventually we'd like it so that the content in sub sections isn't used.

kevindew:
> > Woop amazing progress here on a tricky problem.
> > The first thing that I thought when seeing this was wondering whether you actually want an active record model for the the content. I don't think you need to persist that data to the database as it will end up duplicated (once in sub section other in content group). Instead you can use a plain old Ruby object (PORO) - you could even have it use [ActiveModel](https://guides.rubyonrails.org/active_model_basics.html#validations) if you want it to have it's own validations.
> 
> Thanks :)
> 
> Plan here would be to remove the content from the content field in subsections so that content groups is the only place it's stored. I think one of the last things that we'll do is to write a rake task that ports everything over as eventually we'd like it so that the content in sub sections isn't used.

I see. I think you might find that throws up in it's own set of complications. You may have a bunch of trouble validating the input for instance.

kevindew:
Hey, I've found some time this afternoon to take a look at how this is progressing. Great job making so much progress with it 👍 

There are a few bits that I think you've nailed here such as:

- Moving logic out of SubSectionJsonPresenter
- Pulling errors out of content group as part of validation
- Nice clean integration with the SubSectionJsonPresenter

Whereas there's these bits that I'm a bit concerned about:
- Duplication of data storage
- ContentGroup seems a pretty hard to understand concept when the first one can skip a heading but others do, it strikes me as potentially a mistake in schema structuring that is proliferating
- ContentGroup as a DB model seems to require some pretty heavy overhead for low payoff.
- The concerns of the parsing are rather fragmented with the understanding of this rather spread out.

I had a bit of a playaround myself to see what might work better and came up with a class that could be nested in the Coronavirus::SubSection namespace to model this:

```ruby
class Coronavirus::SubSection::StructuredContent
  HEADER_PATTERN = PatternMaker.call(
    "starts_with hashes then perhaps_spaces then capture(text) and nothing_else",
    hashes: "#+",
    text: '\w.+',
  )

  LINK_PATTERN = PatternMaker.call(
    "starts_with perhaps_spaces within(sq_brackets,capture(label)) then perhaps_spaces and within(brackets,capture(url))",
    label: '\s*\w.+',
    url: '\s*(\b(https?)://)?[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]\s*',
  )

  Header = Struct.new(:text)
  Link = Struct.new(:label, :url)

  attr_reader :items

  def self.error_lines(raw_content)
    return [] unless raw_content

    raw_content.lines.filter_map do |line|
      stripped = line.strip
      next if stripped.empty?

      stripped if !HEADER_PATTERN.match(stripped) && !LINK_PATTERN.match(stripped)
    end
  end

  def self.parseable?(raw_content)
    error_lines(raw_content).empty?
  end

  def self.parse(raw_content)
    items = raw_content.lines.filter_map do |line|
      if (match = HEADER_PATTERN.match(line.strip))
        Header.new(match.named_captures["text"])
      elsif (match = LINK_PATTERN.match(line.strip))
        captures = match.named_captures
        Link.new(captures["label"], captures["url"])
      end
    end

    new(items)
  end

  def initialize(items)
    @items = items
  end

  def links
    items.select { |i| i.is_a?(Link) }
  end

  def groups
    items.each_with_object([]) do |item, memo|
      case item
      when Link
        if memo.empty?
          memo << { title: nil, links: [item] }
        else
          memo.last[:links] << item
        end
      when Header
        memo << { title: item.text, links: [] }
      end
    end
  end
end

```

You could hook that up to the model like so:

```ruby
class Coronavirus::SubSection < ApplicationRecord
  # ...
  after_initialize :populate_structured_content
  attr_reader :structured_content

  def content=(content)
    super.tap { populate_structured_content }
  end

  # ...
private

  def populate_structured_content
    @structured_content = if StructuredContent.parseable?(content)
                            StructuredContent.parse(content)
                          end
  end
end

```

and use it as part of the featured link validation:

```ruby
  def featured_link_must_be_in_content
    return if featured_link.blank? || !structured_content

    unless structured_content.links.any? { |l| l.url == featured_link }
      errors.add(:featured_link, "does not exist in accordion content")
    end
  end

```

and use it in the presenter:

```ruby

  def sub_sections
    sub_section.structured_content.groups do |group|
      {
        title: group[:title],
        list: group[:links].map { |link| build_link(link.title, link.url) }
      }
    end
  end
```

JonathanHallam:
Closed to make pretty fundamental changes as suggested in https://github.com/alphagov/collections-publisher/pull/1311#issuecomment-796957090

PR #1312, opened 2021-03-03
Raise exceptions and rollback failures in coronavirus pages

leenagupte:

# What?

Add transactions to the controller methods. Transactions will automatically rollback if something inside them throws an exception. Previously all of the exceptions were being swallowed and instead we were returning a list of errors, and had to manually raise an `ActiveRecord::Rollback` exception to make the transaction rollback.

To that end `ContentBuilder` and parts of `DraftUpdater` have also been updated to start raising their own exceptions.

### DIFF
--- a/app/controllers/coronavirus/announcements_controller.rb
+++ b/app/controllers/coronavirus/announcements_controller.rb
@@ -29,18 +29,15 @@ module Coronavirus
 
     def destroy
       announcement = page.announcements.find(params[:id])
-      message = { notice: helpers.t("coronavirus.announcements.destroy.success") }
 
       Announcement.transaction do
         announcement.destroy!
-
-        unless draft_updater.send
-          message = { alert: helpers.t("coronavirus.announcements.destroy.failed") }
-          raise ActiveRecord::Rollback
-        end
+        draft_updater.send
       end
 
-      redirect_to coronavirus_page_path(page.slug), message
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.destroy.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError
+      redirect_to coronavirus_page_path(page.slug), alert: helpers.t("coronavirus.announcements.destroy.failed")
     end
 
     def edit
--- a/app/controllers/coronavirus/announcements_controller.rb
+++ b/app/controllers/coronavirus/announcements_controller.rb
@@ -10,11 +10,21 @@ module Coronavirus
 
     def create
       @announcement = page.announcements.new(announcement_params)
-      if @announcement.save && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.create.success")
-      else
-        render :new
+
+      unless @announcement.valid?
+        render :new, status: :unprocessable_entity
+        return
       end
+
+      Announcement.transaction do
+        @announcement.save!
+        draft_updater.send
+      end
+
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.create.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash.now[:alert] = e.message
+      render :new, status: :internal_server_error
     end
 
     def destroy
--- a/app/controllers/coronavirus/announcements_controller.rb
+++ b/app/controllers/coronavirus/announcements_controller.rb
@@ -49,12 +49,22 @@ module Coronavirus
 
     def update
       @announcement = page.announcements.find(params[:id])
+      @announcement.assign_attributes(announcement_params)
 
-      if @announcement.update(announcement_params) && draft_updater.send
-        redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.update.success")
-      else
-        render :edit
+      unless @announcement.valid?
+        render :edit, status: :unprocessable_entity
+        return
       end
+
+      Announcement.transaction do
+        @announcement.update!(announcement_params)
+        draft_updater.send
+      end
+
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.update.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash.now[:alert] = e.message
+      render :edit, status: :internal_server_error
     end
 
   private
--- a/app/controllers/coronavirus/github_changes_controller.rb
+++ b/app/controllers/coronavirus/github_changes_controller.rb
@@ -8,13 +8,10 @@ module Coronavirus
     end
 
     def update
-      message = if draft_updater.send
-                  { notice: helpers.t("coronavirus.github_changes.update.success") }
-                else
-                  { alert: draft_updater.errors.to_sentence }
-                end
-
-      redirect_to github_changes_coronavirus_page_path(page.slug), message
+      draft_updater.send
+      redirect_to github_changes_coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.github_changes.update.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      redirect_to github_changes_coronavirus_page_path(page.slug), alert: e.message
     end
 
     def publish
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -52,18 +52,15 @@ module Coronavirus
 
     def destroy
       sub_section = page.sub_sections.find(params[:id])
-      message = { notice: helpers.t("coronavirus.sub_sections.destroy.success") }
 
       SubSection.transaction do
         sub_section.destroy!
-
-        unless draft_updater.send
-          message = { alert: helpers.t("coronavirus.sub_sections.destroy.failed") }
-          raise ActiveRecord::Rollback
-        end
+        draft_updater.send
       end
 
-      redirect_to coronavirus_page_path(page.slug), message
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.sub_sections.destroy.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError
+      redirect_to coronavirus_page_path(page.slug), alert: helpers.t("coronavirus.sub_sections.destroy.failed")
     end
 
   private
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -17,15 +17,13 @@ module Coronavirus
 
       SubSection.transaction do
         @sub_section.save!
-        raise ActiveRecord::Rollback unless draft_updater.send
+        draft_updater.send
       end
 
-      if draft_updater.errors.any?
-        flash.now["alert"] = draft_updater.errors.to_sentence
-        render :new, status: :internal_server_error
-      else
-        redirect_to coronavirus_page_path(page.slug), { notice: helpers.t("coronavirus.sub_sections.create.success") }
-      end
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.sub_sections.create.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash.now[:alert] = e.message
+      render :new, status: :internal_server_error
     end
 
     def edit
--- a/app/controllers/coronavirus/sub_sections_controller.rb
+++ b/app/controllers/coronavirus/sub_sections_controller.rb
@@ -41,15 +41,13 @@ module Coronavirus
 
       SubSection.transaction do
         @sub_section.save!
-        raise ActiveRecord::Rollback unless draft_updater.send
+        draft_updater.send
       end
 
-      if draft_updater.errors.any?
-        flash.now["alert"] = draft_updater.errors.to_sentence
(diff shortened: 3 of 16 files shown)


app/controllers/coronavirus/github_changes_controller.rb
```diff
@@ -8,13 +8,10 @@ def index
...
-                  { notice: helpers.t("coronavirus.github_changes.update.success") }
-                else
-                  { alert: draft_updater.errors.to_sentence }
-                end
-
-      redirect_to github_changes_coronavirus_page_path(page.slug), message
+      draft_updater.send
+      redirect_to github_changes_coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.github_changes.update.success")
+    rescue Coronavirus::Pages::DraftUpdater::DraftUpdaterError => e
+      redirect_to github_changes_coronavirus_page_path(page.slug), alert: e.message
```
kevindew, line 14: This controller looks good.

Minor change: 

`rescue Coronavirus::Pages::DraftUpdater::DraftUpdaterError => e` can be: `rescue Pages::DraftUpdater::DraftUpdaterError => e` as we're already in the Coronavirus namespace.

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -15,33 +15,25 @@ def content_builder
...
 
     def payload
-      if content_builder.success?
-        Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
-      else
-        raise DraftUpdaterError, content_builder.errors.to_sentence
-      end
+      Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
+    rescue Coronavirus::Pages::ContentBuilder::InvalidContentError,
+           Coronavirus::Pages::ContentBuilder::GitHubConnectionError => e
```
kevindew, line 20: Minor, these can be `ContentBuilder::InvalidContentError` since you're in `Cornavirus::Pages`

kevindew (commented):
This is looking awesome Leena 👍 

There's one part that I think is a little fishy and that we use a variety of exceptions with messages intended for humans (ContentBuilder::GitHubConnectionError, ContentBuilder::InvalidContentError, DraftUpdater::DraftUpdaterError). Typically exception messages are meant for developers and it'd be good to have a distinction between them (for example who knows if a content designer would know with some of the messages [RestClient::Exception](https://github.com/alphagov/collections-publisher/pull/1312/files#diff-19ba8d09bcecea2a8413817969609ffcc5c322077031c74fa9bfe366be50f0ccR70) could raise)

What I'd suggest is that since we're catching DraftUpdaterError at the controller level then this is the one exception that can have human friendly error messages. 

In practice this would probably mean changing bits like this:

```
      if missing_keys.any?
        raise InvalidContentError, "Invalid content - please recheck GitHub and add #{missing_keys.join(', ')}."
      end
```

to 

```
      raise InvalidContentError if missing_keys.any?
```

and then something the payload method being more like:

```
    def payload
      Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
    rescue ContentBuilder::InvalidContentError
      raise DraftUpdaterError, "Invalid content in GitHub YAML"
    rescue ContentBuilder::GitHubConnectionError
      raise DraftUpdaterError, "Unable to load data from GitHub"
    end
```

You'd lose some fidelity doing that but that seems ok, these exceptions shouldn't be intended as a debugging guide. We can always write that data to the log for a developer.

leenagupte:
@kevindew Thanks! That was really helpful 😃

leenagupte:
@kevindew thanks for the advice. I've made changes based on your suggestions to what I have so far. The only thing I haven't done is log the error with Sentry where the error originates. I'll do that when I remove the `error_handler` method from `DraftUpdater`.

kevindew:
> @kevindew thanks for the advice. I've made changes based on your suggestions to what I have so far. The only thing I haven't done is log the error with Sentry where the error originates. I'll do that when I remove the `error_handler` method from `DraftUpdater`.

Super thanks, I haven't had time to look yet but your Sentry comment caught my eye so I thought I'd reply before you start the work.

I'd wonder whether we should be raising these errors with Sentry (or at least some of them), since they don't represent a bug in the code and are handled situations. Given we don't have Sentry alerts configured I'm not sure we gain much from them and already rely on people to tell us when they need dev support.

app/services/coronavirus/pages/content_builder.rb
```diff
@@ -73,6 +72,8 @@ def required_hub_page_keys
 
     def github_raw_data
       YamlFetcher.new(page.raw_content_url).body_as_hash
+    rescue RestClient::Exception
+      raise GitHubConnectionError
```
kevindew, line 69: I notice no tests needed to be changed for this, which makes me wonder if this is untested code (unless it's covered at a high level)? 

Same thought on the next commit too.

app/services/coronavirus/pages/content_builder.rb
```diff
@@ -26,16 +26,6 @@ def data
...
-    def success?
-      data
-      errors.empty?
-    end
-
-    def add_error(error)
-      errors << error
-      errors.flatten!
-    end
-
```
kevindew, line 37: Nice 👍

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -1,6 +1,6 @@
 module Coronavirus::Pages
   class DraftUpdater
-    DraftUpdaterError = Class.new(StandardError)
+    class DraftUpdaterError < RuntimeError; end
```
kevindew, line 3: This is much more idiomatic than the previous version.

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -15,11 +15,13 @@ def content_builder
...
-      else
-        raise DraftUpdaterError, content_builder.errors.to_sentence
-      end
+      Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
+    rescue ContentBuilder::InvalidContentError
+      raise DraftUpdaterError, "Invalid content in one of the sub-sections"
+    rescue ContentBuilder::GitHubInvalidContentError
+      raise DraftUpdaterError, "Invalid content in GitHub YAML"
+    rescue ContentBuilder::GitHubConnectionError
+      raise DraftUpdaterError, "Unable to load content from GitHub"
```
kevindew, line 24: Should these be unit tested too? I wonder if some of these have no coverage.

app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -38,15 +38,5 @@ def discard
...
-
-    def errors
-      @errors ||= []
-    end
-
-    def error_handler(error, message = nil)
-      GovukError.notify(error, extra: { content_id: content_id, page_slug: page.slug })
-      errors << (message || error.message)
-      false
-    end
```
kevindew, line 55: 👍

app/controllers/coronavirus/announcements_controller.rb
```diff
@@ -10,27 +10,34 @@ def new
...
       end
+
+      Announcement.transaction do
+        @announcement.save!
+        draft_updater.send
+      end
+
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.announcements.create.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash[:alert] = e.message
```
kevindew, line 26: I think this will need to be `flash.now` - I think you might find that these flash currently sticks around for the request after this one.

app/controllers/coronavirus/announcements_controller.rb
```diff
@@ -10,27 +10,34 @@ def new
...
+      render :new
```
kevindew, line 27: This should return an error status code too, there is precedent with using `status: :internal_server_error` from the subsections: https://github.com/alphagov/collections-publisher/blob/2c3acd32c13adfa6be393adadd8d3bc5ee8a89cd/app/controllers/coronavirus/sub_sections_controller.rb#L25

app/controllers/coronavirus/sub_sections_controller.rb
```diff
@@ -17,15 +17,13 @@ def create
...
-      if draft_updater.errors.any?
-        flash.now["alert"] = draft_updater.errors.to_sentence
-        render :new, status: :internal_server_error
-      else
-        redirect_to coronavirus_page_path(page.slug), { notice: helpers.t("coronavirus.sub_sections.create.success") }
-      end
+      redirect_to coronavirus_page_path(page.slug), notice: helpers.t("coronavirus.sub_sections.create.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash.now[:alert] = e.message
+      render :new, status: :internal_server_error
```
kevindew, line 26: This looks a nice improvement.

app/controllers/coronavirus/timeline_entries_controller.rb
```diff
@@ -10,11 +10,20 @@ def new
...
+
+      TimelineEntry.transaction do
+        @timeline_entry.save!
+        draft_updater.send
+      end
+
+      redirect_to coronavirus_page_path(page.slug), notice: I18n.t("coronavirus.timeline_entries.create.success")
+    rescue Pages::DraftUpdater::DraftUpdaterError => e
+      flash.now[:alert] = e.message
+      render :new, status: :unprocessable_entity
```
kevindew, line 26: This should probably be an internal_server_error for both consistency with the other pages and to reflect that the error isn't the result of a mistake from the client.

kevindew (commented):
This looks great Leena, thanks for all your work here. This code is much easier to understand.

I'm wondering if there's just a few basic tests missing for the new code written where we end with no coverage (it seems to drop about 0.5% for this app with this PR)

You can view a report of the code coverage if you run `open -a "Google Chrome" ./coverage/index.html` after you've run RSpec.

app/controllers/coronavirus/announcements_controller.rb
leenagupte: I've updated the code to report a 5xx as suggested, though it's worth nothing that the error with publishing-api could be because some invalid content has been merged into GitHub, so whilst it's not an error create by this specific user it _is_ user error and could report an `unprocessable_entity` error.

leenagupte:
@kevindew Thanks for your review and comments. This PR is ready for a re-review.

A couple of things to mention. The number of commits has increased because, after the addition of controller tests, I felt that the "controller" commits were too large to easily review so they have been split up my "method".

Also, I'm not terribly convinced about how I've done the stubbing in the DraftUpdater tests, so some feedback on that will be greatly appreciated. On the bright-side the structure of the tests is in place now. Thanks for reminding me to add them 🤦.

spec/services/coronavirus/pages/draft_updater_spec.rb
```diff
@@ -54,4 +54,25 @@
       )
     end
   end
+
+  describe "#discard" do
+    it "raises an error if there isn't a draft to discard" do
+      stub_request(:post, "#{Plek.current.find('publishing-api')}/v2/content/#{page.content_id}/discard-draft")
+        .and_raise(GdsApi::HTTPUnprocessableEntity)
```
kevindew, line 61: Like a previous comment, it feels more idiomatic for request stubs to return status' that are converted by the underlying library to exceptions. You could also simplify this to:
```suggestion
      stub_any_publishing_api_discard_draft.to_return(status: 422)
```

spec/controllers/coronavirus/announcements_controller_spec.rb
```diff
@@ -98,20 +98,21 @@
...
+      expect { delete :destroy, params: announcement_params }
+        .to change { Coronavirus::Announcement.count }.by(-1)
     end
 
     it "doesn't delete the announcement if draft_updater fails" do
       stub_publishing_api_isnt_available
 
-      expect { subject }.to_not(change { page.reload.announcements.to_a })
+      expect { delete :destroy, params: announcement_params }
+        .to_not(change { page.reload.announcements.to_a })
```
kevindew, line 115:
```suggestion
        .not_to change({ Coronavirus::Announcement.count })
```

Sorry I know the previous code isn't yours. It seems a bit strange that we test a deletion in one way and the inverse of it in a totally different way.

spec/controllers/coronavirus/timeline_entries_controller_spec.rb
```diff
@@ -3,6 +3,8 @@
 RSpec.describe Coronavirus::TimelineEntriesController do
   include CoronavirusFeatureSteps
 
+  render_views
```
kevindew, line 6: 🎉

kevindew (commented):
This looks great and works great in my tests.

I've just got a couple of minor test amends suggestions and then it looks good to merge.

app/controllers/coronavirus/announcements_controller.rb
kevindew: I see what you mean there but I don't think that fits the intent of a client error status code. My understanding of 4xx status codes is that you can resolve them by changing your request, in particular for a 422 it'd be changing the data you submit. Since GitHub data being wrong is outside of what can be achieved in this request that indicates to me that our server is the system with the error, and thus in the 5xx realm.

spec/services/coronavirus/pages/draft_updater_spec.rb
leenagupte: This is much nicer! Thanks.

leenagupte:
@kevindew I've made the suggested tweaks and the PR is ready for a re-review. Thanks!

kevindew (approved):
Awesome 🚀

PR #1324, opened 2021-03-18
Move subsection content validation to model

JonathanHallam:
This attempts to remove validation that currently sits within the subsection json presenter. This is done by creating a structured content model within the subsection space that can create link objects and associated titles. The result of this is that the model layer has more information on what constitutes valid content.

We've been able to remove a lot of code from the json presenter and associated tests because the StructuredContent more closely resembles what's sent to the Publishing API and as such requires less logic.

Trello - https://trello.com/c/LwvWFlnk/160-move-subsection-presenter-validation-into-model-layer

### DIFF
--- a/app/models/coronavirus/sub_section.rb
+++ b/app/models/coronavirus/sub_section.rb
@@ -5,10 +5,34 @@ class Coronavirus::SubSection < ApplicationRecord
   validates :title, :content, presence: true
   validates :page, presence: true
   validate :featured_link_must_be_in_content
+  validate :all_structured_content_valid
+  after_initialize :populate_structured_content
+  attr_reader :structured_content
+
+  def content=(content)
+    super.tap { populate_structured_content }
+  end
 
   def featured_link_must_be_in_content
-    if featured_link.present? && !content.include?(featured_link)
+    return if featured_link.blank? || !structured_content
+
+    unless structured_content.links.any? { |l| l.url == featured_link }
       errors.add(:featured_link, "does not exist in accordion content")
     end
+  end
+
+private
+
+  def populate_structured_content
+    @structured_content = if StructuredContent.parseable?(content)
+                            StructuredContent.parse(content)
+                          end
+  end
+
+  def all_structured_content_valid
+    StructuredContent.error_lines(content).each do |line|
+      errors.add(:content, "unable to parse markdown: #{line}")
+    end
+    errors.present?
   end
 end
--- a/app/models/coronavirus/sub_section/structured_content.rb
+++ b/app/models/coronavirus/sub_section/structured_content.rb
@@ -1,0 +1,56 @@
+class Coronavirus::SubSection::StructuredContent
+  HEADER_PATTERN = PatternMaker.call(
+    "starts_with hashes then perhaps_spaces then capture(text) and nothing_else",
+    hashes: "#+",
+    text: '\w.+',
+  )
+
+  LINK_PATTERN = PatternMaker.call(
+    "starts_with perhaps_spaces within(sq_brackets,capture(label)) then perhaps_spaces and within(brackets,capture(url))",
+    label: '\s*\w.+',
+    url: '\s*(\b(https?)://)?[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]\s*',
+  )
+
+  Header = Struct.new(:text)
+  Link = Struct.new(:label, :url)
+
+  attr_reader :items
+
+  def self.error_lines(raw_content)
+    return [] unless raw_content
+
+    raw_content.lines.filter_map do |line|
+      stripped = line.strip
+      next if stripped.empty?
+
+      stripped if HEADER_PATTERN.match(stripped).blank? && LINK_PATTERN.match(stripped).blank?
+    end
+  end
+
+  def self.parseable?(raw_content)
+    raw_content.present? && error_lines(raw_content).empty?
+  end
+
+  def self.parse(raw_content)
+    items = raw_content.lines.filter_map do |line|
+      header_match = HEADER_PATTERN.match(line.strip)
+      link_match = LINK_PATTERN.match(line.strip)
+
+      if header_match
+        Header.new(header_match[:text])
+      elsif link_match
+        Link.new(link_match[:label], link_match[:url])
+      end
+    end
+
+    new(items)
+  end
+
+  def initialize(items)
+    @items = items
+  end
+
+  def links
+    items.select { |i| i.is_a?(Link) }
+  end
+end
--- a/app/presenters/coronavirus/sub_section_json_presenter.rb
+++ b/app/presenters/coronavirus/sub_section_json_presenter.rb
@@ -1,21 +1,6 @@
 # frozen_string_literal: true
 
 class Coronavirus::SubSectionJsonPresenter
-  class MarkdownInvalidError < RuntimeError; end
-
-  HEADER_PATTERN = PatternMaker.call(
-    "starts_with hashes then perhaps_spaces then capture(title) and nothing_else",
-    hashes: "#+",
-    title: '\w.+',
-  )
-
-  LINK_PATTERN = PatternMaker.call(
-    "starts_with perhaps_spaces within(sq_brackets,capture(label)) then perhaps_spaces and within(brackets,capture(url))",
-    label: '\s*\w.+',
-    url: '\s*(\b(https?)://)?[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]\s*',
-  )
-  # Url pattern from https://stackoverflow.com/a/163684/1014251
-
   attr_reader :sub_section, :priority_taxon
 
   def initialize(sub_section, priority_taxon = nil)
@@ -34,65 +19,22 @@ class Coronavirus::SubSectionJsonPresenter
   end
 
   def sub_sections
-    content_groups.map { |content_group| sub_section_hash_from_content_group(content_group) }
-  end
-
-  # Groups the sub section content into an array of arrays, such that:
-  #   - links and texts are separated and each is put into an inner array
-  #   - each title starts a new inner array containing that title as the first element
-  #     and then the links between that title and the next title (or the end of the content)
-  #   - if there are no titles, all the links go into a single inner array
-  def content_groups
-    sub_section.content.lines.each_with_object([]) do |line, sections|
-      sections << [] if sections.empty? || is_header?(line)
-      line.strip!
-      sections.last << line
-    end
-  end
-
-  # Converts:
-  #
-  #   [header, link, link]
-  #
-  # into:
-  #
-  #    {
-  #      title: <text from header>,
-  #      list: [
-  #        {
-  #          label: <first link label>,
-  #          url: <first link url>
-  #        },
-  #        {
-  #          label: <second link label>,
-  #          url: <second link url>
-  #        }
-  #    {
-  def sub_section_hash_from_content_group(content_group)
-    content_group.each_with_object({ title: nil }) do |line, hash|
-      if is_header?(line)
-        title = HEADER_PATTERN.match(line).named_captures["title"]
-        hash[:title] = title
-      elsif is_link?(line)
-        hash[:list] ||= []
-        hash[:list] << build_link(line)
-      else
-        raise MarkdownInvalidError, "Unable to parse markdown: '#{line}'"
+    sub_section.structured_content.items.each_with_object([]) do |item, memo|
+      case item
+      when Coronavirus::SubSection::StructuredContent::Link
+        if memo.empty?
+          memo << { title: nil, list: [build_link(item.label, item.url)] }
+        else
+          memo.last[:list] << build_link(item.label, item.url)
+        end
+      when Coronavirus::SubSection::StructuredContent::Header
+        memo << { title: item.text, list: [] }
       end
     end
   end
 
-  def is_header?(text)
-    HEADER_PATTERN =~ text
-  end
-
-  def is_link?(text)
-    LINK_PATTERN =~ text
-  end
-
-  def build_link(element)
-    link = LINK_PATTERN.match(element).named_captures.symbolize_keys
-    link.transform_values!(&:strip)
+  def build_link(label, url)
+    link = { label: label, url: url }
     if subtopic_paths.keys.include?(link[:url])
       link[:description] = description_from_raw_content(link[:url])
       link[:featured_link] = true
--- a/app/services/coronavirus/pages/content_builder.rb
+++ b/app/services/coronavirus/pages/content_builder.rb
@@ -2,7 +2,6 @@ module Coronavirus::Pages
   class ContentBuilder
     class GitHubConnectionError < RuntimeError; end
     class GitHubInvalidContentError < RuntimeError; end
-    class InvalidContentError < RuntimeError; end
 
     attr_reader :page
 
@@ -81,8 +80,6 @@ module Coronavirus::Pages
       page.sub_sections.order(:position).map do |sub_section|
         presenter = Coronavirus::SubSectionJsonPresenter.new(sub_section, page.content_id)
         presenter.output
-      rescue Coronavirus::SubSectionJsonPresenter::MarkdownInvalidError => e
-        raise InvalidContentError, e.message
       end
     end
 
--- a/spec/controllers/coronavirus/sub_sections_controller_spec.rb
+++ b/spec/controllers/coronavirus/sub_sections_controller_spec.rb
@@ -54,21 +54,37 @@ RSpec.describe Coronavirus::SubSectionsController do
     end
 
     context "when the subsection is invalid" do
-      let(:title) { "" }
+      context "when the title is invalid" do
+        let(:title) { "" }
 
-      it "doesn't create a subsection" do
-        expect { post :create, params: params }
-          .not_to(change { Coronavirus::SubSection.count })
+        it "doesn't create a subsection" do
+          expect { post :create, params: params }
+            .not_to(change { Coronavirus::SubSection.count })
+        end
+
+        it "returns an unprocessable entity response" do
+          post :create, params: params
+          expect(response).to have_http_status(:unprocessable_entity)
+        end
+
+        it "renders the errors" do
+          post :create, params: params
+          expect(response.body).to include(CGI.escapeHTML("Title can't be blank"))
+        end
       end
 
-      it "returns an unprocessable entity response" do
-        post :create, params: params
-        expect(response).to have_http_status(:unprocessable_entity)
-      end
+      context "when the content is invalid" do
+        let(:content) { "###Title \n [label](/brexit" }
 
-      it "renders the errors" do
-        post :create, params: params
-        expect(response.body).to include(CGI.escapeHTML("Title can't be blank"))
+        it "doesn't create a subsection" do
+          expect { post :create, params: params }
+            .not_to(change { Coronavirus::SubSection.count })
+        end
+
+        it "renders the errors" do
+          post :create, params: params
+          expect(response.body).to include(CGI.escapeHTML("unable to parse markdown"))
+        end
       end
     end
 
@@ -126,21 +142,37 @@ RSpec.describe Coronavirus::SubSectionsController do
     end
 
     context "when the subsection is invalid" do
-      let(:title) { "" }
+      context "when the title is invalid" do
+        let(:title) { "" }
 
-      it "doesn't update a subsection" do
-        expect { patch :update, params: params }
-          .not_to(change { sub_section.reload.title })
+        it "doesn't update a subsection" do
+          expect { patch :update, params: params }
+            .not_to(change { sub_section.reload.title })
+        end
+
+        it "returns an unprocessable entity response" do
+          patch :update, params: params
+          expect(response).to have_http_status(:unprocessable_entity)
+        end
+
+        it "renders the errors" do
+          patch :update, params: params
+          expect(response.body).to include(CGI.escapeHTML("Title can't be blank"))
+        end
       end
 
-      it "returns an unprocessable entity response" do
-        patch :update, params: params
-        expect(response).to have_http_status(:unprocessable_entity)
-      end
+      context "when the content is invalid" do
+        let(:content) { "###Title \n [label](/brexit" }
 
-      it "renders the errors" do
-        patch :update, params: params
-        expect(response.body).to include(CGI.escapeHTML("Title can't be blank"))
+        it "doesn't update the structured content" do
+          expect { patch :update, params: params }
+            .not_to(change { sub_section.reload.structured_content })
+        end
+
+        it "renders the errors" do
+          patch :update, params: params
+          expect(response.body).to include(CGI.escapeHTML("unable to parse markdown"))
+        end
       end
     end
 
--- a/spec/models/coronavirus/sub_section/structured_content_spec.rb
+++ b/spec/models/coronavirus/sub_section/structured_content_spec.rb
@@ -1,0 +1,66 @@
+require "rails_helper"
+
+RSpec.describe Coronavirus::SubSection::StructuredContent do
+  let(:content) { "###title \n [test](/coronavirus)" }
+
+  describe "#parseable?" do
+    it "returns true if content is valid" do
+      expect(described_class.parseable?(content)).to be true
+    end
+
+    it "returns false if hashes are missing from title" do
+      content = "title \n [test](/coronavirus)"
+      expect(described_class.parseable?(content)).to be false
+    end
+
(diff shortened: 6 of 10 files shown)


app/presenters/coronavirus/sub_section_json_presenter.rb
```diff
@@ -3,19 +3,6 @@
 class Coronavirus::SubSectionJsonPresenter
   class MarkdownInvalidError < RuntimeError; end
```
leenagupte, line 5: This exception definition can be removed as it's no longer being raised.

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
```diff
@@ -94,157 +86,6 @@
...
-
-    context "with unknown content" do
-      let(:group) { %w[unknown] }
-
-      it "adds an error" do
-        expect { sub_section_hash }.to raise_error(Coronavirus::SubSectionJsonPresenter::MarkdownInvalidError)
-      end
-    end
-
-    context "when link is to a subtopic path" do
```
leenagupte, line 194: I think these tests should be put back as they test that links to a hub page are being set as "featured".

spec/services/coronavirus/pages/content_builder_spec.rb
```diff
@@ -87,12 +87,6 @@
       expect(subject.sub_sections_data).to eq []
     end
 
-    it "raises an error if the sub-section has invalid content" do
-      create(:coronavirus_sub_section, page: page, content: "I am invalid")
-
-      expect { subject.sub_sections_data }.to raise_error(Coronavirus::Pages::ContentBuilder::InvalidContentError)
```
leenagupte, line 93: The exception definition for `InvalidContentError` needs to be removed as well: https://github.com/alphagov/collections-publisher/blob/23e7800a50ff4848c71b5e7baa78653cd50bac11/app/services/coronavirus/pages/content_builder.rb#L5

As does the rescuing and re-throwing of the exception: https://github.com/alphagov/collections-publisher/blob/23e7800a50ff4848c71b5e7baa78653cd50bac11/app/services/coronavirus/pages/content_builder.rb#L84-L85

leenagupte (commented):
I think the approach is a good one. The idea to check that the content stored as a string can be modelled correctly is great! 🎉 

Well done for figuring out all of the gnarly validation in the presenter and _trying_ not to replicate it in the model.

I've added some inline comments. Most are pretty minor and should be easy to fix.

There are a couple of things I'd like to highlight though. There a couple of places I've asked for more of the "why did we do this?" to be added to the commit message. This will help future us build on this code later without having to spend too much time trying to figure out what's happening. 

That leads me to the second thing. There are two methods in `StructuredContent` are quite difficult to follow and figure out what they're doing and what they return. The methods are `self.parse` and `group`. There are a few things that can be done to remedy this:
- Try and simplify the code
- Make the tests more explicit
- Split the methods out into their own commits and use the commit message to explain what's happening.

I know that some of the naming of things has been taken from the old presenter code, but this is a good opportunity to make some complex code easier to understand.

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+    raw_content.lines.filter_map do |line|
+      stripped = line.strip
+      next if stripped.empty?
+
+      stripped if !HEADER_PATTERN.match(stripped) && !LINK_PATTERN.match(stripped)
+    end
+  end
+
+  def self.parseable?(raw_content)
+    !raw_content.nil? && error_lines(raw_content).empty?
```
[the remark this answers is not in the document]
JonathanHallam: We've swapped this over to use `present?`. For some reason getting rid of it makes a load of tests fail, I think because some tests used to work without `raw_content`. I think this would be a good thing to fix but I think it might be best to add it in a separate pr.

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
+class Coronavirus::SubSection::StructuredContent
+  HEADER_PATTERN = PatternMaker.call(
+    "starts_with hashes then perhaps_spaces then capture(text) and nothing_else",
+    hashes: "#+",
+    text: '\w.+',
```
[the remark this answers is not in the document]
JonathanHallam: The title field actually comes from [`groups`](https://github.com/alphagov/collections-publisher/blob/dc4153154bc4fe05b0e997552e628e86500e597f/app/models/coronavirus/sub_section/structured_content.rb#L60) rather than here so no change needed :)

We'll take a look at the `PatternMaker` stuff later today.

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def initialize(items)
+    @items = items
+  end
+
+  def links
+    items.select { |i| i.is_a?(Link) }
+  end
+
+  def self.lookup_description(_url)
```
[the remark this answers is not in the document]
JonathanHallam: good spot, thanks :)

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,82 @@
+require "rails_helper"
+
+RSpec.describe Coronavirus::SubSection::StructuredContent do
+  let(:content) { "###title \n [test](/coronavirus)" }
+  let(:featured_link) { nil }
```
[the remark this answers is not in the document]
JonathanHallam: nice one thanks, have removed!

app/models/coronavirus/sub_section.rb
```diff
@@ -5,10 +5,38 @@ class Coronavirus::SubSection < ApplicationRecord
...
+
+private
+
+  def populate_structured_content
+    @structured_content = if StructuredContent.parseable?(content)
+                            StructuredContent.parse(content)
+                          end
+  end
+
+  def all_structured_content_valid?
```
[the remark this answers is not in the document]
JonathanHallam: nice one, we've cut some code out here, thanks :)

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+  attr_reader :items
+
+  def self.error_lines(raw_content)
+    return [] unless raw_content
+
+    raw_content.lines.filter_map do |line|
+      stripped = line.strip
+      next if stripped.empty?
+
+      stripped if !HEADER_PATTERN.match(stripped) && !LINK_PATTERN.match(stripped)
```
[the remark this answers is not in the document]
JonathanHallam: We've changed it to use `blank?` rather than present to keep the logic happy but this is now working :)

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      structured_content = described_class.parse(content)
+      expect(structured_content.groups.first[:title]).to be nil
+      expect(structured_content.groups.count).to eq(2)
+    end
+
+    it "assigns links to correct groups" do
+      content = "#title \n [test](/coronavirus) \n #title2 \n [test2](/government)"
+      structured_content = described_class.parse(content)
+      expect(structured_content.groups.first[:list].first.label).to eq("test")
+      expect(structured_content.groups.second[:list].first.label).to eq("test2")
```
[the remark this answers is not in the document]
JonathanHallam: good shout. Have now done this :)

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  describe "#groups" do
+    it "sets the title" do
+      structured_content = described_class.parse(content)
+      expect(structured_content.groups.first[:title]).to eq("title")
+    end
+
+    it "creates multiple groups given multiple titles" do
+      content = "#title \n [test](/coronavirus) \n #title2 \n [test2](/government)"
+      structured_content = described_class.parse(content)
+      expect(structured_content.groups.count).to eq(2)
```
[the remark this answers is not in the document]
JonathanHallam: done!

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(described_class.error_lines(content)).to eq(%w[[test](/coronavirus])
+    end
+
+    it "can return multiple errors" do
+      content = "title \n [test](/coronavirus"
+      expect(described_class.error_lines(content).length).to eq(2)
+    end
+  end
+
+  describe "#groups" do
```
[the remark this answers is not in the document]
JonathanHallam: done! Will also do the other work to make the groups method itself a bit clearer :)

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def self.parse(raw_content)
+    items = raw_content.lines.filter_map do |line|
+      if (match = HEADER_PATTERN.match(line.strip))
+        Header.new(match.named_captures["text"])
+      elsif (match = LINK_PATTERN.match(line.strip))
+        captures = match.named_captures
+        Link.new(captures["label"], captures["url"])
+      end
```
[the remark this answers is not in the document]
JonathanHallam: So we're not expecting the caller to know the difference. This is done by the `groups` method further down and I hope later changes to that can make this a bit clearer. 

We've made some changes to this method that I think make it a lot easier to understand what's going on. We haven't broken it out into separate methods but please let us know if you think we need to go further than we have :)

app/models/coronavirus/sub_section/structured_content.rb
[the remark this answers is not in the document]
leenagupte: Jon isn't this a method that you've added? Doesn't that mean that you've created the tests? Something somewhere isn't passing the data correctly, so either the tests haven't been set-up correctly, or this points to a larger problem somewhere in the code that really needs to be fixed.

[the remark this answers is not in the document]
JonathanHallam: As it turns out it's not to do with the test data but rather that a new subsection is created completely blank when the page is loaded. This means that the issue can be seen by removing the `raw_content.present? &&` portion of this line, booting up and trying to create a new sub section. You'll hit a standard rails error because it tries to create a subsection but falls over when it tries to call `lines` on `nil` [here](https://github.com/alphagov/collections-publisher/pull/1324/files#diff-277a4e00cb4c1ba05ddb7f31004e294012a3c761fc5c96527070045d35c1c324R35). 

I don't really understand why this happens but I'll look into it. I don't think it's a problem that we've introduced here but rather something new that this code has exposed.

app/models/coronavirus/sub_section/structured_content.rb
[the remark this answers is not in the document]
JonathanHallam: I've changed the parse tests to check for specific output after a conversation on slack about the complexity that faffing about with PatternMaker requires 😂

spec/controllers/coronavirus/sub_sections_controller_spec.rb
```diff
@@ -70,6 +70,20 @@
         post :create, params: params
         expect(response.body).to include(CGI.escapeHTML("Title can't be blank"))
       end
+
+      context "when the content is invalid" do
```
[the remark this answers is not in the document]
JonathanHallam: done!

app/presenters/coronavirus/sub_section_json_presenter.rb
JonathanHallam: done!

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    it "returns false if hashes are missing from title" do
+      content = "title \n [test](/coronavirus)"
+      expect(described_class.parseable?(content)).to be false
+    end
+
+    it "returns false if a link is malformed" do
+      content = "###title \n [test](/coronavirus"
+      expect(described_class.parseable?(content)).to be false
+    end
```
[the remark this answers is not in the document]
JonathanHallam: done!

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    new(items)
+  end
+
+  def initialize(items)
+    @items = items
+  end
+
+  def links
```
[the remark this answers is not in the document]
JonathanHallam: sorted :)

spec/services/coronavirus/pages/content_builder_spec.rb
JonathanHallam: good spot. Thanks :)

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
JonathanHallam: Sorry, just so I understand, hub pages are meant to be featured whether they're listed as featured or not?

leenagupte: Yes. It's a bit confusing I know, but the code explicitly sets links to hub pages as featured links: https://github.com/alphagov/collections-publisher/blob/9e8fe1f11407d0191a14170b6ee445d90832a4d6/app/presenters/coronavirus/sub_section_json_presenter.rb#L96-L98

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+      when Link
+        if memo.empty?
+          memo << { title: nil, list: [item] }
+        else
+          memo.last[:list] << item
+        end
+      when Header
+        memo << { title: item.text, list: [] }
+      end
+    end
```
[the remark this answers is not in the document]
JonathanHallam: I've added an explanation to the commit message and done an arguably ropey rename that I'm hoping make things a bit clearer :)

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def links
+    items.select { |i| i.is_a?(Link) }
+  end
+
+  def self.lookup_description(_url)
+    nil
+  end
+
+  def groups
```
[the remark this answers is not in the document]
JonathanHallam: done :)

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
JonathanHallam: Okay cool, in that case, I'll add in a test for `build_link` to do this because that's where it's actually being done and that for some reason wasn't tested directly before.

app/models/coronavirus/sub_section.rb
```diff
@@ -5,10 +5,38 @@ class Coronavirus::SubSection < ApplicationRecord
   validates :title, :content, presence: true
   validates :page, presence: true
   validate :featured_link_must_be_in_content
+  validate :all_structured_content_valid?
+  after_initialize :populate_structured_content
+  attr_reader :structured_content
+
+  def content=(content)
+    super.tap { populate_structured_content }
```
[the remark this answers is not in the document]
JonathanHallam: and done! That's a wrap 😌

app/models/coronavirus/sub_section/structured_content.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def initialize(items)
+    @items = items
+  end
+
+  def links
+    items.select { |i| i.is_a?(Link) }
+  end
+
+  def create_content_groups_from_links_and_headings
```
leenagupte, line 57: This method is not used to validate the content is it? 

If I understand correctly the purpose of this method is purely to "present the data", i.e. to put the structured content in the format expected by the frontend (content store etc)? 

I think then, that it makes more sense for this method to be in the `SubSectionJsonPresenter` where the rest of the payload is being structured. 

Hopefully doing that will mean the method can be less specific as its presence in a presenter will explain much of what it is doing.

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  describe "#create_content_groups_from_links_and_headings" do
+    it "creates multiple groups given multiple titles" do
+      content = "#title \n [test](/coronavirus) \n #title2 \n [test2](/government)"
+      expected_output = [
+        {
+          title: "title",
+          list: [Coronavirus::SubSection::StructuredContent::Link.new("test", "/coronavirus")],
```
leenagupte, line 65: I've suggested above that this method should be moved to the presenter. I think that we do that, then the equivalent to this test, can actually test the structure of the list i.e. that it contains `url`s and `label`s

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+          title: "title2",
+          list: [Coronavirus::SubSection::StructuredContent::Link.new("test2", "/government")],
+        },
+      ]
+      structured_content = described_class.parse(content)
+      expect(structured_content.create_content_groups_from_links_and_headings).to eq(expected_output)
+    end
+  end
+
+  describe "#links" do
```
leenagupte, line 93: Yay! Test!

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
```diff
@@ -86,6 +86,41 @@
...
+        expected_output = { label: "title", url: "/link", description: "description", featured_link: true }
+        expect(subject.build_link("title", "/link")).to eq(expected_output)
+      end
+    end
+
+    context "when link is to a subtopic path" do
+      let(:business) { create :coronavirus_page, :business }
+      let(:fixture_path) { Rails.root.join "spec/fixtures/coronavirus_landing_page.yml" }
+      let(:description) { "Find out about the government response to coronavirus (COVID-19) and what you need to do." }
+      let(:path) { business.base_path }
```
leenagupte: Minor: I don't think this value is needed as it's easy enough for the expectation to call `business.base_path`.

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
```diff
@@ -86,6 +86,41 @@
...
+        sub_section.featured_link = "/link"
+        expected_output = { label: "title", url: "/link", description: "description", featured_link: true }
+        expect(subject.build_link("title", "/link")).to eq(expected_output)
+      end
+    end
+
+    context "when link is to a subtopic path" do
+      let(:business) { create :coronavirus_page, :business }
+      let(:fixture_path) { Rails.root.join "spec/fixtures/coronavirus_landing_page.yml" }
+      let(:description) { "Find out about the government response to coronavirus (COVID-19) and what you need to do." }
```
leenagupte: I was unsure where this description was coming from so I'd consider renaming this to something like `subtopic_content_item_description` or `hub_page_description`.

leenagupte (commented):
Thanks for addressing the comments and especially for updating the commit messages 🎉 .

The changes looks really good, and this is close to being ready. I've added a few inline comments, most of them are minor. 

The main comment I want to highlight here is the recommendation about moving the `create_content_groups_from_links_and_headings` method to `SubSectionJsonPresenter`. 

This method always felt out of place in the model, and I think that's because it's doing the work of a presenter. It's not used in the validations and the only place the method is called is `SubSectionJsonPresenter` which is why I think it should be moved there.
Apologies for not catching this in the previous review.

spec/models/coronavirus/sub_section/structured_content_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    it "returns false if there's a plaintext line in content" do
+      content = "###title \n [test](/coronavirus) \n plan text \n [another](/link)"
+      expect(described_class.parseable?(content)).to be false
+    end
+  end
+
+  describe "#parse" do
+    let(:parsed_content) { described_class.parse(content) }
+    it "creates the expected output" do
```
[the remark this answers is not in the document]
JonathanHallam: done :)

app/models/coronavirus/sub_section/structured_content.rb
JonathanHallam: done :)

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
JonathanHallam: sorted :)

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
JonathanHallam: done :)

spec/models/coronavirus/sub_section/structured_content_spec.rb
JonathanHallam: 😎

spec/models/coronavirus/sub_section/structured_content_spec.rb
JonathanHallam: done!

app/presenters/coronavirus/sub_section_json_presenter.rb
```diff
@@ -29,70 +14,27 @@ def output
...
+    sub_section.structured_content.items.each_with_object([]) do |item, memo|
+      case item
+      when Coronavirus::SubSection::StructuredContent::Link
+        if memo.empty?
+          memo << { title: nil, list: [build_link(item.label, item.url)] }
+        else
+          memo.last[:list] << build_link(item.label, item.url)
+        end
+      when Coronavirus::SubSection::StructuredContent::Header
+        memo << { title: item.text, list: [] }
```
leenagupte, line 31: This is much nicer!

app/presenters/coronavirus/sub_section_json_presenter.rb
```diff
@@ -29,70 +14,27 @@ def output
     @output ||=
       {
         title: title,
-        sub_sections: sub_sections,
+        sub_sections: create_sub_sections,
```
leenagupte, line 17: Minor: I don't think you needed to rename the method.

app/services/coronavirus/pages/content_builder.rb
```diff
@@ -81,8 +81,6 @@ def sub_sections_data
       page.sub_sections.order(:position).map do |sub_section|
         presenter = Coronavirus::SubSectionJsonPresenter.new(sub_section, page.content_id)
         presenter.output
-      rescue Coronavirus::SubSectionJsonPresenter::MarkdownInvalidError => e
-        raise InvalidContentError, e.message
```
leenagupte, line 85: Something has gone a bit wrong with the rebase as I think this change is supposed to happen in the "Remove test from content builder" commit.

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
```diff
@@ -94,153 +86,40 @@
...
-      let(:title_one) { random_title }
-      let(:title_two) { random_title }
-      let(:content) { [link_one, title_one, link_two, title_two, link_three].join("\n") }
-
-      it "returns the content in grouped arrays" do
-        expect(subject.content_groups).to eq([[link_one], [title_one, link_two], [title_two, link_three]])
+  describe "#create_sub_sections" do
+    context "given multiple titles" do
+      let(:content) { "#title \n [test](/coronavirus) \n #title2 \n [test2](/government)" }
+      it "creates multiple groups" do
```
leenagupte, line 126: This is a nice test! 💯

leenagupte (commented):
The approach looks really good! Well done! 🎉 

There's just a little bit of tidy up left to do:
1. The commit message for the first commit is out of date now, and should be updated
2. As mentioned in an inline comment, a change to remove `InvalidContentError` is in the wrong commit.

It's almost there!

app/presenters/coronavirus/sub_section_json_presenter.rb
JonathanHallam: undone, thanks

app/services/coronavirus/pages/content_builder.rb
JonathanHallam: yup, not sure how that happened! Good spot, thanks

spec/presenters/coronavirus/sub_section_json_presenter_spec.rb
JonathanHallam: thanks :D

app/presenters/coronavirus/sub_section_json_presenter.rb
JonathanHallam: Thank you :) just going to fix up the commit messages now!

leenagupte (approved):
🎉 🥇 Thanks for persevering with this!

PR #1333, opened 2021-03-26
Change coronavirus_announcements schema for path -> url change

kevindew:
Trello: https://trello.com/c/CRwjuyqN/278-allow-external-links-from-announcements

This PR does the database changes needed to prepare the Coronavirus::Announcements model for a change from accepting a path to accepting a URL. It also resolves some ambiguity in the DB schema about what can be accepted by setting key fields to no longer accept null.

This is a somewhat dangerous PR, as the migration involves column changes (particularly rename) it can cause the app to error if it is being used during deployment. Therefore I will plan to carefully deploy it while the app is not being used (this seemed easier than a backfill process given this app is only sporadically used during a working day)

I've opened this a small PR to reflect that it's a bit dangerous to deploy so it doesn't get entangled with wider application changes. This reduces the risk that this will be deployed with bugs.

### DIFF
--- a/app/models/coronavirus/announcement.rb
+++ b/app/models/coronavirus/announcement.rb
@@ -1,6 +1,10 @@
 class Coronavirus::Announcement < ApplicationRecord
   self.table_name = "coronavirus_announcements"
 
+  # Maintain use of a path attribute in the model while we transition the app
+  # to use a url attribute instead.
+  alias_attribute :path, :url
+
   belongs_to :page, foreign_key: "coronavirus_page_id"
   validates :title, :path, presence: true
   validates :page, presence: true
--- a/db/migrate/20210325171835_improve_coronavirus_announcements_schema.rb
+++ b/db/migrate/20210325171835_improve_coronavirus_announcements_schema.rb
@@ -1,0 +1,19 @@
+class ImproveCoronavirusAnnouncementsSchema < ActiveRecord::Migration[6.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        # Change from string to allow URL's longer than 255 characters
+        change_column :coronavirus_announcements, :url, :text, null: false
+      end
+
+      dir.down do
+        change_column :coronavirus_announcements, :url, :string
+      end
+    end
+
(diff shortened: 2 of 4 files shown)


db/migrate/20210325171835_improve_coronavirus_announcements_schema.rb
```diff
@@ -0,0 +1,20 @@
...
+      dir.down do
+        change_column :coronavirus_announcements, :url, :string
+      end
+    end
+
+    # These fields are all required so shouldn't allow null
+    change_column_null :coronavirus_announcements, :coronavirus_page_id, false
+    change_column_null :coronavirus_announcements, :title, false
+    change_column_null :coronavirus_announcements, :url, false
+    change_column_null :coronavirus_announcements, :published_at, false
```
Rosa-Fox, line 18: I am confused... shouldn't `published_at` be able to be `null`? It looks fine in the schema below.

app/models/coronavirus/announcement.rb
```diff
@@ -1,6 +1,10 @@
 class Coronavirus::Announcement < ApplicationRecord
   self.table_name = "coronavirus_announcements"
 
+  # Maintain use of a path attribute in the model while we transition the app
+  # to use a url attribute instead.
+  alias_attribute :path, :url
```
Rosa-Fox, line 6: Nice!

Rosa-Fox (commented):
Looks good and works locally :) I left one question!

db/migrate/20210325171835_improve_coronavirus_announcements_schema.rb
kevindew: Oh sorry - that was left over from a quick commit I did so I could undo setting that to null false in integration: https://github.com/alphagov/collections-publisher/commit/f4b03223a777341bc1fc40f7548559a04cfb6d89

I've removed that now.

Rosa-Fox: 👍

Rosa-Fox (approved):

PR #1336, opened 2021-03-29
[Dangerous deployment] Improve schema configuration for coronavirus_announcements fields

kevindew:
Trello: https://trello.com/c/CRwjuyqN/278-allow-external-links-from-announcements

As a bonus (or a nuisance) I did some follow-up work on https://github.com/alphagov/collections-publisher/pull/1335 to set the published_at field to be a date (and rename to published_on as per Rails convention) and remove the ability to set a null position. This finishes work started in https://github.com/alphagov/collections-publisher/pull/1333/commits/724f69f18db4507e4b3d2916467362e6a7d21be9

This change will cause the app to error if someone is using it while the app is deployed. Since this app is used infrequently I plan to just carefully deploy this change rather than do defensive migration work.

This PR is intentionally branched off https://github.com/alphagov/collections-publisher/pull/1335 until that PR is merged.

### DIFF
--- a/app/controllers/coronavirus/announcements_controller.rb
+++ b/app/controllers/coronavirus/announcements_controller.rb
@@ -71,7 +71,7 @@ module Coronavirus
     def announcement_params
       params
         .require(:announcement)
-        .permit(:title, :url, published_at: %i[day month year])
+        .permit(:title, :url, published_on: %i[day month year])
     end
 
     def draft_updater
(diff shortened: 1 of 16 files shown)


leenagupte:
This change LGTM 👍. Thanks for explaining why you couldn't just do a rename in the first commit 💯.

I don't think this can be approved until the other change has gone through, so here's symbolic ✅ from me.

kevindew:
@leenagupte this is ready for review now 🎉

PR #1352, opened 2021-04-28
Stop using the github_data method to gather timeline information

JonathanHallam:
The github_data method makes calls to GitHub to gather information held in yaml files. This was done so that non developers could update content. The output of this method was being modified quite opaquely to include data from timelines that is being held in the database rather than the yaml. This changes things so that the `timeline_data` method is called directly which makes it clearer that timeline data is being accessed from the database rather than the yaml.

Trello - https://trello.com/c/rMnlfIq6/335-get-hidden-search-terms-for-timeline-from-database

### DIFF
--- a/app/services/coronavirus/pages/content_builder.rb
+++ b/app/services/coronavirus/pages/content_builder.rb
@@ -91,7 +91,7 @@ module Coronavirus::Pages
     end
 
     def timeline_data
-      page
+      @timeline_data ||= page
         .timeline_entries
         .order(:position)
         .pluck(:heading, :content)
@@ -139,9 +139,9 @@ module Coronavirus::Pages
     end
 
     def search_terms_in_timeline
-      return [] if github_data["timeline"].blank?
+      return [] if timeline_data.blank?
 
-      timeline = github_data["timeline"]["list"].map do |item|
+      timeline = timeline_data.map do |item|
         [
           item["heading"],
           MarkdownService.new.strip_markdown(item["paragraph"]),


app/services/coronavirus/pages/content_builder.rb
```diff
@@ -139,9 +139,9 @@ def search_terms_in_sub_sections(items)
     end
 
     def search_terms_in_timeline
-      return [] if github_data["timeline"].blank?
+      return [] if page.timeline_entries.blank?
```
leenagupte, line 142: I think you can call `timeline_data` directly here.

leenagupte (commented):
One minor comment, otherwise it looks good!

Great commit message too. Thanks! 🎉

app/services/coronavirus/pages/content_builder.rb
JonathanHallam: I did try this and it worked. I decided not to go with it because `timeline_data` has to go through its own data manipulation loop which seemed a bit heavy handed. I opted for `page.timeline_entries` as a proxy because it's used in the `timeline_data` method on line 95 so is maybe a bit closer to what we're actually checking for. It also saves calling `timeline_data` twice. I don't feel too strongly about it though so happy to change this if you still want? :)

leenagupte: Would that be solved by memoizing `timeline_data`?

JonathanHallam: excellent shout, done!

leenagupte: @JonathanHallam Have you been able to test this change on integration?

JonathanHallam: @leenagupte I have now :) pushed to integration, republished the landing page and checked out the content item. All looks as it should

leenagupte (approved):
Awesome! Thanks! 💯

PR #1355, opened 2021-05-04
Remove live stream pages and data

leenagupte:
Trello: https://trello.com/c/qIMREu2x

# What's changed?

Removes the page to add a live stream video to the coronavirus landing page and any existing live stream data.

# Why?

The current live stream of press conferences used to be streamed on the coronavirus landing page. However this stopped in summer 2020 and hasn't been used since). The code is being removed as it's no being used and its presence makes it harder to add features in collections-publisher.


app/views/layouts/admin_layout.html.erb
```diff
@@ -22,7 +22,7 @@
         text: "Specialist sectors",
         href: topics_path,
         active: active_navigation_item == 'topics' ? 'active' : false,
-        user_has_permission_to_see: !livestream_editor?
+        user_has_permission_to_see: true
```
kevindew, line 25: Minor, it looks like the convention that already exists is to remove this argument when there aren't any permissions involved (see https://github.com/alphagov/collections-publisher/blob/c8676a3ef2565d1917cf73418f1863ea3c438be0/app/views/layouts/admin_layout.html.erb#L39-L40)

db/migrate/20210504120417_drop_coronavirus_live_streams.rb
```diff
@@ -0,0 +1,10 @@
+class DropCoronavirusLiveStreams < ActiveRecord::Migration[6.0]
+  def change
+    drop_table :coronavirus_live_streams do |t|
+      t.datetime "created_at", null: false
+      t.datetime "updated_at", null: false
+      t.string "url", null: false
+      t.string "formatted_stream_date"
+    end
```
kevindew, line 8: Nice that this is reversible 👍

kevindew (approved):
Looks great to me! I couldn't find any other rogue traces. Just a minor comment but don't think you need another review.

app/views/layouts/admin_layout.html.erb
leenagupte: Thanks! Will update and fix up.

PR #1399, opened 2021-06-25
Update rails version to 6.1.4

GDSNewt:
In keeping with GOV.UK’s policy of keeping software current we are updating collections publisher to use rails version 6.1.4.

[Trello](https://trello.com/c/xga0rg5V/433-upgrade-collections-publisher-to-rails-61)


Rosa-Fox (commented):
There is another step to do which is the 3rd part of the acceptance criteria; updating [the config](https://github.com/alphagov/collections-publisher/blob/a31929bcf98b267509f3b016073a1f0fe7533b8a/config/application.rb#L25).

This causes some noisy deprecation warnings in the tests so it would be good to fix those. I'll  have a look at them today and can work with you tomorrow morn on them.

app/views/components/markdown_editor/_blockquote.svg
```diff
@@ -1,3 +1,3 @@
-<svg class="app-c-markdown-editor__toolbar-icon" version="1.1" viewBox="0 0 50 50" width="50" height="50" aria-hidden="true">
+<svg version="1.1" viewBox="0 0 50 50" width="50" height="50">
```
kevindew, line 1: We should move these SVG files from views to the app/assets directory, since they aren't actually views and are image files.

So this file's new location should be `app/assets/images/components/markdown_editor/blockquote.svg` (note the removal of the underscore prefix) with similar for these other markdown components.

app/views/shared/steps/_broken-links-summary.html.erb
```diff
@@ -4,12 +4,12 @@
   <div class="step-broken-links__summary">
     <% if step.broken_links? %>
       <span class="app-broken-link-status app-broken-link-status--fail">
-        <%= render "svgs/cross.svg" %>
+        <%= inline_svg_tag "svgs/cross.svg" %>
```
kevindew, line 7: The file referenced here (app/views/svgs/_cross.svg) should move too. We should put that as `app/assets/images/cross.svg` - similar for the tick one below

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
```
kevindew, line 1: We can delete this file as we've set the defaults to be 6.1 in https://github.com/alphagov/collections-publisher/pull/1399/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9R25

This file is useful to help you opt into new defaults gradually before switching the default to be 6.1

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
...
+#
+# This change is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 6.1.
+# Rails.application.config.action_dispatch.cookies_same_site_protection = :lax
+
+# Generate CSRF tokens that are encoded in URL-safe Base64.
+#
+# This change is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 6.1.
+# Rails.application.config.action_controller.urlsafe_csrf_tokens = true
```
kevindew, line 32: This option will risk breaking form submissions for users that are using the app at the time of migration. I think it's ok to do this as this app gets pretty low usage, but we should be careful with the deploy and be aware in case one of the content team reports a problem.

kevindew (commented):
Great stuff Alex, thanks for fixing the deprecation warnings. 

I've got a few suggestions about file moves and then I think this should be done.

Have you deployed it to integration to test it out?

kevindew:
Nice one fixing the error which caused the missing image problem ⭐ 

We hit [the same problem with Content Publisher](https://github.com/alphagov/content-publisher/pull/2287) and added this: https://github.com/alphagov/content-publisher/commit/1b6308a75882007793f0d49430ad49bf13853226 so that the app would error in the circumstances (so it doesn't slip in again). Can we add that into this? once that's in this looks good to merge 👍

GDSNewt:
> Nice one fixing the error which caused the missing image problem ⭐
> 
> We hit [the same problem with Content Publisher](https://github.com/alphagov/content-publisher/pull/2287) and added this: [alphagov/content-publisher@1b6308a](https://github.com/alphagov/content-publisher/commit/1b6308a75882007793f0d49430ad49bf13853226) so that the app would error in the circumstances (so it doesn't slip in again). Can we add that into this? once that's in this looks good to merge 👍

Thanks Kevin, I've added it in. I think we're good to go 🙂

kevindew (approved):
Nice one

PR #1447, opened 2021-10-01
Create migration to add sub heading to coronavirus subsection

JonathanHallam:
This is in order that we can retire the hubs and add the content therein to
the accordions on the landing page.

Trello - https://trello.com/c/0W3i2dN8/645-add-subheading-field-to-subsections-in-coronavirus-publishing-tool

### DIFF
--- a/db/migrate/20211001131132_add_sub_heading_to_coronavirus_sub_section.rb
+++ b/db/migrate/20211001131132_add_sub_heading_to_coronavirus_sub_section.rb
@@ -1,0 +1,5 @@
+class AddSubHeadingToCoronavirusSubSection < ActiveRecord::Migration[6.1]
+  def change
+    add_column :coronavirus_sub_sections, :sub_heading, :string
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20211001131132_add_sub_heading_to_coronavirus_sub_section.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubHeadingToCoronavirusSubSection < ActiveRecord::Migration[6.1]
+  def change
+    add_column :coronavirus_sub_sections, :sub_heading, :string, limit: 255
```
kevindew, line 3: The setting of limit: 255 doesn't actually do anything here as I recall. MySQL just defaults to 255 chars so you can drop that from the arguments.

JonathanHallam: got myself into a right git hole with that one, no idea how! At any rate it's gone now

kevindew (approved):

PR #1448, opened 2021-10-01
Update db schema after rails upgrade

JonathanHallam:
An update to rails means that there's changes to the db schema that
can only be seen when a migration is run. This means that these changes
will be in the schema rather than being drawn in with an upcoming
migration

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2,8 +2,8 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# This file is the source Rails uses to define your schema when running `rails
-# db:schema:load`. When creating a new database, `rails db:schema:load` tends to
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
 # be faster and is potentially less error prone than running all of your
 # migrations from scratch. Old migrations may fail to apply correctly if those
 # migrations use external dependencies or application code.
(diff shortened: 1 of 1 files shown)


db/schema.rb
```diff
@@ -46,10 +46,11 @@
     t.string "action_link_url"
     t.string "action_link_content"
     t.string "action_link_summary"
+    t.string "sub_heading"
```
leenagupte, line 49: I don't think this should be here. I think you need to rollback the last migration first.

kevindew, line 49: Looks like you've got this from your new migration. You'll have to roll that back first (or do db:reset on the main branch if you're not precious about the data)

leenagupte (commented):
One small issue.

db/schema.rb
kevindew: Jinx 😀

db/schema.rb
```diff
@@ -49,18 +49,17 @@
...
 
-  create_table "coronavirus_timeline_entries", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+  create_table "coronavirus_timeline_entries", charset: "utf8", force: :cascade do |t|
     t.bigint "coronavirus_page_id", null: false
     t.text "content", null: false
     t.string "heading", null: false
     t.integer "position", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.string "national_applicability"
```
kevindew, line 59: I don't think you are meaning to lose this - you may have gone back too far?

JonathanHallam: 😆 how is this happening!

leenagupte: Beat me to it!

leenagupte (approved):

db/schema.rb
JonathanHallam: Okay, now looking healthy 😬

PR #1520, opened 2022-02-08
Add table for the nhs_section of covid landing pg

GDSNewt:
We want to make the NHS section publishable from Collections Publisher,
adding this table will lay down the ground work in making that possible.

[Trello](https://trello.com/c/9lhF8mlD/595-publishable-nhs-section-in-collections-publisher-database-migration)

### DIFF
--- a/app/models/nhs_section.rb
+++ b/app/models/nhs_section.rb
@@ -1,0 +1,3 @@
+class NhsSection < ApplicationRecord
+  validates :heading, length: { maximum: 255 }
+end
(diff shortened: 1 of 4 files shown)


leenagupte (commented):
@GDSNewt I've added an inline comment. As mentioned, I think it would help to understand the proposed structure of the NHS section. Will it be one thing with a free text box? Or are you expecting there to be multiple NHS Sections? Once I understand that I'll be able to give you better feedback.

I think it would be useful to add this information to the commit message as well so other developers can understand why these fields are being added in the way they are.

AgaDufrat:
@GDSNewt, can we close this now? 🧹

leenagupte:
@AgaDufrat Yes please! The NHS section has been removed from the landing page, so we're never going to implement this.

PR #1543, opened 2022-03-12
Migrate Mainstream browse pages to Design system

DilwoarH:
## What
Convert the mainstream browser pages views in Collections publisher to the Design system.

## Why
This will address any accessibility concerns and make the whole app consistent with one design pattern.

## Additional Information

We should move things like for like (or as close as possible). Once we have like for like migration, we can iterate on the user journey if needed.

### Index 

|Before|After|
|------------|------------|
|image|image|

### New

|Before|After|
|------------|------------|
|image|image|

### Edit

|Before|After|
|------------|------------|
|image|image|

### Show (parent)

|Before|After|
|------------|------------|
|image|image|

### Show (child)

|Before|After|
|------------|------------|
|image|image|

### Manage ordering  

|Before|After|
|------------|------------|
|image|image|

### Error handling 

|Before|After|
|------------|------------|
|image|image|

## Guidance for review

Probably easiest to run it locally and then compare it to the app on integration.

https://trello.com/c/iRM9tUxa/307-convert-remaining-views-in-collections-publisher-to-design-system

### DIFF
--- a/app/assets/stylesheets/utilities/_overrides.scss
+++ b/app/assets/stylesheets/utilities/_overrides.scss
@@ -111,3 +111,7 @@
   margin-top: 0;
   background-color: govuk-colour("light-grey");
 }
+
+.govuk-tag--small {
+  @include govuk-font(14, $weight: bold, $line-height: 1);
+}
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -72,6 +72,17 @@ class MainstreamBrowsePagesController < ApplicationController
     @browse_page = find_browse_page
   end
 
+  helper_method :issues_for
+  def issues_for(object, attribute)
+    object.errors.errors.filter_map do |error|
+      if error.attribute == attribute
+        {
+          text: error.options[:message],
+        }
+      end
+    end
+  end
+
 private
 
   def topics_for_select
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -1,5 +1,4 @@
 class MainstreamBrowsePagesController < ApplicationController
-  layout "legacy"
   before_action :require_gds_editor_permissions!
   before_action :protect_archived_browse_pages!, only: %i[edit update publish]
 
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -129,7 +129,21 @@ private
 
   def manage_child_ordering_params
     params.require(:mainstream_browse_page)
-      .permit(:child_ordering, children_attributes: %i[index id])
+      .permit(:child_ordering)
+      .merge(children_params)
+  end
+
+  def children_params
+    return unless params.dig("mainstream_browse_page", "child_ordering") == "curated"
+
+    children_attributes_hash = {}
+
+    params[:ordering].each do |input|
+      child_id, order = input
+      children_attributes_hash[order] = { "index" => order, "id" => child_id }
+    end
+
+    { children_attributes: children_attributes_hash }
   end
 
   def protect_archived_browse_pages!
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -22,6 +22,9 @@ class MainstreamBrowsePagesController < ApplicationController
     if browse_page.update(browse_page_params)
       TagBroadcaster.broadcast(browse_page)
       redirect_to browse_page
+    elsif browse_page_params.keys.include?("child_ordering")
+      @browse_page = browse_page
+      render :manage_child_ordering
     else
       @browse_page = browse_page
       @topics_for_select = topics_for_select
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -22,9 +22,6 @@ class MainstreamBrowsePagesController < ApplicationController
     if browse_page.update(browse_page_params)
       TagBroadcaster.broadcast(browse_page)
       redirect_to browse_page
-    elsif browse_page_params.keys.include?("child_ordering")
-      @browse_page = browse_page
-      render :manage_child_ordering
     else
       @browse_page = browse_page
       @topics_for_select = topics_for_select
@@ -74,6 +71,17 @@ class MainstreamBrowsePagesController < ApplicationController
     @browse_page = find_browse_page
   end
 
+  def update_child_ordering
+    @browse_page = find_browse_page
+
+    if @browse_page.update(manage_child_ordering_params)
+      TagBroadcaster.broadcast(@browse_page)
+      redirect_to @browse_page
+    else
+      render :manage_child_ordering
+    end
+  end
+
   helper_method :issues_for
   def issues_for(object, attribute)
     object.errors.errors.filter_map do |error|
@@ -116,7 +124,12 @@ private
 
   def tag_params
     params.require(:mainstream_browse_page)
-      .permit(:slug, :title, :description, :parent_id, :child_ordering, children_attributes: %i[index id])
+      .permit(:slug, :title, :description, :parent_id)
+  end
+
+  def manage_child_ordering_params
+    params.require(:mainstream_browse_page)
+      .permit(:child_ordering, children_attributes: %i[index id])
   end
 
   def protect_archived_browse_pages!
--- a/app/controllers/mainstream_browse_pages_controller.rb
+++ b/app/controllers/mainstream_browse_pages_controller.rb
@@ -122,7 +122,7 @@ private
   def protect_archived_browse_pages!
     browse_page = find_browse_page
     if browse_page.archived?
-      flash[:danger] = "You cannot modify an archived mainstream browse page."
+      flash[:alert] = "You cannot modify an archived mainstream browse page."
       redirect_to browse_page
     end
   end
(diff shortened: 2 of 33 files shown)


app/views/mainstream_browse_pages/_list-browse-pages.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    if status === "draft"
+      classes.append("govuk-tag--grey")
+    elsif status === "published"
+      classes.append("govuk-tag--green")
+    elsif
+      classes.append("govuk-tag--blue")
+    end
+
+    tag.strong status, class: classes.join(' ')
+  end
```
DilwoarH, line 13: Need to move this into a presenter

DilwoarH: removed

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -111,3 +111,13 @@
...
+
+.govuk-tag-s .govuk-tag {
+  font-size: 70%;
+  vertical-align: middle;
+}
+
+h1 .govuk-tag-l {
+  font-size: 60%;
+  vertical-align: middle;
+}
```
DilwoarH, line 123: instead of this, we can just add font classes to the elements.

https://design-system.service.gov.uk/styles/typography/#font-override-classes

spec/features/navigating_browse_pages_spec.rb
```diff
@@ -20,15 +20,15 @@
   end
 
   def given_there_are_browse_pages
-    create(:mainstream_browse_page, :published, title: "Money and Tax")
-    citizenship = create(:mainstream_browse_page, :published, title: "Citizenship")
-    create(:mainstream_browse_page, parent: citizenship, title: "Voting")
-    british_citizenship = create(:mainstream_browse_page, parent: citizenship, title: "British citizenship")
+    @money_and_tax = create(:mainstream_browse_page, :published, title: "Money and Tax")
```
DilwoarH, line 23: is there a reason why these needed to be assigned?

app/assets/stylesheets/utilities/_overrides.scss
davidgisbey: Absolutely. I completely forgot that was a thing. Thanks!

spec/features/navigating_browse_pages_spec.rb
davidgisbey: It's used on line 44 and 45 to check it's rendered on the show page

app/assets/stylesheets/utilities/_overrides.scss
DilwoarH: I think make all tags small - the big tags look a bit weird next to the small

app/assets/stylesheets/utilities/_overrides.scss
```diff
@@ -111,3 +111,12 @@
...
 }
+
+.govuk-tag--small {
+  @include govuk-font(14, $weight: bold, $line-height: 1);
+}
+
+h1 .govuk-tag-l {
+  font-size: 60%;
+  vertical-align: middle;
+}
```
DilwoarH, line 122: This needs removing

brucebolt (commented):
Great work!

I've left a few comments that are mostly stylistic nitpicks.

app/helpers/header_helper.rb
```diff
@@ -38,4 +38,30 @@ def auto_link(object)
...
+
+  def govuk_heading_with_parent(tag)
+    tag.title_including_parent
+  end
+
+  def govuk_tag_heading(tag:, prepend: nil, append: nil)
+    context = govuk_status(tag)
+    title = tag.title_including_parent
+    title = "#{prepend} #{title}" if prepend
+    title = "#{title}: #{append} " if append
```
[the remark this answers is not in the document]
davidgisbey: Yea i don't know why, but i thought there was one title that needed a prepend and append, but it doesn't look like that's the case. Will change 👍

app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -64,14 +66,26 @@ def archive
...
+  helper_method :issues_for
+  def issues_for(object, attribute)
+    errors_for_input = object.errors.errors.select do |error|
+      error.attribute == attribute
+    end
+    errors_for_input.map do |error|
+      {
+        text: error.options[:message],
+      }
+    end
```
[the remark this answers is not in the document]
davidgisbey: Til. That's so cool. Thanks Bruce!

davidgisbey:
Have addressed all the comments. Thanks for the review @brucebolt

brucebolt (approved):
Thanks for making those changes.  Everything looks good to me 👍.

app/views/mainstream_browse_pages/edit.html.erb
```diff
@@ -1,18 +1,29 @@
...
+      text: 'Edit',
+    }
+  ]
+} %>
 
-<%= form_for @browse_page do |f| %>
-  <%= f.text_field :slug, :disabled => true %>
-  <%= f.text_field :title %>
-  <%= f.text_field :description %>
+<% content_for :title, govuk_tag_heading(tag: @browse_page, prepend: 'Edit') %>
```
ollietreend, line 21: Is this header supposed to include the tag?

Currently it causes a visual jump when moving from `show.html.erb` to `edit.html.erb`.

| Show page | Edit page |
| --- | --- |
| show | edit |

ollietreend (commented):
This is looking beautiful 😍 Nice work!

I've not given this a thorough look through yet – but just adding one initial comment.

app/views/mainstream_browse_pages/_create_update_form.html.erb
```diff
@@ -0,0 +1,60 @@
+<%= form_for @browse_page do |f| %>
+  <% unless update %>
```
ollietreend, line 2: I always find `unless` conditional logic a bit of a mind bender. I wonder if this would benefit from a comment just to explain it in terms that my brain understands.

```suggestion
  <%# Only show parent selection when creating a new page %>
  <% unless update %>
```

app/models/tag.rb
```diff
@@ -73,7 +75,7 @@ def sorted_children
     if child_ordering == "alphabetical"
       children.sort_by { |child| child.title.downcase }
     else
-      children.sort_by(&:index)
+      children.map(&:index).any?(&:nil?) ? children : children.sort_by(&:index)
```
ollietreend, line 78: Could this be refactored into the main `if` statement?

I can see what this line is doing, but am wondering whether it'd be more readable as a three-part `if` statement rather than nesting it.

e.g.

```ruby
if child_ordering == "alphabetical"
  children.sort_by { |child| child.title.downcase }
elsif children.map(&:index).all?(&:present?)
  children.sort_by(&:index)
else
  children
end
```

app/views/mainstream_browse_pages/_redirects.html.erb
```diff
@@ -1,16 +1,29 @@
...
-  </tr>
-  <% redirect_routes.order('created_at DESC').each do |redirect_route| %>
-    <tr>
-      <td><%= redirect_route.created_at.to_formatted_s(:long) %></td>
-      <td><%= link_to redirect_route.from_base_path, website_url(redirect_route.from_base_path), target: '_blank' %></td>
-      <td><%= link_to redirect_route.to_base_path, website_url(redirect_route.to_base_path), target: '_blank' %></td>
-    </tr>
-  <% end %>
-</table>
+<%= render "govuk_publishing_components/components/table", {
```
ollietreend, line 1: Not entirely PR related, but:

What's the difference between using the Table Component vs. the Table Helper?
When/why would we want to use one over the other?

I can see we've used both in this PR, so I'm guessing there are certain situations that suit each best.

ollietreend (commented):
Really nice work ✨ 

I've just added a few more comments

app/views/mainstream_browse_pages/edit.html.erb
DilwoarH: This is how it's currently styled, we decided to stick to the current styling until we have a designer go through each bit and revise it. (I agree with what you're saying, not very pleasing for users)

app/views/mainstream_browse_pages/_redirects.html.erb
DilwoarH: So the table helper is what the table component is built with. We are using this as it allows us to use a specific style of the table (which is standard across the publishing apps). We will probably revise this in a future iteration and just stick to using GOVUK tables.

app/controllers/mainstream_browse_pages_controller.rb
```diff
@@ -64,14 +63,36 @@ def archive
...
+    if @browse_page.update(manage_child_ordering_params)
+      TagBroadcaster.broadcast(@browse_page)
+      redirect_to @browse_page
+    else
+      render :manage_child_ordering
+    end
+  end
+
+  helper_method :issues_for
+  def issues_for(object, attribute)
```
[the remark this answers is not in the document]
davidgisbey: This is the naming used in content publisher and the step by step so i think it's probs better to be consistent. wdyt?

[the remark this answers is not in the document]
ollietreend: I don't have a strong feeling either way, but my instinct is to stick to Rails conventions where possible – even if it means method names differ from labels used in the UI (e.g. users see "issues" in the view, but the variables and methods follow the "errors" naming convention). That way we follow the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for when other Rails developers pick up this application.

I was just about to say: it's no biggie 🙂 it's probably not worth the effort to change it at this point. But then I just spotted 4ea24e33a488358dd1f6a08b9732358ac0c6a42c where you did rename it! 🤓

PR #1547, opened 2022-03-21
[DO NOT MERGE] Add local jQuery

andysellick:

## What
Adds jQuery into this application, using the package.json to include the version currently in the components gem.

**This PR should be merged immediately prior to the removal of jQuery from the rest of GOV.UK, as the change will result in an increased asset size for this application**. Having two copies of jQuery running together appears to not cause any problems, but we want to avoid a situation where collections-publisher has no jQuery supplied.

Also fixes a problem where the autocomplete component wasn't appearing correctly in the component guide for this application.

## Why
We're removing jQuery from the components gem. This application has a lot of jQuery and isn't public facing, so to prevent  blocking it's easiest right now to give collections-publisher its own jQuery.

## Visual changes
None

Trello card: https://trello.com/c/mAfQNVRG/745-investigate-which-apps-we-could-remove-jquery-from?filter=member:andysellick2

### DIFF
--- a/app/assets/javascripts/admin_layout.js
+++ b/app/assets/javascripts/admin_layout.js
@@ -1,3 +1,4 @@
+//= require jquery/dist/jquery
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
 
--- a/app/assets/javascripts/admin_layout.js
+++ b/app/assets/javascripts/admin_layout.js
@@ -1,6 +1,7 @@
 //= require jquery/dist/jquery
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
+//= require components/autocomplete
 
 //= require jquery-ui.sortable.min
 //= require jquery_ujs
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -3,11 +3,11 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function Autocomplete () { }
+  function Autocomplete ($module) {
+    this.$module = $module
+  }
 
-  Autocomplete.prototype.start = function ($module) {
-    this.$module = $module[0]
-
+  Autocomplete.prototype.init = function () {
     var $input = this.$module.querySelector('input')
 
     if ($input) {
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,3 +1,4 @@
+//= require jquery/dist/jquery
 //= require components/autocomplete.js
 //= require jquery-ui.sortable.min
 //= require jquery.clicktoggle
--- a/app/assets/stylesheets/components/_all.scss
+++ b/app/assets/stylesheets/components/_all.scss
@@ -1,2 +1,3 @@
+@import "./autocomplete";
 @import "./markdown-editor";
 @import "./timeline-entry";
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "accessible-autocomplete": "^2.0.4",
     "core-js-bundle": "^3.0.0-alpha.1",
     "es5-polyfill": "^0.0.6",
-    "markdown-toolbar-element": "^0.2.0"
+    "markdown-toolbar-element": "^0.2.0",
+    "jquery": "1.12.4"
   },
   "devDependencies": {
     "jasmine-browser-runner": "^1.0.0",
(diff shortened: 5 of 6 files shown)


leenagupte (commented):
It looks good to me, though I'm not sure how to test that the auto-complete still works. Were you able to test this locally.

andysellick:
> It looks good to me, though I'm not sure how to test that the auto-complete still works. Were you able to test this locally?

I was. Easy way to do it is to copy the example code from the component guide into any template that matches a page that can be rendered locally.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -3,11 +3,11 @@ window.GOVUK = window.GOVUK || {}
...
 (function (Modules) {
-  function Autocomplete () { }
-
-  Autocomplete.prototype.start = function ($module) {
-    this.$module = $module[0]
+  function Autocomplete ($module) {
+    this.$module = $module
+  }
 
+  Autocomplete.prototype.init = function () {
```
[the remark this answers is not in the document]
andysellick: curated_lists (and the other similar JS in this repo) has an `init` function, but it isn't one of our modules, so it isn't initialised by the central modules startup code in the components gem and doesn't require any modification. Here's curated lists:

```
(function () {
  'use strict'
  window.GOVUK = window.GOVUK || {}
  var $ = window.jQuery
  var csrfToken = $('meta[name="csrf-token"]').attr('content')

  GOVUK.curatedLists = {
    init: function () {
      // etc.
```

and here's a typical module:

```
window.GOVUK = window.GOVUK || {}
window.GOVUK.Modules = window.GOVUK.Modules || {};

(function (Modules) {
  function Autocomplete ($module) {
    this.$module = $module
  }

  Autocomplete.prototype.init = function () {
   // etc.
```

Note the lack of constructor in the first example (`function NameOfModule($module) {}`) and the lack of any mention of `window.GOVUK.Modules`.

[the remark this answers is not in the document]
leenagupte: Thank you!

kevindew:
I don't think you need to bundle a version of jQuery for this app.

It's a bit of a confusing app because it's half govuk_admin_template, half govuk_publishing_components. The govuk_admin_template side should automatically include the jQuery from govuk_admin_template, whereas - hopefully - the govuk_publishing_components side doesn't need jQuery. 

I'm just giving this a spin locally to work out what's going on because there seems to be imports in the different files that I think shouldn't be there, but I could be wrong and there is some lingering dependencies scattered between the two.

leenagupte:
> > It looks good to me, though I'm not sure how to test that the auto-complete still works. Were you able to test this locally?
> 
> I was. Easy way to do it is to copy the example code from the component guide into any template that matches a page that can be rendered locally.

That's a great idea. It works like a charm! 🎉

andysellick:
> I don't think you need to bundle a version of jQuery for this app.

@kevindew just checked this locally with the branch of the [components gem without jQuery](https://github.com/alphagov/govuk_publishing_components/pull/2613) and got jQuery related errors on the first page. Checking the page source doesn't show any inclusion of a local jQuery (the errors are caused by jQuery-UI trying to initialise without jQuery being included).

Screenshot 2022-03-23 at 11 23 45

DilwoarH:
> > I don't think you need to bundle a version of jQuery for this app.
> 
> @kevindew just checked this locally with the branch of the [components gem without jQuery](https://github.com/alphagov/govuk_publishing_components/pull/2613) and got jQuery related errors on the first page. Checking the page source doesn't show any inclusion of a local jQuery (the errors are caused by jQuery-UI trying to initialise without jQuery being included).
> 
> Screenshot 2022-03-23 at 11 23 45

I think this is because of this: https://github.com/alphagov/collections-publisher/blob/main/app/assets/javascripts/admin_layout.js#L4

I believe we initially needed this for the reorderable component which has since moved to the publishing components gem and doesn't need jQuery UI anymore.

app/assets/javascripts/admin_layout.js
```diff
@@ -1,5 +1,7 @@
+//= require jquery/dist/jquery
```
DilwoarH, line 1: I don't think we need jQuery (as far as I am aware)

app/assets/javascripts/admin_layout.js
```diff
@@ -1,5 +1,7 @@
...
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
+//= require components/autocomplete
 
 //= require jquery-ui.sortable.min
 //= require jquery_ujs
```
DilwoarH, line 7:
```suggestion

```

andysellick: Hmm just tried that and I'm still getting errors from missing jQuery.

andysellick:
I'm going to close this PR in favour of https://github.com/alphagov/collections-publisher/pull/1550, thanks all for the reviews and comments.

PR #1550, opened 2022-03-23
Disentangle layouts for govuk_publishing_components jQuery removal 

kevindew:

This builds on https://github.com/alphagov/collections-publisher/pull/1547 but avoids adding jQuery as a dependency. To reach that point I had to do some surgery of the layouts.

It turned out that we were finding that we needed jQuery because there was coupling of the CSS and JS files in the two distinct layouts this application uses (design system based one and the legacy govuk_admin_template one). This was happening because the design system layout will automatically include files named application.js and application.css

To resolve this. I've:

- renamed the layout files to be clearer in purpose (design_system.html and legacy.html) and updated the layout calls so we default to the design_system and need to specify the legacy one
- I've renamed the JS and SCSS root files to reflect the naming requirements of the layout_for_admin template. We now have application and application_legacy JS/SCSS root files.
- I've unpicked cases where components are in the legacy js file and unused jQuery dependencies are in the design system JS
- I've switched from using jquery_ujs to rails-ujs to remove the jQuery expectation.
- It removes the JS testing set-up as there isn't actually any tests and it depends on a jQuery installation.

I've been testing this with [govuk_publishing_components on the jQuery removed branch](https://github.com/alphagov/govuk_publishing_components/pull/2613) and not found any issues or any tests that fail - I would be grateful for a second opinion on this though.

This change will also work before the govuk_publishing_components change is merged so it doesn't need to wait for a release and can be merged whenever we're happy.


andysellick (commented):
I've tested this locally with a non-jQuery version of the components gem and can't see any errors, so the jQuery side of things looks like it's working perfectly 🌟 

Happy to approve but I'm less familiar with some of the innards of this app, maybe @DilwoarH could take a look?

DilwoarH (approved):
Looks good to me 👍 

A bit concerning that we don't have JS tests...

kevindew:
> Looks good to me 👍
> 
> A bit concerning that we don't have JS tests...

Yup, though I'd prefer having none than a fake one.

Thanks for the review. Hope it doesn't cause too much pain for the design system branch.

PR #1566, opened 2022-04-05
Remove announcements

gclssvglx:
## What

Remove announcements

## Why

In [1st April - Remove NHS Box, Announcements, Timeline Entries, Statistics sections from landing page and move DA Links](https://trello.com/c/zxAAM52Y) we stopped rendering the announcements section on the landing page. Now we need to remove the publishing code.

The risk level content is being hidden, but it used to be rendered as part of the announcements section and can also be removed.

[Trello](https://trello.com/c/JyDtH6Dr/815-remove-announcements-section-from-collections-publisher-and-yaml-file)


db/schema.rb
```diff
@@ -10,19 +10,9 @@
...
-    t.string "title", null: false
-    t.text "url", null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.integer "position", null: false
-    t.date "published_on"
-  end
-
-  create_table "coronavirus_pages", charset: "utf8", force: :cascade do |t|
+  create_table "coronavirus_pages", charset: "utf8mb3", force: :cascade do |t|
```
Rosa-Fox, line 15: I wondered why the charset changed, but it seems like utf8 is an alias of utf8mb3 so should be fine.

Rosa-Fox (approved):
I had a good look through and couldn't find anything announcements related left 🥇 . LGTM!

db/schema.rb
gclssvglx: Yeah, this was automatic when I did `rails db:migrate` locally - news to be but seems to be a [MySQL specific](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html) charset.

PR #1574, opened 2022-04-11
Remove GitHub yaml dependency

leenagupte:

Trello: https://trello.com/c/80EkwoDu
Depends on: [Content being harded-coded in collections](https://github.com/alphagov/collections/pull/2748)

# What?
Removes the functionality to pull in and publish content from [govuk-coronavirus-content](https://github.com/alphagov/govuk-coronavirus-content)

# Why?

The content on the [coronavirus landing page](https://www.gov.uk/coronavirus) has been significantly reduced, with most of the content that was being pulled in from the YAML file, like the statistics section having been deleted.

Therefore, rather than continue to allow the content on the coronavirus landing page to edited in two places, the govuk-coronavirus-content Github repo will be removed as a dependency, and the remaining content will be hard-coded into [collections](https://github.com/alphagov/collections). This will give us one less codebase to maintain. Pulling in GitHub in this way also somewhat complicated the code in Collections Publisher. Removing this functionality now will make the publishing off the landing page easier to maintain for as long as it remains.

# Expected changes

|Before|After|
|:------|:----|
|Screenshot 2022-04-12 at 10 47 14|Screenshot 2022-04-12 at 10 27 05|

# Note to reviewer
It's not possible to test the previewing of drafts until the frontend has been updated and the required [topic section](https://github.com/alphagov/govuk-coronavirus-content/blob/master/content/coronavirus_landing_page.yml#L19) has been moved into collections.


app/services/coronavirus/pages/draft_updater.rb
```diff
@@ -16,10 +16,6 @@ def content_builder
 
     def payload
       Coronavirus::PagePresenter.new(content_builder.data, base_path).payload
-    rescue ContentBuilder::GitHubInvalidContentError
-      raise DraftUpdaterError, "Invalid content in GitHub YAML"
-    rescue ContentBuilder::GitHubConnectionError
-      raise DraftUpdaterError, "Unable to load content from GitHub"
```
gclssvglx, line 22: Nice that these can go 👍

app/services/coronavirus/pages/content_builder.rb
```diff
@@ -12,8 +12,8 @@ def initialize(page)
 
     def data
       @data ||= begin
-        validate_content
-        data = github_data
+        data = {}
+        data["title"] = "Coronavirus (COVID-19): guidance and support"
```
gclssvglx, line 12: Nice - introducing a hash with the same name - very neat 🥇

spec/services/coronavirus/pages/content_builder_spec.rb
```diff
@@ -61,13 +28,15 @@
...
+      @data["title"] = "Coronavirus (COVID-19): guidance and support"
+      @data["header_section"] = header_json
+      @data["sections"] = [sub_section_json]
+      @data["hidden_search_terms"] = hidden_search_terms
     end
 
-    it "returns github and model data" do
-      expect(subject.data).to eq data
+    it "returns model data" do
+      expect(subject.data).to eq @data
```
gclssvglx, line 39: Would be really nice to keep `data` instead of creating `@data`. Maybe move the `before` block into the `it` test, then `data` becomes possible, and since it's only use for this test...?

spec/support/coronavirus_feature_steps.rb
```diff
@@ -68,8 +68,8 @@ def then_the_content_is_sent_to_publishing_api
     )
   end
 
-  def i_see_a_publish_landing_page_link
-    expect(page).to have_link(I18n.t("coronavirus.pages.index.landing_page_edit.something_else"))
+  def then_i_see_an_edit_landing_page_link
+    expect(page).to have_link(I18n.t("coronavirus.pages.index.landing_page_edit.sections"))
```
gclssvglx, line 51: I struggled with this for a bit - thinking why not just remove the test, but I can see why you've updated it to test the other link instead. Thanks for keeping a test on this page - totally the right thing to do 💯

gclssvglx (approved):
This looks really great - so much code resigned to the bin 😄  Just one really nit-picky suggestion, but feel free to ignore it. Thanks @leenagupte - great job 🥇

spec/services/coronavirus/pages/content_builder_spec.rb
leenagupte: I like that suggestion. I've updated and fixed up.

Rosa-Fox (commented):
@leenagupte looks great! I found a tiny thing which is that in factories.rb the 

```
    github_url { Faker::Internet.url(host: "example.com") }
    raw_content_url { Faker::Internet.url(host: "example.com") }
```

lines can be removed from:

```
  factory :coronavirus_page, class: Coronavirus::Page do
    name { Faker::Company.industry }
    slug { "landing" }
    content_id { SecureRandom.uuid }
    sections_title { Faker::Lorem.sentence }
    github_url { Faker::Internet.url(host: "example.com") }
    raw_content_url { Faker::Internet.url(host: "example.com") }
    base_path { "/#{File.join(Faker::Lorem.words)}" }
  end
```

ooh and they can be removed from the coronavirus_pages table too!

leenagupte:
@Rosa-Fox Thanks for your review. I've removed the database columns as suggested. Good spot! Thank you to @gclssvglx to running the migration for me 😩.

Rosa-Fox (approved):

PR #1582, opened 2022-04-22
Use consistent language to describe tagging and topic pages

AgaDufrat:
Clear consistent language in the publishing apps will help publishers tag
content better, and understand the implications of this. Better tagging will
make their content more findable.


app/models/tag.rb
```diff
@@ -6,7 +6,7 @@ class Tag < ApplicationRecord
   ORDERING_TYPES = %w[alphabetical curated].freeze
 
   belongs_to :parent, class_name: "Tag"
-  has_many :children, class_name: "Tag", foreign_key: :parent_id
+  has_many :children, class_name: "Tag", foreign_key: :parent_id # called Subtopics in the frontend
```
hannako, line 9: 👍

app/views/topics/show.html.erb
```diff
@@ -85,7 +85,7 @@
           <% if @topic.child? %>
             <div class="govuk-!-margin-bottom-1">
               <% if @topic.published? %>
-                <%= link_to 'Archive mainstream browse page',
+                <%= link_to 'Archive specialist topic page',
```
hannako, line 88: This has confused me! Was it incorrect on the main branch, ie was it meant to say **Archive specialist sector page** before? ie there was a bug.

hannako (commented):
I can't quite match up the "Current text" column in Jen's doc (linked in the PR description) to the before state in this PR.  I'm guessing because some things have already been updated by the switch to the design system eg breadcrumbs on the specialist sector pages already say `topic` instead of `specialist sector page`.

 But if Jen has reviewed and is happy, the code changes look fine.

hannako (approved):

app/views/topics/show.html.erb
AgaDufrat: I think it sneaked in during the swap to GOV.UK Design System

PR #1584, opened 2022-04-25
Fix flash messages

AgaDufrat:
The new layout using GOV.UK Design System renders flash notices only if the
keys are "alert" or "notice". Any messages stored under other keys such as
"success" or "danger" were not displayed.

https://trello.com/b/iACGoJN3/govuk-explore-doing

### DIFF
--- a/spec/features/managing_browse_pages_spec.rb
+++ b/spec/features/managing_browse_pages_spec.rb
@@ -42,6 +42,13 @@ RSpec.feature "Managing browse pages" do
     then_i_see_a_validation_error
   end
 
+  scenario "Unique slug validation unique slug fails" do
+    given_there_is_published_browse_page
+    when_i_visit_the_new_browse_page_form
+    and_i_fill_in_the_form_with_already_existing_slug
+    then_i_see_uniqueness_validation_error
+  end
+
   def given_there_is_published_browse_page
     @page = create(:mainstream_browse_page, :published, slug: "citizenship", title: "Citizenship")
   end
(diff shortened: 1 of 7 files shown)


hannako (changes requested):
Great work spotting and implementing a fix for this! 

It might be nice to add links in your commit message to [here](https://github.com/alphagov/collections-publisher/blob/60771a63f39e632c43e2baa5d1ef4abd65d1a926/app/views/layouts/design_system.html.erb#L48) and [here](https://github.com/alphagov/collections-publisher/blob/60771a63f39e632c43e2baa5d1ef4abd65d1a926/app/views/layouts/design_system.html.erb#L54) to show why you've had to make the changes to the key names. 

It also looks like the error_summary partial [needs the error to have](https://github.com/alphagov/collections-publisher/blob/60771a63f39e632c43e2baa5d1ef4abd65d1a926/app/views/shared/_error_summary.html.erb#L8) a `message` attribute. If you try to add a topic with a slug that's already taken it just says `taken` which is the error type. ie add in a message [here](https://github.com/alphagov/collections-publisher/blob/60771a63f39e632c43e2baa5d1ef4abd65d1a926/app/models/tag.rb#L20)

I haven't checked but wonder if there are other validations that are missing and error message attribute too.

hannako:
P.S. you could maybe make a card for this so we track the work as it's not really related to the label change stuff.

Tetrino (commented):
👍

Tetrino:
@AgaDufrat I only just caught your comment that you've got more coming, let me know when you're ready :)

AgaDufrat:
I've done that now (see screenshots in https://github.com/alphagov/collections-publisher/pull/1584#issuecomment-1110728312). Ready for review :)

spec/features/managing_browse_pages_spec.rb
```diff
@@ -90,6 +97,8 @@ def and_i_fill_in_the_form
     @content_id = extract_content_id_from(current_path)
   end
 
+  alias_method :and_i_fill_in_the_form_with_already_existing_slug, :and_i_fill_in_the_form
```
hannako, line 100: this is cool!

hannako (approved):
Nice!

PR #1585, opened 2022-04-25
Fix url validation regex for unpublishing step by steps

theseanything:
This update the regex for unpublishing step by step to be more permissive. Previously url paths with multiple parts i.e. extra "/"s would be considered invalid. This also tidies up the code to return true or false, rather than the match number or nil.


MartinJJones (approved):
Looks good to me 👍

sihugh (approved):

PR #1594, opened 2022-05-05
Remove Beta Banner from collections publisher

DilwoarH:
## What
Remove beta banner in Collections Publisher

## Why
This is incorrect as the product is not beta and is not being rebuilt - this may be considered misleading to the user.

https://trello.com/c/aFpyXJdm/415-remove-beta-banner-in-collections-publisher

## Before

Screenshot 2022-05-05 at 2 14 34 am

## After

Screenshot 2022-05-05 at 9 37 08 am

### DIFF
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -36,11 +36,6 @@
     ].delete_if{ |item| item[:user_has_permission_to_see] == false },
   }%>
   <div class="govuk-width-container">
-    <%= render "govuk_publishing_components/components/phase_banner", {
-      app_name: "Collections Publisher",
-      phase: "beta"
-    } %>
-
     <%= yield(:back_link) %>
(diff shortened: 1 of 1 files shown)


ollietreend (commented):
It looks like this change means that the UI no longer shows 'Collections Publisher' anywhere obvious for the user to know which app they're in – since the beta banner is what showed the app name in a prominent position.

Is this okay? Do users benefit from knowing which Publishing app they're using, or is that not useful information?

DilwoarH:
> It looks like this change means that the UI no longer shows 'Collections Publisher' anywhere obvious for the user to know which app they're in – since the beta banner is what showed the app name in a prominent position.
> 
> Is this okay? Do users benefit from knowing which Publishing app they're using, or is that not useful information?

Good spot - I've added the product name in

davidgisbey (approved):

PR #1595, opened 2022-05-06
Implement new flow for adding, editing, reordering and deleting lists 

davidgisbey:
## Description 

The majority of Collections Publisher has been ported over to use the GOV.UK Design System with the notable exception of managing lists for related links for Mainstream Browse Pages and Topics.

Some work has been done by Nikin to redesign this page https://collections-publisher.integration.publishing.service.gov.uk/tags/f881f972-6094-4c7d-849c-9143461a9307/lists so that is more accessible for users as the current functionality does not fully work without JS and cannot be viably used via screen reader or keyboard.

This is the first of two PRs adding the new flows for the lists page on collections publisher. This PR handles all the available functionality for lists, but does not implement functionality for adding items to lists. 

All of the functionality is hidden behind the `Redesigned lists` permission. If a user does not have this permission, all functionality should continue to behave as normal. 

Where possible i've moved code for the existing flow into separate files, rather than using conditionals (views etc.). The notable exception for this is the controller where i've moved existing behaviour into else clauses of each action. Hopefully this should make pulling the code out easy.

## Changes made

The functionality that is added in this PR is the following:

### Updated the child show page to use the new design 

- Implements the new design for the tag/mainstream browse child show page. All of the changes are `curated_links_preview` partial. This is rendered in the mainstream browse and topic show pages.

|Before|After|
|-----------|-----------|
|image|image|

### Added create list functionality

- Wires up the `Add list` link to the lists#new action
- Adds the new endpoint

image

### Added edit list name functionality

- Wires up the `Rename list` link to the lists#edit action
- Moves the existing `edit` view into a legacy file 
- adds the new `edit` view and handles dealing with the two responses in the controller action

image

### Added reorder lists functionality

- Adds the reorder endpoint

image

### Added delete list functionality 

- Adds a confirm destroy endpoint
- Enforces the correct redirection behaviour in the `destroy` controller action 

image

## Trello card 

### Designs 

https://www.figma.com/file/SpMRDQQarfICRCSyhVVmVq/Collections-Publisher-Curate-List?node-id=0%3A1

## Guidance for review 

I'd highly recommend reviewing this per commit as it's broken down into sets of two commits. One which adds a feature spec and the next which implements the code which gets the test passing.

Lists belong to Mainstream Browse Page & Topic tags. This means that a lot of the redirection relies on polymorphism for redirection after a successful action. It's worth checking that the functionality works correctly for both types of tag.

If you want to deploy the branch out to integration and test it you'll have to give yourself the 'Redesigned lists' permission on Signon.

This looks bigger than it is due to approximately 200 lines of code being moved to legacy files 👍 

### Question: Do we need controller specs or fleshed our feature specs?

For the current implementation there are no controller specs. I know that post Rails 5, controller specs have been `discouraged` (see discussion here https://github.com/rspec/rspec-rails/issues/1838).

Previously i've seen a combination of unit tests being used via view components and slightly more involved feature specs which test things like the flash.

However, a lot of our apps continue to use controller specs. Collections has some controller spec coverage.  I know this is a tricky as it's not a one solution works for all our applications type question.

My gut says to use controller specs as it's a well established pattern in a lot of our apps and as far as i know there's no plans to further develop Collections Publisher in the near future.

I think one of the two approaches above should probably be taken as this app is CD and not implementing a solution would cause a decent drop to test coverage percentage in the short term before the legacy code is removed.


app/controllers/application_controller.rb
```diff
@@ -27,6 +27,10 @@ def coronavirus_editor?
     current_user.has_permission? "Coronavirus editor"
   end
 
+  def redesinged_lists_permission?
```
ollietreend, line 30: Tiny nit pick: there's a typo in this method name

```suggestion
  def redesigned_lists_permission?
```

spec/support/common_feature_steps.rb
```diff
@@ -9,6 +9,10 @@ def given_i_am_a_gds_editor
     stub_user.name = "Test author"
   end
 
+  def and_i_have_the_redesinged_lists_permission
```
ollietreend, line 12: Ditto here

```suggestion
  def and_i_have_the_redesigned_lists_permission
```

ollietreend (commented):
This is only a partial review – but I thought I'd add these comments early while I keep going through it.

app/controllers/lists_controller.rb
```diff
@@ -74,23 +74,22 @@ def update
...
 
-    @tag.mark_as_dirty! if saved
-
-    respond_to do |format|
-      format.html do
-        if saved
-          flash[:success] = "List updated"
-        else
-          flash[:danger] = "Could not save your list"
+      respond_to do |format|
```
davidgisbey, line 93: Just realised this commit should definitely be squashed into the previous one. I'll sort it once you've finished your review so it doesn't update the commit SHA's

app/controllers/application_controller.rb
davidgisbey: Good spot! Have fixed it per commit. Thanks 👍

app/controllers/tags_controller.rb
```diff
@@ -1,10 +1,42 @@
...
+
+  def update_list_ordering
+    if save_ordering
+      @tag.mark_as_dirty!
+      ListPublisher.new(@tag).perform
+      flash["notice"] = "Lists reordered successfully"
+
+      redirect_to polymorphic_path(@tag)
+    else
+      render :manage_list_ordering
```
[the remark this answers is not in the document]
davidgisbey: Yea good point. I'll update it to remove it as it's not possible to trigger a validation with js on anyway

app/controllers/lists_controller.rb
```diff
@@ -8,66 +8,119 @@ def index
...
+    @list = @tag.lists.build(list_params)
+    @list.index = (@tag.lists.maximum(:index) || 0) + 1
 
-    if list.save
-      @tag.mark_as_dirty!
-      flash[:success] = "List created"
+    if redesigned_lists_permission?
+      if @list.save
+        @tag.mark_as_dirty!
+        ListPublisher.new(@tag).perform
```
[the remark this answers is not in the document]
davidgisbey: Yea good shout. This is a hangover from when you could reorder lists which dynamically hit endpoints without being forced to publish the changes (you could just navigate to another page and it wouldn't ever publish the edition). I've removed it from all the endpoints which use the ListPublisher service.

ollietreend (approved):
This looks great @davidgisbey 👍🏻 nice work.

PR #1609, opened 2022-05-25
Add list item reordering

davidgisbey:
## Description 

This PR follows on from #1595 which built the functionality for adding, editing, reordering and deleting lists and #1601 which added the ability to add new list items to a list.

The focus of this PR is adding the ability for a user to reorder list items within a list

### Adds a reorder page

This functions much like the other reorder pages and uses the reorderable list from the publishing components gem.

image

## Guidance for review

Per commit will definitely be easiest. Probs worth giving the designs/last pr a glance if you're not familiar with the card.

Guidance on the publishing components used in the frontend can be found here https://components.publishing.service.gov.uk/component-guide/

## Trello card 

https://trello.com/c/9VoxrVPi/413-rebuild-curate-lists-page-on-collections-publisher

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #1601, opened 2022-05-18
Add list show page and edit list item functionality

davidgisbey:
## Description 

This PR follows on from #1595 which built the functionality for adding, editing, reordering and deleting lists.

The focus of this PR is adding the ability for a user to add additional list items (links) to a list. The ability to reorder list items, move one to a different list and delete a list item will be added in follow up PRs

### Adds a list show page 

A new page that shows all the list items and available actions to add new items, reorder items, move an item to a different list and remove an item from a list. 

For now only the `Add links to list` link is wired in

image

### Add list items page

Adds checkboxes for all the tagged documents which are associated with the mainstream browse page or topic that are not currently in the list

image

## Guidance for review

Per commit will definitely be easiest. Probs worth giving the designs/last pr a glance if you're not familiar with the card.

Guidance on the publishing components used in the frontend can be found here https://components.publishing.service.gov.uk/component-guide/

## Trello card 

https://trello.com/c/9VoxrVPi/413-rebuild-curate-lists-page-on-collections-publisher

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #1612, opened 2022-05-26
Add move to different list functionality

davidgisbey:
### Previous PRs

https://github.com/alphagov/collections-publisher/pull/1595 - built the functionality for adding, editing, reordering and deleting lists
https://github.com/alphagov/collections-publisher/pull/1601 - added the list show page and ability to add new list items to a list.
https://github.com/alphagov/collections-publisher/pull/1609 - reorder list item functionality
https://github.com/alphagov/collections-publisher/pull/1610 - Deleting a list item 

The focus of this PR is adding the ability for a user to move a list item to another list.

### Adds a move link page

image

## Guidance for review

Per commit will definitely be easiest. Probs worth giving the designs/last pr a glance if you're not familiar with the card.

Guidance on the publishing components used in the frontend can be found here https://components.publishing.service.gov.uk/component-guide/

## Trello card 

### DIFF
--- a/app/controllers/list_items_controller.rb
+++ b/app/controllers/list_items_controller.rb
@@ -85,10 +85,28 @@ class ListItemsController < ApplicationController
     @list_item = @list.list_items.find(params[:id])
   end
 
+  def move
+    @list_item = @list.list_items.find(params[:id])
+  end
+
+  def update_move
+    @list_item = @list.list_items.find(params[:id])
+    @list_item.errors.add(:new_list_id, "Choose a list") if move_list_item_params.blank?
+    render :move and return if @list_item.errors.present?
+
+    new_list = @tag.lists.find(move_list_item_params)
+    new_index = new_list.list_items.map(&:index).max.to_i + 1
+    @list_item.update!(list: new_list, index: new_index)
+    ListPublisher.new(@tag).perform
+    flash[:notice] = "Item moved to new list successfully"
+
+    redirect_to tag_list_path(@tag, @list)
+  end
+
 private
 
   def get_layout
-    if redesigned_lists_permission? && action_name.in?(%w[confirm_destroy])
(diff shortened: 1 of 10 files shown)


app/controllers/list_items_controller.rb
```diff
@@ -85,10 +85,28 @@ def confirm_destroy
...
+    @list_item = @list.list_items.find(params[:id])
+  end
+
+  def update_move
+    @list_item = @list.list_items.find(params[:id])
+    @list_item.errors.add(:new_list_id, "Choose a list") if move_list_item_params.blank?
+    render :move and return if @list_item.errors.present?
+
+    new_list = @tag.lists.find(move_list_item_params)
+    new_index = new_list.list_items.map(&:index).max.present? ? new_list.list_items.map(&:index).max + 1 : 1
```
benjamineskola, line 98: What about:

```suggestion
    new_index = new_list.list_items.map(&:index).max.to_i + 1
```

If I'm getting this right, it'll only be not `present?` if the list is empty or `index` returned nil, i.e. in either of those cases `max` is nil. `nil.to_i` is 0, so we'd want to add 1 in either case; this way we can avoid recalculating the map and max.

app/models/tag.rb
```diff
@@ -147,6 +147,12 @@ def top_level_mainstream_browse_page?
     false # should be overridden in subclasses
   end
 
+  def lists_that_do_not_include_list_item(list_item)
+    lists.reject do |list|
+      list.list_items.map(&:base_path).any? { |base_path| base_path == list_item.base_path }
```
benjamineskola, line 152:
```suggestion
      list.list_items.any? { |item| item.base_path == list_item.base_path }
```

benjamineskola (commented):
Just a couple of suggestions, neither is critical.

davidgisbey:
Have actioned. Thanks foe the review!

PR #1617, opened 2022-06-07
Update content of flash messages for lists and list items

davidgisbey:
## Description 

This adds additional content based on feedback from Design that the flash messages should be slightly more verbose and use consistent language.

The document with the requested changes can be found here https://docs.google.com/document/d/1FWW9oY6vD6WTJH5QJh2xqV2rc1onU9ajC63Bo2xUYno/edit

### New list 

|Before|After|
|-----------|-----------|
|image|image|

### Edit list

|Before|After|
|-----------|-----------|
|image|image|

### Delete list 

|Before|After|
|-----------|-----------|
|image|image|

### Adding links

|Before|After|
|-----------|-----------|
|image|image|

**and singular**

image
 
### Removing link

|Before|After|
|-----------|-----------|
|image|image|

### Reordering links

|Before|After|
|-----------|-----------|
|image|image|

### Moving link to different list

|Before|After|
|-----------|-----------|
|image|image|

## Trello card 

### DIFF
--- a/app/controllers/lists_controller.rb
+++ b/app/controllers/lists_controller.rb
@@ -28,7 +28,7 @@ class ListsController < ApplicationController
     if redesigned_lists_permission?
       if @list.save
         ListPublisher.new(@tag).perform
-        flash[:notice] = "List created"
+        flash[:notice] = "#{@list.name} list created"
 
         redirect_to polymorphic_path(@tag)
       else
@@ -56,7 +56,7 @@ class ListsController < ApplicationController
     if redesigned_lists_permission?
       if list.destroyed?
         ListPublisher.new(@tag).perform
-        flash[:notice] = "List deleted"
+        flash[:notice] = "#{list.name} list deleted"
       else
         flash[:alert] = "Could not delete the list"
       end
@@ -80,7 +80,7 @@ class ListsController < ApplicationController
     if redesigned_lists_permission?
       if @list.update(list_params)
         ListPublisher.new(@tag).perform
-        flash[:notice] = "List updated"
+        flash[:notice] = "#{@list.name} list updated"
 
         redirect_to polymorphic_path(@tag)
       else
@@ -122,7 +122,7 @@ class ListsController < ApplicationController
 
     if save_list_items
       ListPublisher.new(@tag).perform
-      flash[:notice] = "Items added to list successfully"
+      flash[:notice] = "#{helpers.pluralize(edit_list_items_params.count, 'link')} successfully added to the list"
 
       redirect_to tag_list_path(@tag, @list)
     else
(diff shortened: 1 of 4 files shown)


nnagewad:
👍🏽

spec/controllers/lists_controller_spec.rb
```diff
@@ -264,7 +276,7 @@
     context "Tag is a MainstreamBrowsePage" do
       let(:tag) { create(:mainstream_browse_page, :published) }
 
-      it "creates a new list item and makes the correct calls to the Publishing API" do
+      it "updates the list link ordering and makes the correct calls to the Publishing API" do
```
kevindew, line 279: Nice fix on the terminology here.

app/controllers/lists_controller.rb
```diff
@@ -122,7 +122,8 @@ def update_list_items
 
     if save_list_items
       ListPublisher.new(@tag).perform
-      flash[:notice] = "Items added to list successfully"
+      new_list_item_count = edit_list_items_params.count
+      flash[:notice] = "#{new_list_item_count} #{'link'.pluralize(new_list_item_count)} successfully added to the list"
```
kevindew, line 126:
```suggestion
      flash[:notice] = "#{'pluralize(edit_list_items_params.count, 'link')} successfully added to the list"
```

minor, this seems a little cleaner and save the need for an extra variable.

spec/controllers/lists_controller_spec.rb
```diff
@@ -110,16 +111,25 @@
...
-        patch :update_list_items, params: { tag_id: tag.content_id, id: list.id, list: { list_items: ["/new-list"] } }
+        patch :update_list_items, params: {
+          tag_id: tag.content_id,
+          id: list.id,
+          list: { list_items: ["/new-list", "/newer-list"] },
+        }
 
-        new_list_item = list.reload.list_items.last
+        new_list_item1 = list.reload.list_items[-2]
+        new_list_item2 = list.list_items.last
```
kevindew, line 121: The `-2` here is a bit cryptic.

Assuming this is an active record collection I wonder if you can do: `new_list_item1, new_list_item2 = list.list_items.last(2)`

kevindew (commented):
Nice changes, screenshots look great. I've just made a couple of small code suggestions.

spec/controllers/lists_controller_spec.rb
davidgisbey: Ooooh that's really cool. Thanks!

app/controllers/lists_controller.rb
davidgisbey: I did think about using this, but it requires inclusion of the `ActionView::Helpers::TextHelper` module. Do you think that's preferable?

kevindew: Oh do we not have that already included in order to be able to call `pluralize` on a string? I think you can call helpers from a controller by doing `helpers.pluralize` with the helpers object having access to all your view helper methods (https://www.bigbinary.com/blog/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers)

davidgisbey:
Have addressed the comments. Thanks very much for the review!

ollietreend (approved):
Nice – looks good to me 👍🏻

PR #1655, opened 2022-08-09
Allow draft Level 1 Specialist topics to be removed 

AgaDufrat:
## Context 

Find and view - Navigation and presentation team will be archiving a lot of specialist topics as part of our epic to retire the specialist topic tree. We did work to allow Level 1 Specialist topics to be archived.

For completeness and consistency, we should also allow draft level 1 Specialist topics to be removed in the UI.
There is currently a draft level 1 topic in production titled "To delete". 

<kbd>Screenshot 2022-08-09 at 11 24 03</kbd>

[Trello](https://trello.com/c/trPPWKvB/1185-allow-draft-level-1-specialist-topics-to-be-removed-m)

---

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -154,15 +154,15 @@ class Tag < ApplicationRecord
   end
 
   def can_be_archived?
-    raise NotImplementedError
+    raise NoMethodError
   end
 
   def can_be_removed?
-    raise NotImplementedError
+    raise NoMethodError
   end
 
   def can_have_email_subscriptions?
-    raise NotImplementedError
+    raise NoMethodError
   end
 
 private
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -157,6 +157,10 @@ class Tag < ApplicationRecord
     raise NotImplementedError
   end
 
+  def can_be_removed?
+    level_two? && draft?
+  end
+
   def can_have_email_subscriptions?
     raise NotImplementedError
   end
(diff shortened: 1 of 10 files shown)


app/models/tag.rb
```diff
@@ -158,7 +158,7 @@ def can_be_archived?
   end
 
   def can_be_removed?
-    level_two? && draft?
+    raise NotImplementedError
```
chao-xian, line 161: We shouldn't use this type of [Error](https://oleg0potapov.medium.com/ruby-notimplementederror-dont-use-it-dff1fd7228e5) - it'd be better to raise an Error with a message?

Tetrino: @AgaDufrat The rest looks good to me so if you can flip this change in some point I can approve 👍

AgaDufrat: oh, interesting. I first read about the template method pattern (and the use of `NotImplementedError` for it) in Sandi Metz’s “Practical Object-Oriented Design in Ruby” https://cek.io/blog/2014/05/06/poodr-ch-6
I'll change it here and in other places.

Tetrino: I've also been using `NotImplementedError` but I don't want to override Kelv on it (we may have internal precedence).

Tetrino (approved):

app/models/tag.rb
chao-xian: Response from one of our Lead Devs is that they avoid using `NotImplementedError` for this context.

kevindew (approved):
Looks good to me

PR #1749, opened 2022-11-23
Use content ids to present curated lists (aka groups)

AgaDufrat:
## Context

When a content item is tagged to curated Mainstream browse or Specialist topic (parent class Tag) and this content item’s slug changes the link disappears from the topic page. This is because the curated lists/groups are based on base paths. 

The “contents” slugs don’t get updated as a part of dependency resolution in Publishing API.

## What
Use content_ids to present curated lists (aka groups) instead of base_paths, as they are less volatile.

### Notes 
The “contents” key will be removed after the migration.

## Depends on
- https://github.com/alphagov/publishing-api/pull/2201

## Post-deployment tasks

- [ ] Republish all Mainstream browse pages
- [ ] Republish all Specialist topics

---

### DIFF
--- a/app/presenters/groups_presenter.rb
+++ b/app/presenters/groups_presenter.rb
@@ -7,7 +7,8 @@ class GroupsPresenter
     @tag.lists.ordered.map do |list|
       {
         name: list.name,
-        contents: list.tagged_list_items.map(&:base_path),
+        contents: list.tagged_list_items.map(&:base_path), # should be removed once migration to content_ids is complete
+        content_ids: list.tagged_list_items.map(&:content_id),
       }
     end
   end
--- a/spec/controllers/list_items_controller_spec.rb
+++ b/spec/controllers/list_items_controller_spec.rb
@@ -35,6 +35,7 @@ RSpec.describe ListItemsController, type: :controller do
               {
                 "name" => list.name,
                 "contents" => [],
+                "content_ids" => [],
               },
             ],
             "internal_name" => tag.title,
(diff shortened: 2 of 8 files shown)


spec/controllers/list_items_controller_spec.rb
```diff
@@ -196,12 +197,18 @@
                   "contents" => [
                     list_item2.base_path,
                   ],
+                  "content_ids" => %w[
+                    456
```
kevindew, line 201: Should this be list_item2.content_id as per previous usage? similar for below

spec/controllers/lists_controller_spec.rb
```diff
@@ -142,6 +142,12 @@
                     "/new-list",
                     "/newer-list",
                   ],
+                  "content_ids" => %w[
+                    123
+                    456
```
kevindew, line 147: Similar point about above where some of the items in the above list use variables rather than hardcoding (not sure why it's only some). Again similar comments for changes below.

app/presenters/groups_presenter.rb
```diff
@@ -8,6 +8,7 @@ def groups
       {
         name: list.name,
         contents: list.tagged_list_items.map(&:base_path),
```
kevindew, line 10: Should we flag this with a comment that it should be removed once migration to content_ids?

kevindew (commented):
Looks good, just a couple of comments on the test consistency

spec/controllers/list_items_controller_spec.rb
kevindew: Ah sure, do they not get accessed when calling the content_id method? I guess it should be calling the [content_id on tagged documents](https://github.com/alphagov/collections-publisher/blob/7818ba97bc2958ab0614ae4a863229cecd875752/app/models/list_item.rb#L16) ?

kevindew (approved):

PR #1772, opened 2022-12-16
Add rake task to republish tags with curated lists

AgaDufrat:
The existing `send_published_tags` task does not represent "groups" which are used to render curated lists. 
https://github.com/alphagov/collections-publisher/blob/main/app/presenters/tag_presenter.rb#L79-L84

We need to republish tags with curated lists / groups to migrate from using "contents" to "content_ids".

Code inspired by [ListPublisher](https://github.com/alphagov/collections-publisher/blob/main/app/services/list_publisher.rb) and TagRepublisher.

### DIFF
--- a/app/services/list_republisher.rb
+++ b/app/services/list_republisher.rb
@@ -1,0 +1,61 @@
+class ListRepublisher
+  def republish_tags(tags)
+    done = 0
+    tags.find_each do |tag|
+      next if tag.lists.blank?
+
+      update_groups(tag)
+      republish_tag(tag)
+
+      done += 1
+
+      if (done % 100).zero?
+        log "#{done} completed..."
+      end
+    end
+
+    log "All done"
+  end
+
+private
+
+  def update_groups(tag)
+    groups = GroupsPresenter.new(tag).groups
+    tag.update!(published_groups: groups, dirty: false)
+  end
+
+  def republish_tag(tag)
+    with_retry do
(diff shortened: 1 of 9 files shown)


app/services/list_republisher.rb
```diff
@@ -0,0 +1,61 @@
...
+    done = 0
+    tags.find_each do |tag|
+      next if tag.lists.blank?
+
+      update_groups(tag)
+      republish_tag(tag)
+
+      done += 1
+
+      if (done % 100).zero?
```
hannako, line 12: What's the significance of `% 100` 

Could you use each_with_index instead of the numerator? Or am I missing something - which is VERY LIKELY tbh!

AgaDufrat: Blindly copied from https://github.com/alphagov/collections-publisher/blob/main/app/services/tag_republisher.rb
I don't have special feelings about `% 100` 😆 and wouldn't question Tijmen.
To me it serves the purpose of outputting something to the console to indicate that it's doing something.

ollietreend: I guess it's to reduce the amount of log output – so you'd end up with:

```
0 completed...
100 completed...
200 completed...
```

Rather than a (much noisier):

```
0 completed...
1 completed...
2 completed...
3 completed...
```

ollietreend (approved):
This seems reasonable to me! 👍🏻

PR #1889, opened 2023-05-25
Enable specialist topics archiving

BeckaL:


app/services/generate_single_page_list_params_service.rb
```diff
@@ -0,0 +1,16 @@
+module GenerateSinglePageListParamsService
+  def generate_single_page_list_params(content_item)
```
hannako, line 2: I think we could maybe put a comment here, with a link to the code in email alert frontend.

hannako (commented):
This is looking really great Becka ⭐ Thank you so much for all the work you've put in - hugely appreciated!! I hope the comments make sense. None are really blockers - just ideas to try and make it a bit clearer what we're implementing. As it's complicated 🙃

app/services/email_alerts_updater.rb
```diff
@@ -11,6 +13,20 @@ def initialize(item:, successor:, govuk_request_id: nil)
...
+  def handle
+    if successor.is_a?(ContentItem) && successor.document_collection_maps_to_tag?(item)
+      bulk_migrate
+    else
+      unsubscribe
+    end
+  end
+
+  def bulk_migrate
+    from_slug = subscriber_list_slug
```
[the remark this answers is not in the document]
BeckaL: Good point. I've called it `item_subscriber_list` since it could also be a mainstream browse page (obviously not for the bulk migrate, but where we use it in the bulk_unsubscribe) and so this matches the language of this class

app/services/email_alerts_updater.rb
```diff
@@ -0,0 +1,45 @@
+class EmailAlertsUpdater
+  def self.call(...)
+    new(...).unsubscribe
+  end
+
+  attr_reader :item, :body, :successor, :govuk_request_id
+
+  def initialize(item:, successor:, govuk_request_id: nil)
```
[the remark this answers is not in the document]
BeckaL: Yeah agreed, I spent some time looking at this a few weeks ago and thought there might be a bit of YAGNI going on. Agree that it's not used anywhere, have removed.

spec/models/content_item_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "rails_helper"
+require "gds_api/test_helpers/content_store"
+
+RSpec.describe "ContentItem" do
```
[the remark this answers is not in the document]
BeckaL: Leaving for now depending on what we decide on the above

app/services/email_alerts_updater.rb
```diff
@@ -11,6 +13,20 @@ def initialize(item:, successor:, govuk_request_id: nil)
     @govuk_request_id = govuk_request_id
   end
 
+  def handle
+    if successor.is_a?(ContentItem) && successor.document_collection_maps_to_tag?(item)
```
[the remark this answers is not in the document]
BeckaL: Oh nice naming, much better than what I came up with. Done

app/services/generate_single_page_list_params_service.rb
BeckaL: OK done, this is my comment:

```  # This is equivalent to code introduced into email alert frontend:
  # https://github.com/alphagov/email-alert-frontend/blob/main/app/services/subscriber_list_params/generate_single_page_list_params_service.rb
  # to handle subscriber lists with reverse links.
  # It will only be needed in here temporarily while we migrate specialist topics
```

Happy to finesse in any way, I'm not sure about language here

app/services/email_alerts_api_params_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+  def document_collection_subscriber_list_params(content_item)
+    {
+      "url" => content_item.base_path,
+      "title" => content_item.title,
+      "content_id" => content_item.content_id,
+      "description" => content_item.description,
+    }.merge(links(content_item))
+  end
+
+  def links(content_item)
```
hannako, line 15: Minor - maybe we should call this `document_collection_links`

spec/services/email_alerts_updater_spec.rb
```diff
@@ -5,55 +5,95 @@
...
+      let(:successor) { ContentItem.new(successor_data) }
 
-      expect {
-        described_class.call(
-          item: OpenStruct.new(subscriber_list_search_attributes: { "links" => { topics: %w[ABC] } }, content_id: "ABC"),
-          successor:,
+      before do
+        stub_email_alert_api_creates_subscriber_list(
+          "title" => "Successor document collection",
+          "links" => { "document_collections" => [topic.content_id] },
```
hannako, line 83: I know it doesn't matter, but I'd stick a random uiid in here instead, as it's a bit confusing seeing `topic.content_id` which would be used for the `details[mapped_specialist_topic_content_id]`

spec/services/email_alerts_updater_spec.rb
```diff
@@ -5,55 +5,95 @@
...
+          "content_id" => "835af2ae-ed12-49f0-9b3c-d6795d028484",
+          "description" => "description for Successor document collection",
+          "slug" => "slug",
         )
-      }.to raise_error(GdsApi::HTTPNotFound)
+      end
+
+      it "makes a call the email-alert-api bulk migrate endpoint with the slug, and generated email body and sender_message_id" do
```
hannako, line 90: Cut and paste error here - no email body etc

spec/services/email_alerts_updater_spec.rb
```diff
@@ -5,55 +5,95 @@
...
+        stub_email_alert_api_bulk_migrate(to_slug: "slug", from_slug: "tax-credits-and-child-benefit-child-benefit")
+
+        expect(Services.email_alert_api).to receive(:bulk_unsubscribe).never
+
+        expect { described_class.call(item: topic, successor:) }.to_not raise_error
```
hannako, line 95: Same comment as above - think we should check the stub is requested

hannako (commented):
Looks great! Excited to test on integration!

spec/services/email_alerts_updater_spec.rb
BeckaL: Oh I think that's the point here, i.e. it should be the same thing because it's a mapped specialist topic content id?

hannako: No, this is the reverse link, so it would be the content id of the document collection page itself. Sorry - complicated! Hope that makes sense 😅

app/services/email_alerts_updater.rb
```diff
@@ -0,0 +1,64 @@
+class EmailAlertsUpdater
+  include EmailAlertsApiParamsHelper
+
+  def self.call(...)
+    new(...).handle
+  end
+
+  attr_reader :item, :body, :successor
```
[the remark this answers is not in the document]
BeckaL: no good spot, removed

app/services/email_alerts_updater.rb
```diff
@@ -22,13 +38,22 @@ def unsubscribe
...
+
+  def topic_subscriber_list_slug
+    params = specialist_topic_subscriber_list_params(item)
+    subscriber_list = Services.email_alert_api.find_subscriber_list(params)
     subscriber_list.dig("subscriber_list", "slug")
   end
 
+  def existing_subscriber_list_slug_for_document_collection
+    params = document_collection_subscriber_list_params(successor)
+    subscriber_list = Services.email_alert_api.find_or_create_subscriber_list(params).parsed_content["subscriber_list"]
```
hannako, line 53: Tiny thing - do you need `.parsed_content`

eg https://github.com/alphagov/email-alert-frontend/blob/main/app/controllers/content_item_signups_controller.rb#L35-L38

hannako (approved):
Looks great Becka, great work!! 🥇

app/services/email_alerts_updater.rb
BeckaL: Oh good point, updated 👍

sivsubra-tw (approved):
LGTM 👍  Thanks for the context @BeckaL !

PR #1902, opened 2023-06-19
Temporary rake task to update the redirect for an archived specialist topic

hannako:
A specialist topic was archived and redirected to the wrong page. This task allows us to update it. 
See commit message for implementation details

https://trello.com/c/ClsVzfXo/1867-change-redirect-of-1-archived-specialist-topic-page

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -84,4 +84,36 @@ namespace :publishing_api do
     )
     puts "Links patched for root page..."
   end
+
+  desc "Update redirect for an archived tag"
+  task :update_redirect, %i[content_id new_redirect_path] => :environment do |_task, args|
+    new_redirect_path = args[:new_redirect_path]
+    unless new_redirect_path.present? && new_redirect_path.starts_with?("/")
+      raise "new_redirect_path is missing or invalid"
+    end
+
+    if Services.content_store.content_item(new_redirect_path)
+      tag = Tag.find_by(content_id: args[:content_id])
+
+      raise "No tag can be found for that content id" if tag.blank?
+
+      raise "This task can only be used for archived topics" unless tag.state == "archived" && tag.type == "Topic"
+
+      puts "Updating redirect for #{tag.title}"
+      puts "From: '#{tag.redirect_routes.pluck(:to_base_path).uniq.first}'"
+      puts "To: '#{new_redirect_path}'"
+
+      redirects = tag.redirect_routes
+
+      redirects.each do |redirect|
+        redirect.update!(to_base_path: new_redirect_path)
+      end
+
+      tag.reload
+      presenter = ArchivedTagPresenter.new(tag)
+      ContentItemPublisher.new(presenter).send_to_publishing_api
+
+      puts "Task complete"
+    end
+  end
 end
--- a/spec/lib/tasks/publishing_api_spec.rb
+++ b/spec/lib/tasks/publishing_api_spec.rb
@@ -1,5 +1,6 @@
 require "rails_helper"
 require "gds_api/publishing_api/special_route_publisher"
+require "gds_api/test_helpers/content_store"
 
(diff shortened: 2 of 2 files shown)


Tetrino (approved):

lib/tasks/publishing_api.rake
```diff
@@ -84,4 +84,35 @@ namespace :publishing_api do
...
     puts "Links patched for root page..."
   end
+
+  desc "Update redirect for an archived tag"
+  task :update_redirect, %i[content_id new_redirect_path] => :environment do |_task, args|
+    unless args[:new_redirect_path].present? && args[:new_redirect_path].starts_with?("/")
+      raise "new_redirect_path is missing or invalid"
+    end
+
+    if Services.content_store.content_item(args[:new_redirect_path])
```
sivsubra-tw, line 94: Hi, is it helpful to extracting this out to a Boolean function so that we know what the content_item function is doing with the new_redirect_path?

lib/tasks/publishing_api.rake
```diff
@@ -84,4 +84,35 @@ namespace :publishing_api do
...
+      tag = Tag.find_by(content_id: args[:content_id])
+
+      raise "No tag can be found for that content id" if tag.blank?
+
+      raise "This task can only be used for archived topics" unless tag.state == "archived" && tag.type == "Topic"
```
sivsubra-tw, line 99: Is it helpful to add the states and types to an enum?

lib/tasks/publishing_api.rake
```diff
@@ -84,4 +84,35 @@ namespace :publishing_api do
...
+
+      puts "Updating redirect for #{tag.title}"
+      puts "From: '#{tag.redirect_routes.pluck(:to_base_path).uniq.first}'"
+      puts "To: '#{args[:new_redirect_path]}'"
```
sivsubra-tw, line 103: Is it worth extracting the args[:new_redirect_path] into a separate variable as it is being used multiple times?

spec/lib/tasks/publishing_api_spec.rb
```diff
@@ -21,3 +22,69 @@
...
+
+    it "raises an error if we attempt to update the url of a published tag" do
+      stub_content_store_has_item(valid_redirect_url)
+
+      create(:topic, content_id: published_content_id)
+      expected_message = "This task can only be used for archived topics"
+
+      expect { Rake::Task["publishing_api:update_redirect"].invoke(published_content_id, valid_redirect_url) }.to raise_error(expected_message)
+    end
+  end
```
sivsubra-tw, line 89: Is it worth adding another test for when tag type is not Topic?

lib/tasks/publishing_api.rake
hannako: Thanks for the review Siva!

This is just a quick way for us to see if a valid redirect path has been provided. It could just be a `Services.content_store.content_item(args[:new_redirect_path])` without being called as part of an `if`. It's just so that if the content store does not have a valid content item, this line will raise an error and halt the rake task.

Given more time, and a longer term need for this task, I would have liked to create a `redirect` object and include the [valid_govuk_path_validator](https://github.com/alphagov/collections-publisher/blob/main/app/validators/valid_govuk_path_validator.rb). But as time is short and we are removing the code once the task has been run, I've just copied the validation code into this rake task as a quick fix.

Hope that makes sense and you're happy to leave as it is?

lib/tasks/publishing_api.rake
hannako: Unless you feel strongly, as this is a temporary task which I will remove shortly I would probably lean towards leaving as it is! Is that OK with you?

lib/tasks/publishing_api.rake
sivsubra-tw: Sounds good, thank you!

PR #1903, opened 2023-06-20
Revert "Temporary rake task to update the redirect for an archived specialist topic"

hannako:
This task has now been run on production and can be removed. If we want to support this functionality more permanently, we should improve the validation on the redirect content item.

https://trello.com/c/UUhob9oc/1870-remove-temporary-rake-task-in-email-alert-api-and-collections-publisher

Reverts alphagov/collections-publisher#1902


sivsubra-tw (approved):
LGTM 👍 Thank you for removing the now-redundant code!

PR #1956, opened 2023-08-31
Add erb-lint to Collections Publisher

anatron:


app/views/topics/show.html.erb
cynthia-anya, line 1: Linter has added "noopener" keyword to the "rel" attribute. This unsets the Window.opener property and has the same behaviour as target="_blank".

cynthia-anya (approved):
All looks as expected.
'rel: "noopener"' was added to a link, but this is shouldn't be an issue

PR #1960, opened 2023-09-04
Migrate email subscribers to a taxonomy topic list when archiving some specialist topics

hannako:
## What

Update the email alert updater class to bulk migrate specialist topic email subscribers to a new list.

## How
This application serves the UI to curate and archive specialist topic pages. Specialist topics power email alerts, ie there is an email signup button on specialist topic pages ([eg](https://www.gov.uk/topic/dealing-with-hmrc/paying-hmrc)). Subscribing to a specialist topic subscriber list will result in an email being sent to users when a major change is published to any content tagged to the specialist topic.

When we retire and redirect specialist topics, we need to handle their email subscriptions differently depending on what page the topic is being redirected to (in code the page being redirected to is referred to as the `successor`). Prior to this PR, collections publisher's email_alerts_updater service handled scenarios 1 and 2. This PR adds support for scenario 3

Scenario 1: When a specialist topic is being archived and redirected to an existing page on gov.uk, we are bulk unsubscribing its email subscribers. This will be for the majority of specialist topics.

Scenario 2: When a specialist topic is being archived and redirected to a document collection conversion (ie a document collection that contains all of the content from the original specialist topic just in a new document type) we are migrating the specialist topic email subscribers to the document collection subscriber list. 

Scenario 3: For a very small number of specialist topics which we are converting to document collections, the department would like their email subscribers to be migrated to a taxonomy topic list, instead of to the document collection subscriber list. This is because taxonomy topic email subscriptions are more similar to specialist topic subscriptions - they trigger an email when a piece of content tagged to the topic is updated. Document collection email subscriptions by contrast are "page level" and much smaller in scope.

https://trello.com/c/siCRG1zk/1978-extend-the-collections-publisher-tag-archiver-service-to-handle-document-collections-which-have-a-taxonomytopicemailoverride-fie


catalinailie (approved):
LGTM

app/services/email_alert_api/params_formatter.rb
```diff
@@ -0,0 +1,34 @@
...
+      { "links" => { "document_collections" => [content_item.content_id] } }
+    end
+
+    def specialist_topic_subscriber_list_params(topic)
+      topic.subscriber_list_search_attributes
+    end
+
+    def taxonomy_topic_subscriber_list_params(content_item)
+      {
+        "title" => content_item["title"],
```
anatron, line 28: Assuming the content_item in this method has the same data type as the content_item passed to `document_collection_subscriber_list_params` in line 7, why do we access title using square brackets here in line 28 and using a dot operator in line 10?

hannako: Ah yes this is confusing.

In line 7 we are passing an instance of the ContentItem class. So we can call the title method etc on that object. In line 28, we are passing this through, ie it is just a hash. Hence needing the square brackets.

I think I will amend this. Will push a fresh commit shortly. Thanks for the thorough review!

app/services/email_alert_api/subscriber_list_updater.rb
```diff
@@ -0,0 +1,85 @@
...
+  class SubscriberListUpdater
+    include ParamsFormatter
+
+    def self.call(...)
+      new(...).handle
+    end
+
+    attr_reader :item, :successor
+
+    def initialize(item:, successor:)
```
sivsubra-tw, line 11: Assuming item here is `content_item`, is it worth calling it as such for consistency with other files?

spec/services/email_alert_api/subscriber_list_updater_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+                                .to_return(status: 200)
+
+          expect(Services.email_alert_api).to receive(:bulk_unsubscribe).never
+
+          described_class.call(item: topic, successor:)
+
+          expect(bulk_migrate_to_taxonomy_topic_stub).to have_been_requested
+        end
+
+        it "successor has a mapped specialist topic content id that matches the specialist topic's content id and also a taxonomy topic email override" do
```
sivsubra-tw, line 140: Am I right in stating that the `it` is describing the setup here (and not the behavioural expectations)?  e.g. should it actually be `it "should subscribe to associated taxonomy topic"?

sivsubra-tw (approved):
Changes themselves look good - just some nitpicky suggestions. Thanks for your work! 

cc: @anatron

app/services/email_alert_api/subscriber_list_updater.rb
hannako: Ah damn, I should have done that in https://github.com/alphagov/collections-publisher/pull/1965 

Yes, I'll update, thanks

spec/services/email_alert_api/subscriber_list_updater_spec.rb
hannako: You're right, and this is a consequence of the first commit where I've grouped the context blocks by the outcome: migrate or unsubscribe. Which is unconventional, and I obsessed over this file for a long time 😂 But I think (not sure if you'll agree!) that if you read the whole file overall it makes the behaviour of the service easier to understand.

app/services/email_alert_api/subscriber_list_updater.rb
hannako: https://github.com/alphagov/collections-publisher/pull/1960/commits/d10b1a8f1a427d39e78504fdd456a43768449f4a

hannako:
Thanks for the review comments @sivsubra-tw I've replied and pushed a change - could you let me know when you're happy for me to merge. Thank you

anatron:
@hannako Happy for it to be merged whenever you're ready!

PR #1977, opened 2023-09-22
Reinstate rake task to update the redirect for an archived specialist topic

hannako:
Reverts alphagov/collections-publisher#1903

A specialist topic has been archived and redirected to the wrong page, and we once again need to update the redirect!


FelixMarcusMillne (approved):
Sure, urgent requirement - given this is happening again, let's get this properly written, rather than removing it again, please?

catalinailie (approved):

PR #2082, opened 2024-02-08
Redirect the root specialist topic page to /browse

hannako:
Once all Level one and level two specialist topics have been redirected, we can redirect the root page to [Services and Information](https://www.gov.uk/browse)

https://trello.com/c/QrOJukDq/2373-retire-final-root-specialist-topic-page-s


Tetrino (approved):
LGTM

PR #2099, opened 2024-03-05
Remove specialist topic code from repo 1/4

georges1996:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

https://trello.com/c/RLc9saF6/2461-remove-specialist-topic-code-from-collections-publisher-m-l

This PR just removes most references to topics mainly cosmetic references in views and controllers and some usecases. This became unwieldy so moved the removal of the tag topic object and its related components here: https://github.com/alphagov/collections-publisher/pull/2102

PR #2102, opened 2024-03-11
Remove specialist topic code (dropping Tags of type topic) 2/4

georges1996:

PR 1/2: https://github.com/alphagov/collections-publisher/pull/2099

Ticket: https://trello.com/c/RLc9saF6/2461-remove-specialist-topic-code-from-collections-publisher-m-l

**dropping Tags of type `topic`**


beccapearce (approved):

beccapearce:
Looks good and works locally

app/presenters/mainstream_browse_page_presenter.rb
```diff
@@ -9,7 +9,6 @@ def format
 
   def links
     super.merge(
-      "related_topics" => @tag.topics.order(:title).map(&:content_id),
```
hannako, line 12: It looks like we missed removing `related_topics` from collections.  https://github.com/search?q=repo%3Aalphagov%2Fcollections%20related_topics&type=code

@georges1996 please could you open a PR to remove that code from collections first, and link it here? Reason being - the process we've followed so far is to remove any dependency on specialist topics from the frontend first, so that we can feel confident that when these fields are missing from the content items nothing will blow up. So we should stick with that process. 

@unoduetre please could you remove `related_topics` from [these examples](https://github.com/search?q=repo%3Aalphagov%2Fpublishing-api%20related_topics&type=code) in publishing API. There might be a dependency here so that George's PR in collections has to go live first.

hannako (approved):

anatron (approved):
LGTM

PR #2105, opened 2024-03-12
Add GA4 analytics

patrickpatrickpatrick:
## What

Use the `load-analytics' script of govuk_publishing_components to add GA4 analytics to collections-publisher. Includes extra configuration to set the correct domains for GA4 to initialise and code to automatically accept all cookies to make tracking work.

## Why

UA is being deprecated, we need to switch our apps to use GA4 instead.

[Relevant Trello Card](https://trello.com/c/U1GlVZRF/196-implement-page-view-tracking-in-ga4-for-mpe-owned-publishing-apps)

### DIFF
--- a/app/assets/javascripts/domain-config.js
+++ b/app/assets/javascripts/domain-config.js
@@ -1,0 +1,24 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.vars = window.GOVUK.vars || {}
+window.GOVUK.vars.extraDomains = [
+  {
+    name: 'production',
+    domains: [
+      'collections-publisher.publishing.service.gov.uk'
+    ],
+    initialiseGA4: true,
+    id: 'GTM-P93SHJ4Z',
+    gaProperty: 'UA-26179049-6'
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/domain-config.js
```diff
@@ -0,0 +1,41 @@
...
+  {
+    name: 'production',
+    domains: [
+      'collections-publisher.publishing.service.gov.uk'
+    ],
+    initialiseGA4: true,
+    id: 'GTM-MG7HG5W',
+    gaProperty: 'UA-26179049-1',
+    gaPropertyCrossDomain: 'UA-145652997-1',
+    consentApiUrl: 'production'
```
andysellick, line 13: I don't think you'll need to define `consentApiUrl` for collections-publisher, as this relates to the single cookie consent code and it won't be using that. I think these lines can be safely removed.

patrickpatrickpatrick: Thanks! Will remove these lines :)

PR #2111, opened 2024-03-18
Drop the tag_associations table 3/4

georges1996:

Once we have removed all the associated code to the tag_associations table, we can now drop the table. Note I have done this in a way where we can roll back the migration in case anything goes wrong, in order to regenerate the table structure. 

https://trello.com/c/RLc9saF6/2461-remove-specialist-topic-code-from-collections-publisher-m-l

### DIFF
--- a/db/migrate/20240318080818_drop_associations_table.rb
+++ b/db/migrate/20240318080818_drop_associations_table.rb
@@ -1,0 +1,22 @@
+class DropAssociationsTable < ActiveRecord::Migration[7.1]
+  def change
+    remove_foreign_key "tag_associations", "tags",
+                       column: "to_tag_id",
+                       name: "tag_associations_to_tag_id_fk",
+                       on_delete: :cascade
+
+    remove_foreign_key "tag_associations", "tags",
+                       column: "from_tag_id",
+                       name: "tag_associations_from_tag_id_fk",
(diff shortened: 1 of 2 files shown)


db/migrate/20240318080818_drop_associations_table.rb
```diff
@@ -0,0 +1,10 @@
+class DropAssociationsTable < ActiveRecord::Migration[7.1]
+  def change
+    drop_table :tag_associations do |t|
```
unoduetre, line 3: Might be useful to drop index separately in case one wants to reverse this migration.

unoduetre (changes requested):

beccapearce (approved):

db/migrate/20240318080818_drop_associations_table.rb
unoduetre: Yes, you can reverse it, but the result won't be the same. You can compare the schema before and after reversing.

unoduetre: I've confirmed that is not the case, so it needs to be added.

unoduetre (approved):
Approved.

anatron (approved):
LGTM!

PR #2118, opened 2024-03-20
Migration to remove tags of type topic 4/4

georges1996:
This is the final PR to remove topics from this app. At this point all code and reference to topics has been removed and this last PR runs a migration to remove **Tags** of type **Topic** from the database.

Ticket: https://trello.com/c/RLc9saF6/2461-remove-specialist-topic-code-from-collections-publisher-m-l

Subsequent PR's
- https://github.com/alphagov/collections-publisher/pull/2099
- https://github.com/alphagov/collections-publisher/pull/2102
- https://github.com/alphagov/collections-publisher/pull/2111

### DIFF
--- a/db/migrate/20240320072303_remove_tags_of_type_topic.rb
+++ b/db/migrate/20240320072303_remove_tags_of_type_topic.rb
@@ -1,0 +1,5 @@
+class RemoveTagsOfTypeTopic < ActiveRecord::Migration[7.1]
+  def change
+    Tag.where(type: "Topic").destroy_all
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_03_18_080818) do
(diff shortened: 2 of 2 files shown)


beccapearce (approved):

db/migrate/20240320072303_remove_tags_of_type_topic.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveTagsOfTypeTopic < ActiveRecord::Migration[7.1]
+  def change
+    Tag.where(type: "Topic").delete_all
```
unoduetre, line 3: Might be better to not use the model (e.g. see https://rails.rubystyle.guide/#define-model-class-migrations) but if they are used in other migrations, then maybe it doesn't add any value?

hannako: I'm extremely risk averse when it comes to db migrations!

- Bit of a nitpick, but `Tag.where(type: Topic.sti_name)` might be better practice I think?

- Why did you pick `delete_all` instead of `destroy_all`?

anatron (approved):
LGTM

PR #2151, opened 2024-04-28
Don't use String::gsub where we mean String::strip.

sengi:
Resolves https://github.com/alphagov/collections-publisher/security/code-scanning/2 (not an actual security issue though).


ChrisBAshton (approved):
LGTM ⭐

PR #2219, opened 2024-07-22
Upgrade govuk publishing components gov.uk frontend v5

davidtrussler:
[Trello](https://trello.com/c/HYqXLv8w/347-upgrade-govukpublishingcomponents-gem-dependency-v5-of-govuk-frontend-collections-publisher)

This work updates Collections Publisher to V5 of the Design System. The only visual changes this makes are shown below. All other issues identified when testing (see [here](https://trello.com/c/3KcURoRn)) are now resolved.

Visual changes: 
- Tags now appear with the first letter capitalised rather than all lower case
- The font size of the list of subtopics is increased from 14px to 16px. The Design System is moving away from 14px text. 

|Updated|Current|
|-|-|
|Screenshot 2024-07-22 at 15 45 45|Screenshot 2024-07-22 at 15 45 59|

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -197,7 +197,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (39.2.4)
+    govuk_publishing_components (40.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,4 +1,5 @@
 //= link_tree ../images
 //= link application.css
 //= link application.js
+//= link es6-components.js
 //= link domain-config.js
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -2,8 +2,15 @@
 // system aspects of Collections Publisher
 
 //= require govuk_publishing_components/dependencies
-//= require govuk_publishing_components/all_components
-//= require govuk_publishing_components/analytics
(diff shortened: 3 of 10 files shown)


Gemfile
```diff
@@ -17,7 +17,8 @@ gem "terser"
 gem "gds-api-adapters"
 gem "gds-sso"
 gem "govuk_app_config"
-gem "govuk_publishing_components"
+# gem "govuk_publishing_components"
+gem "govuk_publishing_components", "40.0.0"
```
mtaylorgds, line 21: We can probably just revert this line back to `gem "govuk_publishing_components"` now, can't we?

app/views/navigation_rules/edit.html.erb
```diff
@@ -18,7 +18,7 @@
 <% content_for :title, @step_by_step_page.title %>
 <% content_for :context, "Sidebar settings" %>
 
-<%= render "shared/steps/form_notice", message: notice, status: "success" %>
+<%= render "shared/steps/form_notice", message: notice, status: nil %>
```
mtaylorgds, line 21: What is this (and a similar change in `app/views/step_by_step_pages/index.html.erb`) about?

mtaylorgds (commented):
Couple of small comments.

Gemfile
davidtrussler: 👍 Yes, I have changed that to match the equivalent in Publisher.

app/views/navigation_rules/edit.html.erb
davidtrussler: I have restored these to their original state now since these changes are not necessary for this piece of work.

mtaylorgds (approved):
LGTM

PR #2472, opened 2025-05-29
Revert "Bump rails from 7.1.3.4 to 8.0.2"

Tetrino:
Reverts alphagov/collections-publisher#2413

Have an issue with Parse which is preventing some pages from loading that was only made apparent in the e3e tests. Struggling to track it down so reverting for now.


syed-ali-tw (approved):

PeterHattyar (approved):
All looks good

PR #2474, opened 2025-05-29
Rails 8 upgrade - second attempt.

Tetrino:
Previous assumption was that JSON would be valid, turns out it is not,
however according to the documentation YAML is treated perfectly fine
when with an Array. Previous serialisation attempt failed due to this.
 
https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html


config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,30 @@
...
+# When both `If-Modified-Since` and `If-None-Match` are provided by the client
+# only consider `If-None-Match` as specified by RFC 7232 Section 6.
+# If set to `false` both conditions need to be satisfied.
+#++
+# Rails.application.config.action_dispatch.strict_freshness = true
+
+###
+# Set `Regexp.timeout` to `1`s by default to improve security over Regexp Denial-of-Service attacks.
+#++
+# Regexp.timeout = 1
```
baisa, line 30: Do we still need this file as we have `config.load_defaults "8.0"` ?

PR #2783, opened 2026-07-06
Remove metadata component

andysellick:

## What
Remove instances of the [metadata component](https://components.publishing.service.gov.uk/component-guide/metadata) from the application and replace with custom styling.

## Why
The appearance of the metadata component here was being overridden by custom styles within the application. This is against our principle of component isolation as it makes it harder to make changes to components without instances like this breaking. The required styles also differ significantly from the appearance of the component.

## Visual changes
Shouldn't be any visual difference. This applies to the content of the grey box on the bottom right of the below screenshot, on the step by step and coronavirus pages.

Screenshot 2026-07-06 at 11 18 43

### DIFF
--- a/spec/features/reviewing_step_by_steps_spec.rb
+++ b/spec/features/reviewing_step_by_steps_spec.rb
@@ -111,7 +111,7 @@ RSpec.feature "Reviewing step by step pages" do
   end
 
   def and_the_step_by_step_status_should_be(status)
-    expect(page).to have_css(".gem-c-metadata", text: "Status: #{status}")
+    expect(page).to have_css(".metadata", text: "Status: #{status}")
   end
 
   def and_i_submit_the_form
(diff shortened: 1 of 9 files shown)


spec/features/reviewing_step_by_steps_spec.rb
```diff
@@ -111,7 +111,7 @@ def and_i_should_see_my_requested_changes_in_the_change_notes
   end
 
   def and_the_step_by_step_status_should_be(status)
-    expect(page).to have_css(".gem-c-metadata", text: "Status: #{status}")
```
AshGDS, line 114: I think screen readers will still announce the colon when it's in the `:after` styles but do we know if it's  read out any differently now that it's not it in the actual text?

AshGDS (dismissed):
LGTM 👍

spec/features/reviewing_step_by_steps_spec.rb
andysellick: That's a fair question. I'm not sure, to avoid doubt I've switched it so the colon is back in the HTML.

AshGDS: Thanks 👍

AshGDS (approved):

